PR #543, opened 2014-06-19
Adding header metadata so the page reduces to a small size on the phone.

carolyncole:
This fixes the issue when you bring up sufia on a small device it just shows a really small version of the page instead of the compressed column view of the page.


cam156:
@jrochkind I'm guessing it was just copy paste from stack overflow, not any knowledge on my part, but that was more than 3 years ago.  I would suggest testing it.

PR #678, opened 2014-10-07
Move the Collection model into the Sufia namespace

jcoyne:
Many engines may define a Collection model, so we don't want to pollute
the namespace.  For backwards compatibility there is a Collection model
generated into the application's directory that extends the
Sufia::Collection.  This ensures that we don't need to update the
hasModel assertions in the Fedora repository.

As this is a breaking change (if the Collection model is not generated
into the application), it's recommended that this PR target the next
major revision of Sufia (5.0)

### DIFF
--- a/sufia-models/app/models/collection.rb
+++ b/sufia-models/app/models/collection.rb
@@ -1,3 +1,0 @@
-class Collection < ActiveFedora::Base
-  include Sufia::Collection
-end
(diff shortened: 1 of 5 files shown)


mjgiarlo:
Don't merge this until the fedora-4/master branch is merged back into master.

PR #828, opened 2015-01-04
Users controller should not call an undefined method. Fixes #532.

### LINKED ISSUE #532: Users controller calls User#populate_attributes
but that method doesn't exist in Sufia. Move to scholarsphere?

mjgiarlo:
The default implementation is to return nil -- downstream implementations should populate user attributes in implementation- or organization-specific ways. E.g., ScholarSphere uses campus LDAP for this.


awead:
:+1:

mjgiarlo:
Explicit `nil` removed from method. Ping: @jcoyne

PR #830, opened 2015-01-07
removing collapse that makes the edit detail not display.

cam156:
We are mixing jquery slide toggle and bootstrap collapse and getting nothing.  Changed to use only slide toggle since that is what is being used in the dashboard.

### DIFF
--- a/app/assets/stylesheets/sufia/_batch-edit.scss
+++ b/app/assets/stylesheets/sufia/_batch-edit.scss
@@ -19,3 +19,6 @@
   padding-top: 5px;
 }
 
+.expanded-details {
+  display: none;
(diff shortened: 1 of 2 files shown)


jcoyne:
I already did something similar in the fedora-4/master branch, but I went the other way, using bootstrap and not jquery-slide toggle.

jcoyne:
Here's the fedora-4/master commit: https://github.com/projecthydra/sufia/commit/b5153862fa73bc4a71faa23df970f7e54d367b57

cam156:
@jcoyne I generated a new PR based on cherry picking your commit and resolving the conflicts.

PR #839, opened 2015-01-09
Changing the file query to be a solr query for loading the file id, so w...

cam156:
...e do not wait lareg file content to load.  This speed up the runs significantly on systems with large files or large numbers of files

### DIFF
--- a/sufia-models/lib/sufia/models/stats/user_stat_importer.rb
+++ b/sufia-models/lib/sufia/models/stats/user_stat_importer.rb
@@ -12,11 +12,11 @@ module Sufia
         start_date = date_since_last_cache(user)
 
         stats = {}
-        files_for_user(user).each do |file|
-          view_stats = FileViewStat.statistics(file.id, start_date, user.id)
+        file_ids_for_user(user).each do |file_id|
+          view_stats = FileViewStat.statistics(file_id, start_date, user.id)
           stats = tally_results(view_stats, :views, stats)
 
-          dl_stats = FileDownloadStat.statistics(file.id, start_date, user.id)
+          dl_stats = FileDownloadStat.statistics(file_id, start_date, user.id)
           stats = tally_results(dl_stats, :downloads, stats)
         end
 
@@ -38,8 +38,12 @@ private
       end
     end
 
-    def files_for_user(user)
-      ::GenericFile.where(Solrizer.solr_name('depositor', :symbol) => user.user_key)
+    def file_ids_for_user(user)
+      ids = []
+      ::GenericFile.find_in_batches("#{Solrizer.solr_name('depositor', :symbol)}:\"#{user.user_key}\"", fl:"id") do |group|
+        ids.concat group.map { |doc| doc["id"] }
+      end
+      ids
     end
 
     # For each date, add the view and download counts for this


sufia-models/lib/sufia/models/stats/user_stat_importer.rb
```diff
@@ -38,8 +38,16 @@ def date_since_last_cache(user)
       end
     end
 
-    def files_for_user(user)
-      ::GenericFile.where(Solrizer.solr_name('depositor', :symbol) => user.user_key)
+    def file_ids_for_user(user)
+      page_size = 1000
+      ids = []
+      resp = ActiveFedora::SolrService.instance.conn.get "select", params:{fq:["#{Solrizer.solr_name("has_model", :symbol)}:\"info:fedora/afmodel:GenericFile\"", "#{Solrizer.solr_name('depositor', :symbol)}:\"#{user.user_key}\""],fl:["id", "#{Solrizer.solr_name('depositor', :symbol)}"], rows: page_size}
```
jcoyne: Use `GenericFile.to_class_uri` instead of `info:fedora/afmodel:GenericFile`

jcoyne: Use `ActiveFedora::SolrService.query("", fq: ..., fl: ..., rows: ...)` instead

sufia-models/lib/sufia/models/stats/user_stat_importer.rb
```diff
@@ -38,8 +38,16 @@ def date_since_last_cache(user)
...
+      total = resp["response"]["numFound"]
+      while (ids.length < total) && (resp["response"]["docs"].count > 0)
+        ids.concat resp["response"]["docs"].map{|doc| doc["id"]}
+        resp = ActiveFedora::SolrService.instance.conn.get "select", params:{fq:["#{Solrizer.solr_name("has_model", :symbol)}:\"info:fedora/afmodel:GenericFile\"", "#{Solrizer.solr_name('depositor', :symbol)}:\"#{user.user_key}\""],fl:["id", "#{Solrizer.solr_name('depositor', :symbol)}"], rows: page_size, start: resp["response"]["start"]+page_size}
```
jcoyne: except for `start` you're duplicating what's on line 44. Can you store the options in a variable?

sufia-models/lib/sufia/models/stats/user_stat_importer.rb
jcoyne: Actually GenericFile.find_in_batches might be the best way, then you'd only need to pass in the depositor.

jcoyne: See: https://github.com/projecthydra/active_fedora/blob/d90da9bf3329ccc4f978c2d02d8567373adf13ab/lib/active_fedora/relation/finder_methods.rb#L146-L149

cam156:
@jcoyne find_in_batches is great!  Thanks for pointing me to that.

sufia-models/lib/sufia/models/stats/user_stat_importer.rb
```diff
@@ -38,8 +38,12 @@ def date_since_last_cache(user)
       end
     end
 
-    def files_for_user(user)
-      ::GenericFile.where(Solrizer.solr_name('depositor', :symbol) => user.user_key)
+    def file_ids_for_user(user)
+      ids = []
+      ::GenericFile.find_in_batches("#{Solrizer.solr_name('depositor', :symbol)}:\"#{user.user_key}\"",{fq:"#{Solrizer.solr_name("has_model", :symbol)}:\"#{::GenericFile.to_class_uri}\"",fl:["id", "#{Solrizer.solr_name('depositor', :symbol)}"]}) do |group|
+        ids.concat group.map{|doc| doc["id"]}
```
mjgiarlo: Style change suggestion:

```ruby
ids << group.map { |doc| doc["id"] }
```

cam156: @mjgiarlo According to ruby docs that code will create an array of arrays: http://www.ruby-doc.org/core-2.2.0/Array.html#method-i-3C-3C
That is definitely not what I want, unless I am missing something.
I could change it to a +, but that returns a new array, which is also not what I want, and slower.
Would you settle for my code with better spacing, or let me know why you code is better?

mjgiarlo: Oops! I confused #concat for #push. No, definitely stick with #concat instead of the shovel operator.  Spacing changes would be icing on the cake.  Thx!

cam156:
@mjgiarlo fixed up the spacing as we discussed in IRC.

mjgiarlo:
:+1:

sufia-models/lib/sufia/models/stats/user_stat_importer.rb
```diff
@@ -38,8 +38,12 @@ def date_since_last_cache(user)
       end
     end
 
-    def files_for_user(user)
-      ::GenericFile.where(Solrizer.solr_name('depositor', :symbol) => user.user_key)
+    def file_ids_for_user(user)
+      ids = []
+      ::GenericFile.find_in_batches("#{Solrizer.solr_name('depositor', :symbol)}:\"#{user.user_key}\"",{fq:"#{Solrizer.solr_name("has_model", :symbol)}:\"#{::GenericFile.to_class_uri}\"",fl:["id", "#{Solrizer.solr_name('depositor', :symbol)}"]}) do |group|
```
jcoyne: I don't think you need this: `fq:"#{Solrizer.solr_name("has_model", :symbol)}:\"#{::GenericFile.to_class_uri}\""`, the find_in_batches should already handle that.

jcoyne: Also, why bother with depositor in the `fl`?

cam156:
@jcoyne fixed it up!

sufia-models/lib/sufia/models/stats/user_stat_importer.rb
```diff
@@ -38,8 +38,12 @@ def date_since_last_cache(user)
       end
     end
 
-    def files_for_user(user)
-      ::GenericFile.where(Solrizer.solr_name('depositor', :symbol) => user.user_key)
+    def file_ids_for_user(user)
+      ids = []
+      ::GenericFile.find_in_batches("#{Solrizer.solr_name('depositor', :symbol)}:\"#{user.user_key}\"",{fl:["id"]}) do |group|
```
jcoyne: Just one more thing, I promise. You can do `fl: 'id'` there's no need to wrap that in an array. Then, you can also remove the curly braces since there's only one hash parameter to find_in_batches.

cam156:
@jcoyne I'm holding you to last change :)

PR #842, opened 2015-01-19
Make sure versions are sorted by the created field

hectorcorrea:
Code was still sorting by the old (date) field but the test failed to catch the error because there was only one item on the list used during the test.

### DIFF
--- a/app/presenters/sufia/version_list_presenter.rb
+++ b/app/presenters/sufia/version_list_presenter.rb
@@ -9,7 +9,7 @@ module Sufia
     private
 
       def wrapped_list
-        @wrapped_list ||= @raw_list.map { |v| Sufia::VersionPresenter.new(v) }.sort(&:date).tap { |l| l.first.try(:current!) }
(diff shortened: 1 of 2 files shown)


app/presenters/sufia/version_list_presenter.rb
```diff
@@ -9,7 +9,7 @@ def initialize(version_list)
     private
 
       def wrapped_list
-        @wrapped_list ||= @raw_list.map { |v| Sufia::VersionPresenter.new(v) }.sort(&:date).tap { |l| l.first.try(:current!) }
+        @wrapped_list ||= @raw_list.map { |v| Sufia::VersionPresenter.new(v) }.sort{|a,b| b.version.created <=> a.version.created}.tap { |l| l.first.try(:current!) }
```
mjgiarlo: Looks good, and can you tweak the whitespace for `sort`?  Like:

```ruby
...}.sort { |a, b| b.version.created <=> a.version.created }.tap...
```

hectorcorrea: Done.

mjgiarlo:
:+1:

PR #843, opened 2015-01-19
Don't show collections in shared files list.  Story scholarsphere#48

val99erie:


spec/controllers/my/shares_controller_spec.rb
```diff
@@ -24,6 +24,10 @@
           r.edit_users += [@other_user.user_key]
           r.save!
         end
+        @my_collection = Collection.new(title: "My collection").tap do |c|
+          c.apply_depositor_metadata(@user.user_key)
+        end
+        @my_collection.save!
       end
```
cam156: @val99erie I know this is unrelated to the fix you added, but we should change this over to the let syntax and clean up the objects after the test completes.  This test is leaving files (and now a collection) behind that could corrupt other tests.

spec/controllers/my/shares_controller_spec.rb
```diff
@@ -2,29 +2,48 @@
...
-        @other_user = FactoryGirl.create(:user)
-        @my_file = FactoryGirl.create(:generic_file, depositor: @user)
-        @unshared_file = FactoryGirl.create(:generic_file, depositor: @other_user)
-        @shared_with_me = FactoryGirl.create(:generic_file).tap do |r|
-          r.apply_depositor_metadata @other_user
-          r.edit_users += [@user.user_key]
+      end
+
+      after do
+        GenericFile.destroy_all
```
jcoyne: These are unnecessary on the fedora-4 branch, and cause the test suite to run slower. We use ActiveFedora::Cleaner instead. Please remove them.

mjgiarlo:
:+1:

PR #846, opened 2015-01-20
Permit the parameters that allow you to remove a permission

jcoyne:
Previously, attempting to remove a permission would fail because
permissions_attributes id and _destroy were not permitted


mjgiarlo:
:clap:

PR #847, opened 2015-01-20
Removed a rescue block

jcoyne:
Reasoning:
  * It was an untyped, so it's hard to say which conditions might be
expected. This means it was obfuscating errors we should be solving like #846
  * It was untested
  * It didn't work. It didn't set the correct ivars to render the edit view


val99erie:
I agree that this rescue block should be removed.  Those missing ivars are part of the problem for redmine bug 9713.

mjgiarlo:
:+1:

PR #849, opened 2015-01-20
Making test more flexible so it should work with different version of ra...

cam156:
...ils

### DIFF
--- a/spec/lib/sufia/messages_spec.rb
+++ b/spec/lib/sufia/messages_spec.rb
@@ -16,10 +16,10 @@ describe Sufia::Messages do
     Object.send(:remove_const, :TestClass)
   end
 
-  let(:batch_id)  { "1" }
-  let(:single)    { double(noid: "1", to_s: "File 1") }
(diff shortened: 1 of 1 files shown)


spec/lib/sufia/messages_spec.rb
```diff
@@ -31,30 +31,34 @@ class TestClass
   end
 
   describe "#single_success" do
-    let(:expected) { '<span id="ss-1"><a href="/files/1">File 1</a> has been saved.</span>' }
     it "should render a success message for a single file" do
-      expect(message.single_success(batch_id, single)).to eq(expected)
+      node = Capybara::Node::Simple.new(message.single_success(batch_id, single))
+      expect(node).to have_selector("span[id=\"ss-1\"]", text:"File 1 has been saved.")
```
mjgiarlo: You don't need to change this here, but for future reference you could use single quotes here to prevent having to escape the double quotes.

mjgiarlo:
Nice :+1:

PR #855, opened 2015-01-21
Disabled invalid option in the form.  Story 9717

val99erie:
The controller error was already fixed as part of story 9713.  I just made a change to the form to disable the radio button option for the current version so that the user cannot submit an invalid version.

### DIFF
--- a/app/assets/stylesheets/sufia/_styles.scss
+++ b/app/assets/stylesheets/sufia/_styles.scss
@@ -82,6 +82,10 @@ label.required, span.required {
   }
 }
 
+label.disabled {
+  color: $gray-med-dark;
+}
+
 @media only screen and (min-width: 768px) {
   #masthead {
     min-height: 8em;
--- a/app/views/generic_files/_versioning.html.erb
+++ b/app/views/generic_files/_versioning.html.erb
@@ -21,9 +21,10 @@
     <% @version_list.each do |version| %>
       <div class="form-group">
         <div class="col-xs-12">
-          <label class="radio">
-            <%= radio_button_tag(:revision, version.label, version.current?) %>
-            Restore from <%= version.created %> [by <%= version.committer %>]
+          <label class="radio <%= version.current? ? 'disabled' : '' %>">
+            <%= radio_button_tag(:revision, version.label, false, disabled: version.current?, required: 'required') %>
+            <%= version.current? ? 'Current version:' : 'Restore from' %>
+            <%= version.created %> [by <%= version.committer %>]
           </label>
         </div>
       </div>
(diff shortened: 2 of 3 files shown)


app/views/generic_files/_versioning.html.erb
```diff
@@ -1,33 +1,38 @@
 <div id="versioning_display" class="tab-pane">
   <h2>Versions</h2>
   <div class="well">
+
     <%= form_for [sufia, @generic_file], html: { multipart: true } do |f| %>
       <%= hidden_field_tag('redirect_tab', 'versions') %>
-    <h3>Upload New Version</h3>
```
cam156: Is there are reason you are swapping the order of the forms?

jcoyne:
@val99erie Looks like you need to rebase so this can merge cleanly.

app/assets/stylesheets/sufia/_styles.scss
```diff
@@ -82,6 +82,10 @@ label.required, span.required {
   }
 }
 
+label.disabled {
+  color: #666;
```
jcoyne: http://youtu.be/q9kT37eIkaY?t=1m49s

awead: :smiling_imp:

jcoyne: I wonder if `$input-bg-disabled` would be better than hard coding this.

jcoyne: Or `$gray-light`

val99erie:
@jcoyne - I added a new gray color.  The existing $gray-medium was too light to see against the gray background, and the $gray-dark was indistinguishable from the text for the non-disabled options.

app/views/generic_files/_versioning.html.erb
```diff
@@ -21,9 +21,11 @@
     <% @version_list.each do |version| %>
       <div class="form-group">
         <div class="col-xs-12">
-          <label class="radio">
-            <%= radio_button_tag(:revision, version.label, version.current?) %>
-            Restore from <%= version.created %> [by <%= version.committer %>]
+          <% disable_if_current_version = version.current? ? 'disabled' : '' %>
+          <label class="radio <%= disable_if_current_version %>">
```
cam156: Not sure I like the layer of indirection here.  Why not just use version.current? ? 'disabled' : '' in the label.  disable_if_current_version is only being used here correct?

val99erie: @cam156 You are correct.  I'll change it.

jcoyne:
:+1:

PR #858, opened 2015-01-22
Bumping to require the latest version of hydra-collections to get the de...

cam156:
...positor indexed into the collection


mjgiarlo:
:+1: if Travis :sunny:

PR #860, opened 2015-01-22
NOIDs should not be second-class citizens.

mjgiarlo:
Prior to Fedora 4, IDs in Hydra resembled Fedora identifiers ('namespace:id'), and in Sufia we had two opinions about this: 1) the namespace should never be shown to users; 2) the default ID in Fedora 3, a sequential integer, should be overridden by a configurable NOID. In Fedora 4, IDs no longer have a namespace, and so rather than continue to have two kinds of identifiers floating around the system, we have just one: the object identifier, which takes the form of a NOID. This PR eliminates nearly all references to noids in the code (since they're just object identifiers), and makes usage of NOIDs configurable. If you set Sufia.config.enable_noids to false, you will see Fedora 4's UUIDs instead of NOIDs.


sufia-models/app/services/sufia/indexing_service.rb
```diff
@@ -1,15 +1,4 @@
 module Sufia
   class IndexingService < ActiveFedora::IndexingService
```
jcoyne: Do we need this class anymore?

mjgiarlo: Nope.

sufia-models/lib/generators/sufia/models/templates/config/sufia.rb
```diff
@@ -105,8 +105,11 @@
   # If you have ffmpeg installed and want to transcode audio and video uncomment this line
   # config.enable_ffmpeg = true
 
-  # Specify the Fedora pid prefix:
-  # config.id_namespace = "sufia"
+  # To disable minting of NOIDs for files and collections in favor of Fedora UUIDs
+  # config.enable_noids = false
```
jcoyne: This should be `true` so it matches the default.

sufia-models/app/services/sufia/indexing_service.rb
mjgiarlo: Removed in latest push.

sufia-models/lib/generators/sufia/models/templates/config/sufia.rb
mjgiarlo: Fixed in latest push.

sufia-models/app/services/sufia/generic_file_indexing_service.rb
```diff
@@ -1,5 +1,5 @@
 module Sufia
-  class GenericFileIndexingService < IndexingService
+  class GenericFileIndexingService
```
jcoyne: Should extend `ActiveFedora::IndexingService`

sufia-models/lib/generators/sufia/models/templates/config/sufia.rb
```diff
@@ -105,8 +105,11 @@
   # If you have ffmpeg installed and want to transcode audio and video uncomment this line
   # config.enable_ffmpeg = true
 
-  # Specify the Fedora pid prefix:
-  # config.id_namespace = "sufia"
+  # To disable minting of NOIDs for files and collections in favor of Fedora UUIDs
```
jcoyne: Change text to `Sufia uses NOIDs for files and collections instead of Fedora UUIDs`

sufia-models/app/services/sufia/generic_file_indexing_service.rb
mjgiarlo: Derp.  Done.

sufia-models/lib/generators/sufia/models/templates/config/sufia.rb
mjgiarlo: fixed with latest push.

jcoyne:
:heavy_plus_sign: :100:  I'm :grinning:

app/helpers/sufia/sufia_helper_behavior.rb
```diff
@@ -38,7 +38,7 @@ def sufia_thumbnail_tag(document, options)
       # file
       else
         path = if document.image? || document.pdf? || document.video? || document.office_document?
-          sufia.download_path document.noid, datastream_id: 'thumbnail'
+          sufia.download_path document, datastream_id: 'thumbnail'
         elsif document.audio?
```
cam156: Should this be document.id?

cam156: NM I see you are consistently not using id in path helpers

app/helpers/generic_file_helper.rb
```diff
@@ -18,7 +18,7 @@ def render_download_icon title = nil
   end
 
   def render_download_link text = nil
-    link_to (text || "Download"), sufia.download_path(@generic_file.noid), { id: "file_download", target: "_new", data: { label: @generic_file.id } }
+    link_to (text || "Download"), sufia.download_path(@generic_file.id), { id: "file_download", target: "_new", data: { label: @generic_file.id } }
   end
```
cam156: This should not have id to match up with you other uses of path helpers

app/views/collections/_sort_and_per_page.html.erb
```diff
@@ -12,7 +12,7 @@
    <div class="sort-toggle">
      <%# kind of hacky way to get this to work on catalog and folder controllers.  May be able to simple do {action: "index"} but I'm not sure -%>
      <% unless @response.response['numFound'] < 2 %>
-        <%= form_tag collections.collection_path(@collection.noid), method: :get, class: 'per_page form-inline' do %>
+        <%= form_tag collections.collection_path(@collection.id), method: :get, class: 'per_page form-inline' do %>
           <%= label_tag(:sort, "<span>Sort By:</span>".html_safe) %>
```
cam156: I think just passing the collection here to match with other paths might be good

app/views/homepage/_recent_document.html.erb
```diff
@@ -9,7 +9,7 @@
       <% end %>
       <td>
         <h3>
-          <span class="sr-only">Title</span><%= link_to truncate(recent_document.title_or_label, length: 28, separator: ' '), sufia.generic_file_path(recent_document.noid)  %>
+          <span class="sr-only">Title</span><%= link_to truncate(recent_document.title_or_label, length: 28, separator: ' '), sufia.generic_file_path(recent_document.id)  %>
           <% if display_access %>
```
cam156: Here too...

app/views/homepage/_recent_document.html.erb
```diff
@@ -22,7 +22,7 @@
         </h3>
         <p>
           <span class="sr-only">Filename</span><%= link_to truncate(recent_document.label, length: 25),
-              sufia.generic_file_path(recent_document.noid), title: recent_document.label %><br />
+              sufia.generic_file_path(recent_document.id), title: recent_document.label %><br />
```
cam156: And Here

spec/models/solr_document_spec.rb
```diff
@@ -11,15 +11,6 @@
     end
   end
 
-  describe "to_param" do
```
cam156: Should this really be removed completely or should it check to see that id was assigned

sufia-models/app/jobs/resolrize_job.rb
```diff
@@ -4,13 +4,6 @@ def queue_name
   end
 
   def run
-    require 'active_fedora/version'
```
cam156: Is this really related to this commit?  Seems unrelated to me

app/helpers/sufia/sufia_helper_behavior.rb
mjgiarlo: You're right, I'm relying on `#to_param` magic rather than explicitly specifying the `#id` method. (And whether you use UUIDs or NOIDs, `#to_param` will return the object's identifier.)

app/helpers/generic_file_helper.rb
mjgiarlo: Good catch!

app/views/collections/_sort_and_per_page.html.erb
mjgiarlo: Agreed.

app/views/homepage/_recent_document.html.erb
mjgiarlo: Thx!

app/views/homepage/_recent_document.html.erb
mjgiarlo: I detect a trend.

spec/models/solr_document_spec.rb
mjgiarlo: Nope, it shouldn't be -- this was from an earlier run through the code when I excised that completely. I'll add the test back.

sufia-models/app/jobs/resolrize_job.rb
mjgiarlo: resolrize_job was basically broken in its current form because it reindexed with an (F3-style) assumption that identifiers start with PIDs. This change makes reindexing work again under F4.

app/helpers/sufia/sufia_helper_behavior.rb
```diff
@@ -154,8 +154,8 @@ def search_form_action
     end
 
     def render_visibility_link document
-      link_to render_visibility_label(document), sufia.edit_generic_file_path(document.noid, {anchor: "permissions_display"}),
-        id: "permission_"+document.noid, class: "visibility-link"
+      link_to render_visibility_label(document), sufia.edit_generic_file_path(document.id, {anchor: "permissions_display"}),
```
awead: but your'e using `id` in this path. Should it just be `document` as above in @cam156 comment?

lib/generators/sufia/templates/catalog_controller.rb
```diff
@@ -124,7 +124,7 @@ def self.modified_field
       all_names = config.show_fields.values.map{|val| val.field}.join(" ")
       title_name = solr_name("title", :stored_searchable)
       field.solr_parameters = {
-        qf: "#{all_names} noid_tsi file_format_tesim all_text_timv",
+        qf: "#{all_names} file_format_tesim all_text_timv",
```
hectorcorrea: I assume it is intentional here not to pass "id" in the "qf" parameter. Right?

app/views/collections/_show_document_list_row.html.erb
```diff
@@ -1,10 +1,10 @@
-<% noid = document.noid %>
-<tr id="document_<%= noid %>" class="<%= cycle("", "zebra") %>">
+<% id = document.id %>
```
awead: Couldn't this just be `document` and avoid the extra variable setting?

spec/services/noid_spec.rb
```diff
@@ -1,16 +1,9 @@
...
-      let(:id) { 'abc123' }
-      it { is_expected.to eq 'sufia:abc123' }
-    end
-    context "when the passed in pid has a namespace" do
-      let(:id) { 'ksl:abc123' }
-      it { is_expected.to eq 'ksl:abc123' }
-    end
+  describe "#treeify" do
+    subject { Sufia::Noid.treeify(id) }
+    let(:id) { 'abc123def45' }
```
awead: Is there a min or max on the number of characters in the noid that Sufia will "treeify" ?

app/views/collections/_show_document_list_row.html.erb
jcoyne: @awead No, because it's used to make a string for the `id` attribute.

sufia-models/lib/generators/sufia/models/templates/config/sufia.rb
```diff
@@ -105,8 +105,11 @@
   # If you have ffmpeg installed and want to transcode audio and video uncomment this line
   # config.enable_ffmpeg = true
 
-  # Specify the Fedora pid prefix:
-  # config.id_namespace = "sufia"
+  # Sufia uses NOIDs for files and collections instead of Fedora UUIDs
```
hectorcorrea: It might be good to clarify that NOIDs are in the form "nnnn" where as Fedora UUIDs are in the form "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx". Not sure how much common knowledge this is.

sufia-models/app/jobs/resolrize_job.rb
cam156: :ok_hand:

lib/generators/sufia/templates/catalog_controller.rb
mjgiarlo: Yep! The ID is automatically included in the solr document.

app/views/collections/_show_document_list_row.html.erb
mjgiarlo: True, though I did miss a bunch of to_param opportunities in this file, so thanks for raising this.

spec/services/noid_spec.rb
mjgiarlo: https://github.com/projecthydra/sufia/blob/master/sufia-models/app/services/sufia/noid.rb#L27

sufia-models/lib/generators/sufia/models/templates/config/sufia.rb
mjgiarlo: done!

mjgiarlo:
Testing latest round of changes, then pushing.

PR #873, opened 2015-01-23
Resource type should not be required

jcoyne:
It was not required in previous versions of Sufia, so this ensures the behavior doesn't change unnecessarily.

### DIFF
--- a/app/views/batch/_metadata.html.erb
+++ b/app/views/batch/_metadata.html.erb
@@ -28,14 +28,14 @@
       <div class="col-sm-6">
       <h3>Applies to all files just uploaded</h3>
       <%= f.input :resource_type, as: :select_with_help, collection: Sufia.config.resource_types,
-          input_html: { class: 'form-control', multiple: true, required: true } %>
+          input_html: { class: 'form-control', multiple: true } %>
 
-      <%= f.input :tag, as: :multi_value_with_help, input_html: { required: true } %>
+      <%= f.input :tag, as: :multi_value_with_help %>
 
-      <%= f.input :creator, as: :multi_value_with_help, input_html: { required: true } %>
+      <%= f.input :creator, as: :multi_value_with_help %>
 
       <%= f.input :rights, as: :select_with_modal_help, collection: Sufia.config.cc_licenses,
-          input_html: { class: 'form-control', multiple: true, required: true } %>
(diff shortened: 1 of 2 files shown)


mjgiarlo:
Is this change related or unrelated to build breakage?

mjgiarlo:
Restarted the build.

app/views/batch/_metadata.html.erb
```diff
@@ -28,14 +28,14 @@
       <div class="col-sm-6">
       <h3>Applies to all files just uploaded</h3>
       <%= f.input :resource_type, as: :select_with_help, collection: Sufia.config.resource_types,
-          input_html: { class: 'form-control', multiple: true, required: true } %>
+          input_html: { class: 'form-control', multiple: true } %>
 
-      <%= f.input :tag, as: :multi_value_with_help, input_html: { required: true } %>
+      <%= f.input :tag, as: :multi_value_with_help %>
 
-      <%= f.input :creator, as: :multi_value_with_help, input_html: { required: true } %>
```
cam156: Creator has been a required field.

app/views/batch/_metadata.html.erb
```diff
@@ -28,14 +28,14 @@
       <div class="col-sm-6">
       <h3>Applies to all files just uploaded</h3>
       <%= f.input :resource_type, as: :select_with_help, collection: Sufia.config.resource_types,
-          input_html: { class: 'form-control', multiple: true, required: true } %>
+          input_html: { class: 'form-control', multiple: true } %>
 
-      <%= f.input :tag, as: :multi_value_with_help, input_html: { required: true } %>
+      <%= f.input :tag, as: :multi_value_with_help %>
```
cam156: Tag has been a required field.

app/views/batch/_metadata.html.erb
jcoyne: But that incantation does not affect whether it shows on the form as required or not. The input sets the required attribute if the field is required on the form. See: https://github.com/projecthydra/sufia/blob/master/app/forms/sufia/forms/generic_file_edit_form.rb#L6

app/views/batch/_metadata.html.erb
jcoyne: it still is.

PR #875, opened 2015-01-26
Don't show update order unless there are featured works. Fixes #865

### LINKED ISSUE #865: When no works have been featured and 'Save Order' is clicked an error is raised
As an admin user on the home page, given that no works have been featured, when 'Save Order' is clicked, results in an error page

jcoyne:


mjgiarlo:
Should this also include a view spec that ensures the button isn't there, or do folks think this is sufficiently well tested already?

jcoyne:
@mjgiarlo what do you think?

mjgiarlo:
I think it'd be good to see one.

jcoyne:
@mjgiarlo added some tests.

mjgiarlo:
Failed.

PR #879, opened 2015-01-27
Redirects to 'my shares' when submiting files on behalf of somebody else

hectorcorrea:


spec/controllers/batch_controller_spec.rb
```diff
@@ -22,6 +22,23 @@
       end
     end
 
+    describe "when submiting files on behalf of another user" do
+      let(:somebody_else_file) do 
+        f = GenericFile.create(title: ['Original Title'])
+        f.apply_depositor_metadata('current_user')
+        f.save!
+        f.on_behalf_of = 'somebody@else.org'
```
hectorcorrea: I couldn't figure out a way to set `on_behalf_of` without the extra save. Setting it as a parameter in the call to `create` or in between `apply_depositor_metadata` and `save` throws error `NoMethodError: undefined method `log_event' for nil:NilClass`

spec/controllers/batch_controller_spec.rb
```diff
@@ -22,6 +22,23 @@
       end
     end
 
+    describe "when submiting files on behalf of another user" do
+      let(:somebody_else_file) do 
+        f = GenericFile.create(title: ['Original Title'])
```
jcoyne: Try this and avoid an extra save: 
```ruby
f = GenericFile.create(title: ['Original Title']) do |gf|
  gf.apply_depositor_metadata 'current_user'
end
```

hectorcorrea: That looks nicer. I've updated the code.

spec/controllers/batch_controller_spec.rb
jcoyne: I think you were getting that error because it couldn't find the user with the email: `somebody@else.org`

app/controllers/batch_controller.rb
```diff
@@ -32,5 +36,12 @@ def initialize_fields(file)
     file.initialize_fields
   end
 
+  def uploading_on_behalf_of? batch
+    return false if batch.generic_files.count == 0  
+    on_behalf_of = batch.generic_files[0].on_behalf_of
+    return false if on_behalf_of == nil or on_behalf_of.empty?
```
jcoyne: how about `if on_behalf_of.blank?`

app/controllers/batch_controller.rb
```diff
@@ -32,5 +36,12 @@ def initialize_fields(file)
...
+    current_user.email != batch.generic_files[0].on_behalf_of
```
jcoyne: You never want to hard_code email, because some projects use a different user identifier (such as Scholarsphere). Use `current_user.user_key` instead.

app/controllers/batch_controller.rb
hectorcorrea: Done.

app/controllers/batch_controller.rb
hectorcorrea: Done.

app/controllers/batch_controller.rb
```diff
@@ -32,5 +36,10 @@ def initialize_fields(file)
     file.initialize_fields
   end
 
+  def uploading_on_behalf_of? batch
+    return false if batch.generic_files.count == 0 or batch.generic_files[0].on_behalf_of.blank?
```
jcoyne: Doing `batch.generic_files.count` will cause an extra count query to solr.  Instead can we do this?
```ruby
exemplar = batch.generic_files.first
return false if exemplar.nil? || exemplar.on_behalf_of.blank?
current_user.user_key != exemplar.on_behalf_of
```

hectorcorrea: Good catch. I've updated the code.

PR #882, opened 2015-01-29
Store old entries for featured researcher.  Redmine story #9146

val99erie:


app/controllers/content_blocks_controller.rb
```diff
@@ -1,8 +1,24 @@
...
+
   def update
-    @content_block.update(params.require(:content_block).permit(:value))
+    @content_block.update(update_params)
     redirect_to :back
   end
+
+protected
+
+  def create_params
```
mjgiarlo: Where is this method called?

config/locales/sufia.en.yml
```diff
@@ -1,6 +1,7 @@
 en:
   sufia:
     account_name:       "My Institution Account Id"
+    account_label:      "User"
```
mjgiarlo: Where is this key used?

app/controllers/content_blocks_controller.rb
jcoyne: Yeah, cancancan calls it.

mjgiarlo:
Looks very good.  I just have a couple questions, which I left in the code.

The other thing I might like to see in this PR is a generator that does the right thing for folks upgrading rather than installing anew. (IOW, it might be nice for the release that contains this code to have a step that takes care of this via, e.g., `rails g sufia:featured` or some such.

app/controllers/content_blocks_controller.rb
mjgiarlo: OK, good.

config/locales/sufia.en.yml
val99erie: Actually, this isn't related to this story.  I just noticed that it was missing when I was looking at the permissions tab when editing a generic file:
https://github.com/projecthydra/sufia/blob/master/app/views/generic_files/_permission_form.html.erb#L44

I should probably take it out of this PR and submit it separately.

mjgiarlo: You can leave it in, 'sok.  Thanks, @val99erie

mjgiarlo:
I'm :+1: on this. Comments from others? @jcoyne @cam156 etc.

val99erie:
I could add a separate generator, but it wouldn't only cover the featured_researchers.  The content_blocks table contains entries for the "marketing text" and the "about" page too.  For this PR, I only added a column onto an existing table.

Should I make a generator for all types of content_blocks?  Or should the generator assume that the content_blocks table already exists?

mjgiarlo:
I believe the upgrade400 and install generators already create the content blocks. I think you can safely assume that table already exists

app/helpers/content_block_helper.rb
```diff
@@ -1,22 +1,42 @@
...
+        concat edit_button(content_block)
+        concat new_button(content_block) if show_new
       } 
-      concat form_for([sufia, content_block] ) { |f|
-        concat f.text_area :value, id: "text_area_#{content_block.name}", class: "tinymce", rows: 20, cols: 120
-        concat f.submit 'Save', class: "btn btn-primary"
-      }
-
+      concat edit_form(content_block)
+      concat new_form(content_block.name) if show_new
```
cam156: @val99erie Should this be a save  here instead of an edit?

cam156: NM:  My mistake

mjgiarlo:
:+1:

PR #884, opened 2015-01-29
Use the latest versions of hydra-head

jcoyne:


spec/controllers/downloads_controller_spec.rb
```diff
@@ -5,10 +5,10 @@
   describe "with a file" do
     let(:depositor) { FactoryGirl.find_or_create(:archivist) }
     let(:file) do
-      GenericFile.new.tap do |f|
+      GenericFile.create do |f|
         f.apply_depositor_metadata(depositor.user_key)
-        f.add_file(File.open(fixture_path + '/world.png'), 'content', 'world.png')
-        f.save!
+        f.label = 'world.png'
```
mjgiarlo: Can you say more about this addition?

mjgiarlo: We used to have a method that took care of setting the label and a default title, and I see the actor now does this itself in create_content. (And is "label" even a thing in Fedora 4?  I thought was a Fedora 3 built-in.)

jcoyne: Label was a fedora 3 builtin. We've replaced it with a simple predicate. `add_file` in hydra-head used to set that label, but we've deprecated that module so now we're doing it manually.

mjgiarlo: :+1:

app/jobs/ingest_local_file_job.rb
```diff
@@ -20,12 +20,13 @@ def run
     path = File.join(directory, filename)
 
     generic_file.label = File.basename(filename)
-    generic_file.add_file(File.open(path), 'content', generic_file.label)
+    generic_file.add_file(File.open(path), path: 'content', original_name: generic_file.label, mime_type: mime_type(filename))
     generic_file.record_version_committer(user)
     generic_file.save!
```
mjgiarlo: Should we be using Actor#create_content here?

jcoyne: Yes we should (See https://github.com/projecthydra/sufia/blob/latest_hydra_head/app/jobs/ingest_local_file_job.rb#L15)  I took a stab at that, but I was getting too many unrelated issues, so I backed it out.

mjgiarlo: OK. I'll merge this and put in an issue.

PR #887, opened 2015-01-30
refactoring single_use_links_controller and single_use_links_viewer_cont...

weiweishi:
...roller into behavior modules


app/controllers/single_use_links_viewer_controller.rb
```diff
@@ -17,53 +14,7 @@ def initialize(user, single_use_link)
       can :read, ActiveFedora::Base do |obj|
         single_use_link.valid? && single_use_link.itemId == obj.id && single_use_link.destroy!
       end if single_use_link
-    end
```
cam156: @weiweishi Just curious, why did you leave the Ability here instead of in the behavior?

cam156:
@weiweishi It looks like your PR need to be rebased.  Also do you have a signed license agreement with hydra?

mjgiarlo:
@weiweishi does have an active CLA.

app/controllers/single_use_links_viewer_controller.rb
mjgiarlo: Hmmm, we might want to put this in the behavior like @cam156 suggested, but can we also put it behind a method that downstream adopters can override? I suspect that's why @weiweishi left this out of the controller, so that this isn't foisted on all adopters.

mjgiarlo:
@weiweishi for the future, you don't have to close the PR to make changes to it!  just make changes in your branch locally, squash commits as necessary, force-push, and the PR will be updated.

weiweishi:
Thanks @cam156 and @mjgiarlo. 

It's a good idea to wrap the Ability into a method, instead of a Class, so that it can be overridden if needed. I didn't move it because it's not a "behavior". I'll work on that. 

I have some trouble with merging the master branch in our local fork and being a new adopter of Git and so few changes to the fileset, it seems easier to start fresh then trying to untangle the mess.  This ride should be much smoother next time around.

weiweishi:
Do you guys see a problem to add this Ability to Sufia::Ability?

PR #892, opened 2015-01-30
Refactor single use links

weiweishi:


sufia-models/app/models/concerns/sufia/ability.rb
```diff
@@ -45,10 +46,16 @@ def stats_abilities
       alias_action :stats, to: :read
     end
 
+    def single_use_links_abilities
+      can :read, SingleUseLink
```
jcoyne: This says, every user can read every SingleUseLink.  I'm quite certain that wasn't the previous behavior.

app/controllers/single_use_links_viewer_controller.rb
```diff
@@ -1,69 +1,3 @@
...
-    def initialize(user, single_use_link)
-      @user = user || User.new
-
-      @single_use_link = single_use_link
-
-      can :read, ActiveFedora::Base do |obj|
-        single_use_link.valid? && single_use_link.itemId == obj.id && single_use_link.destroy!
-      end if single_use_link
-    end
-  end
```
weiweishi: Original behavior is here. With my limited understanding of CanCan, it seems to be as long as the single_use_link exists and valid, any current user or new user would be able to read. Not too sure how to validate an instance of single_use_link from ability though. Still reading on Cancan.

mjgiarlo: @weiweishi Why not move the Ability class back into the controller like before, and wipe out your changes to ability.rb?

PR #895, opened 2015-01-30
refactoring single use links

weiweishi:


jcoyne:
:+1: Looks good. Thanks.

mjgiarlo:
@jcoyne If the viewer controller behavior depends on SingleUseLinksViewerController::Ability.new, will this trip up downstream adopters?  Currently the controller is dependent upon the behavior and the behavior is also dependent upon the controller.  If a downstream adopter wants to alter the ability, is the pattern for them to define their own SingleUseLinksViewerController::Ability class, and the behavior will automagically pick that one up rather than the one in Sufia?

jcoyne:
Would anyone want a different behavior for a single-use-link? that Ability seems pretty fundamental to what a single-use-link is. they could always skip the "`authorize_resource` callback if necessary.

PR #900, opened 2015-02-06
Adding an overall timeout for the derivatives and a task timeout for ind...

cam156:
...ividual derivative tasks so that the entire derivatives process does not take longer a certain amount of time, and each task can be set enmass instead of needing to be configured individually

In ScholarSphere we are seeing derivatives jobs just hang on the queue.  It may be related to deployment, or stopping and starting the jobs queue, but the jobs should not run forever.  It causes other jobs to back up behind.  This sets a overall timeout for creating any derivatives and include configuring an over all timeout for derivative tasks.  

Part of the reason I am doing this at this level is there is no timeout in derivatives for images.  Yesterday an image file hung on the queue for more than an hour before I found it and killed it.  When I reran it it took less than a minute to run.

### DIFF
--- a/sufia-models/app/jobs/create_derivatives_job.rb
+++ b/sufia-models/app/jobs/create_derivatives_job.rb
@@ -8,7 +8,11 @@ class CreateDerivativesJob < ActiveFedoraPidBasedJob
     if generic_file.video?
       return unless Sufia.config.enable_ffmpeg
     end
-    generic_file.create_derivatives
+    status = Timeout::timeout(Sufia.config.derivatives_timeout) do
+      generic_file.create_derivatives
+    end
     generic_file.save
+  rescue Timeout::Error => ex
+    raise Hydra::Derivatives::TimeoutError, "Unable to generate derivatives for \"#{generic_file.id}\"\nThe derivatives generation took longer than #{Sufia.config.derivatives_timeout} seconds to execute"
   end
 end
--- a/spec/jobs/create_derivatives_job_spec.rb
+++ b/spec/jobs/create_derivatives_job_spec.rb
@@ -171,4 +171,22 @@ describe CreateDerivativesJob do
       end
     end
   end
+  context "with timeout set very small" do
+    let(:mime_type) { 'audio/wav' }
+    let(:file_name) { 'piano_note.wav' }
+    before do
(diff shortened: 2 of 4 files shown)


mjgiarlo:
Failures in all three builds, sadly.

cam156:
Foo!  The wrong thing got sent up...  I'll fix it!

cam156:
@mjgiarlo Tis passing now...

mjgiarlo:
@cam156 ++

sufia-models/app/jobs/create_derivatives_job.rb
```diff
@@ -8,7 +8,11 @@ def run
     if generic_file.video?
       return unless Sufia.config.enable_ffmpeg
     end
-    generic_file.create_derivatives
+    status = Timeout::timeout(Sufia.config.derivatives_timeout) do
```
mjgiarlo: Wait, doesn't hydra-derivatives already provide this?  Like, can't you just set:

```ruby
Hydra::Derivatives::Image::Processor.timeout = Sufia.config.derivatives_timeout
````

?

cam156: HydraDerivatives sets a timeout for each item that it creates, unfortunately image processing hung and that does not have a timeout option.  Additionally I just want to be certain the entire derivatives generation does not go on forever. The timeouts are way down in the weeds in hydra derivatives, so other portions could hang and those timeouts may not catch the thing hanging.

cam156: We are using the latest hydra derivatives and sufia and still seeing jobs lasting forever.  It is causing the entire system to seem like it is broken.  I want to have some control over something that is causing the system to break.

jcoyne:
I think it makes sense to have different timeouts per each type of derivative. One expects video compression to take a lot longer than creating video derivatives.  Why don't you just set an Image derivative timeout in Scholarsphere?

mjgiarlo:
Trying to solve this in hydra-derivatives instead. Thoughts, @cam156 @jcoyne?

https://github.com/projecthydra-labs/hydra-derivatives/pull/51

cam156:
@jcoyne and @mjgiarlo I'm willing to try just a video timeout, and add the overall timeout into ScholarSphere later if it is needed.  The derivatives hanging means that no thumbnails get generated for anything, so I thought it might affect others as well.  I am not 100% sure that the timeout is at the individual task level, which was why I put the timeout at the top level.

PR #902, opened 2015-02-11
Move bl-gallery generator from sufia-models generator to the sufia generator

jcoyne:


cam156:
@jcoyne Looks good.  What is the reason for this change?

jcoyne:
@cam156 because blacklight-gallery is presentation, not model.  Furthermore, sufia-models does not require blacklight-gallery, so if you're just using sufia-models, the generator fails.

PR #908, opened 2015-02-14
If RAILS_VERSION is not set use sass-rails 5

jcoyne:


cam156:
@jcoyne Not sure I get why we need this change?  Is there a newer version of Rails that is compatible?

jcoyne:
@cam156 this allows it to work with Rails 4.2 without setting RAILS_VERSION

PR #910, opened 2015-02-16
Adding a rescue for a TimeoutError as we are seeing those in production

cam156:


spec/jobs/sufia_resque_queue_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+  context "no errors" do
+    it "queues the job" do
+      expect(::Resque).to receive(:enqueue_to).and_return(true)
+      subject.push(job)
+    end
+  end
+
+  context "timeout error" do
+    it "retries on timeout" do
+      expect(::Resque).to receive(:enqueue_to).and_raise(Redis::TimeoutError)
```
mjgiarlo: It looks here like you may be conflating stubs with tests here.  I would expect these tests to set the right stubs in before blocks that trigger timeouts and successful runs and then the tests to make sure the behavior at a higher layer is as expected.

cam156: @mjgiarlo   I was expecting the enqueue_to to be called twice once it would raise and error and the next time it would go through.  The test is that it gets called a second time.  So I would see the first call being an allow, but the second call is definitely and expect.

If you would like to rewrite be my guest.

cam156: @mjgiarlo check out the rspec readme https://github.com/rspec/rspec-mocks/blob/master/README.md#setting-responses
There is an example of the syntax I used there...

mjgiarlo: @cam156 the syntax is A-OK!

Why would the job pushed on line 18 be expected to raise an error if the job on line 10 didn't?  I know the tests pass but I have a sinking suspicion here that the tests may not be doing what you want them to do.

mjgiarlo: I'll take a look.

mjgiarlo:
OK, @cam156, I pushed my changes to tease apart stubs from message expectations. Feedback welcome!

spec/jobs/sufia_resque_queue_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+      expect(::Resque).to receive(:enqueue_to).once.and_return(true)
+      subject.push(job)
+    end
+  end
+
+  context 'when one run times out' do
+    before do
+      call_count = 0
+      allow(::Resque).to receive(:enqueue_to) do
+        call_count += 1
```
cam156: @mjgiarlo I really liked my syntax much better than this here.  It was really easy (in my mind) to see what was going to happen).  This is very convoluted. That said I can live with this syntax.

mjgiarlo: I don't love this implementation, but I've tried to disentangle the code that makes a failure get raised from the code that checks to see that the job gets pushed.  The implementation before seemed to use `expect` as both a stub and a test. IOW, using `expect` asserts that the code below it (`subject.push(job)`) is expected to raise a timeout error -- why would `subject.push(job)` raise a timeout error on that line without prior `allow` stubs to create that error condition?

Not trying to split hairs here! Just trying to understand how your test was passing, and if it was doing what we thought it should do (because in the Sufia test suite we usually use `expect` for triggering tests and `allow` for setting up behaviors). Thoughts, @jcoyne ?  The original test here was:

```ruby
  expect(::Resque).to receive(:enqueue_to).and_raise(Redis::TimeoutError)
  expect(::Resque).to receive(:enqueue_to).and_return(true)
  subject.push(job)
```

jcoyne: I don't have a preference. Rock paper scissors?

PR #911, opened 2015-02-17
Updates hover states for better feedback and interaction, refs #9772

mtribone:
The hover states for the Upload, Create Collection, View Collections, View Files buttons and text at the top of the user dashboard do not have enough contrast visually. The hover from black text to blue text isn't enough of a jump for users with low vision or color blindness, Also the icons do not change state.

half a second transition between these two states:

screen shot 2015-02-17 at 3 03 45 pm
screen shot 2015-02-17 at 3 04 05 pm


jcoyne:
:+1: Looks good to me.

mjgiarlo:
:+1:

PR #912, opened 2015-02-18
add class accessors for presenter and form class for batch form and sing...

weiweishi:
...le use link view - follow up to #906


jcoyne:
:+1: Looks good to me. Thanks for contributing.

PR #921, opened 2015-02-19
Remove the duplicate showing of title and move the description up above ...

cam156:
...the table to match collection show view.

### DIFF
--- a/app/helpers/generic_file_helper.rb
+++ b/app/helpers/generic_file_helper.rb
@@ -27,6 +27,10 @@ module GenericFileHelper
     end
   end
 
+  def display_multiple value
+    Array(value).join(" | ")
+  end
+
   private
 
   def download_image_tag title = nil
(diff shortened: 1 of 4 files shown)


jcoyne:
@cam156 make sure you check that title and description didn't fall off the edit form.

jcoyne:
I think the way you want to accomplish this is by changing _show_descriptions.html.erb like this:

```erb
<% present_terms(@presenter, @presenter.terms_with_values - [:title, :description]) do |r, term| %>
```

cam156:
@jcoyne Thanks for the tip.  I was hoping it was really simple, but that was not the case.

awead:
@cam156 props for the "in progress" label. I think we should make use of that more often in our workflow!

hectorcorrea:
Looks good to me. I love the array subtraction `@presenter.terms - [:title, :description]` to get the list of terms to render.

PR #925, opened 2015-02-20
Changing to use the latest jquery hide show instead of the old style cla...

### LINKED ISSUE #916: Uploading with an error means you hang forever
When I upload multiple files
And one errors ( easiest way to get an error is a zero length file)
Then I never get a way to edit the batch metadata
And the page hangs forever

screen shot 2015-02-18 at 1 11 42 pm

Back in the day a message would show after the last upload saying one or more of your files have errored, you can edit your batch here of go to your dashboard.  To be fair I think this may have been missing before the fedora4 upgrade, but it has caused our users a lot of pain lately due to the redis timeout errors.

cam156:
...ss name.  fixes #916

### DIFF
--- a/app/views/generic_files/upload/_alerts.html.erb
+++ b/app/views/generic_files/upload/_alerts.html.erb
@@ -1,17 +1,9 @@
-<div class="alert alert-info hide" id="success">
-  You have successfully uploaded some of your files.  Either continue to upload or edit use the links below to abandon the rest of your added files.
-  <p>
-  <%= link_to 'Add Descriptions', GenericFilesController.upload_complete_path(@batch_id), class: 'btn' %>
-  </p>
+<div class="alert alert-warning" id="fail">
+  <%= t("sufia.upload.alert.fail_html", reload_href: link_to(t("sufia.upload.alert.fail_restart_href_text"), sufia.new_generic_file_path), contact_href: link_to(t("sufia.upload.alert.contact_href_text"), sufia.contact_form_index_path)) %>
 </div>
 
-<div class="alert hide" id="fail">
-  There was a problem during upload. Please click the &quot;Start Upload&quot; button or <%= link_to 'start over', sufia.new_generic_file_path %>
+<div class="alert alert-warning" id="partial_fail">
+  <%= t("sufia.upload.alert.partial_fail_html", metadata_href: link_to(t("sufia.upload.alert.partial_fail_metadata_href_text"), GenericFilesController.upload_complete_path(@batch_id), class: 'btn'), contact_href: link_to(t("sufia.upload.alert.contact_href_text"), sufia.contact_form_index_path)) %>
 </div>
 
-<div class="alert hide" id="partial_fail">
-  One or more files did not upload successfully. To continue using the files uploaded use one of the links below.<br />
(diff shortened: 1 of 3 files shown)


app/views/generic_files/upload/_alerts.html.erb
```diff
@@ -1,17 +1,10 @@
...
-  You have successfully uploaded some of your files.  Either continue to upload or edit use the links below to abandon the rest of your added files.
-  <p>
-  <%= link_to 'Add Descriptions', GenericFilesController.upload_complete_path(@batch_id), class: 'btn' %>
-  </p>
-</div>
-
-<div class="alert hide" id="fail">
-  There was a problem during upload. Please click the &quot;Start Upload&quot; button or <%= link_to 'start over', sufia.new_generic_file_path %>
+<div class="alert alert-warning" id="fail">
+  There was a problem during upload, none of your files uploaded correctly. Please <%= link_to 'start over', sufia.new_generic_file_path %>.
```
mjgiarlo: Would it be reasonable to ask that:

1) This text include a link to the /help/ URL so that folks know how to get assistance, and
2) That this be put in a locale file for overriding?

(If you think that's a separate issue, we can ticket that separately!)

app/views/generic_files/upload/_alerts.html.erb
```diff
@@ -1,17 +1,10 @@
-<div class="alert alert-info hide" id="success">
```
mjgiarlo: Why strike this whole div?

cam156: It was used before we auto forwarded, so it is never shown.  As far as I could see it was cruft and I removed it.

app/views/generic_files/upload/_alerts.html.erb
cam156: I thought about that, but I was not certain how to I18N the links.  If you have a good example I could look into it.

app/views/generic_files/upload/_alerts.html.erb
mjgiarlo: I dig cruft removal!  Thoughts, @jcoyne ?

jcoyne: Sure. I'd never seen this bit before, so take it away.

app/views/generic_files/upload/_alerts.html.erb
mjgiarlo: How about:

```yaml
en:
   sufia:
      upload:
         error_html: "None of your files were uploaded. Get <a href='%{help_url}'>, or %{restart_link}."
```

```erb
<div class="alert alert-warning" id="fail">
<%= t('sufia.upload.error_html', help_url: sufia.static_path('help'), restart_link: link_to('start over', sufia.new_generic_file_path %>
</div>
```

?

app/views/generic_files/upload/_alerts.html.erb
mjgiarlo: :ok_hand:

cam156:
@mjgiarlo It got a bit loopy, but I added I18N in.  Can you think of a better way?

mjgiarlo:
Works for me, @cam156 :+1:

PR #926, opened 2015-02-20
Keep FFmpeg instructions together

rotated8:
The (former) instructions for compiling FFmpeg for Ubuntu were below the
tag cloud. This moves them to the bottom of the FFmpeg section. The url
now points to the general compilation guide rather than the Ubuntu
specific one.


cam156:
@rotated8 :+1:  Looks good!  Just waiting on travis

Issue #914, opened 2015-02-18
There are two down arrow in search sort


cam156:
When I search sufia
And I have multiple search results
And I see a sort bar
Then there should be only one down arrow
(Just blue no black one)

![screen shot 2015-02-18 at 12 53 35 pm](https://cloud.githubusercontent.com/assets/1599081/6253164/6607bc34-b76d-11e4-8491-fa6ad7c503d3.png)

jcoyne:
Should the "10 per page" also use the blue arrow?

cam156:
@mtribone Yes or no to per page being blue too.

@jcoyne There is a larger issue that we have different sorting bars in collections dashboard, and search, but I have not yet put a ticket in for that.

mtribone:
@cam156 @jcoyne I would say no to the blue arrow on all of them. We want the default Blacklight/Bootstrap and don't need the background images anymore.

![screen shot 2015-02-18 at 1 29 03 pm](https://cloud.githubusercontent.com/assets/4163828/6253843/62a74b7c-b772-11e4-8a3a-e2be93319cd4.png)

jcoyne:
@mtribone :+1: to just the black arrows ever where.

PR #934, opened 2015-02-25
Remove arrow in relevance sorting label

awead:
Fixes issue reporting in #914 by updating the catalog_controller generator to match the current one in Blacklight.

### DIFF
--- a/spec/views/catalog/sort_and_per_page.html.erb_spec.rb
+++ b/spec/views/catalog/sort_and_per_page.html.erb_spec.rb
@@ -15,13 +15,17 @@ describe 'catalog/_sort_and_per_page.html.erb', :type => :view do
     allow(view).to receive(:blacklight_config).and_return(CatalogController.blacklight_config)
   end
 
-
-
   it 'appears on page without error' do
     render
     page = Capybara::Node::Simple.new(rendered)
     expect(page).to have_selector('span.page_entries', count: 1)
     expect(rendered).to include("<strong>4</strong> - <strong>6</strong> of <strong>20</strong>")
+  end
+  
+  it 'displays the relevance option for sorting' do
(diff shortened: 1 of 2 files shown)


mjgiarlo:
Hate to be that guy, but could you add a test that fails w/o this change and passes w/ it?  It seems we haven't had tests for this in the past.

awead:
@mjgiarlo yeah, good point. Fortunately, there was a quick view test I could piggy-back off of.

mjgiarlo:
:+1:

spec/views/catalog/sort_and_per_page.html.erb_spec.rb
```diff
@@ -15,13 +15,18 @@
...
-
-
   it 'appears on page without error' do
     render
     page = Capybara::Node::Simple.new(rendered)
     expect(page).to have_selector('span.page_entries', count: 1)
     expect(rendered).to include("<strong>4</strong> - <strong>6</strong> of <strong>20</strong>")
   end
+  
+  it 'does not use a directional arrow for relevance sorting' do
```
jcoyne: I prefer we just test for what it does do. There are lots of things it does not do that we are not testing for.

spec/views/catalog/sort_and_per_page.html.erb_spec.rb
```diff
@@ -15,13 +15,18 @@
...
+    render
+    page = Capybara::Node::Simple.new(rendered)
+    expect(rendered).to include("<li><a href=\"/catalog?sort=score+desc%2C+date_uploaded_dtsi+desc\">relevance</a></li>")
```
jcoyne: This is sufficient, we don't need the next line.

PR #938, opened 2015-02-26
Use hasSize to calculate collection size

awead:
Uses the hasSize property on content files to calculate the total size, in bytes, of the collection.

### DIFF
--- a/spec/models/collection_spec.rb
+++ b/spec/models/collection_spec.rb
@@ -19,22 +19,17 @@ describe Collection, :type => :model do
   end
 
   describe "::bytes" do
-
+    subject { @collection.bytes }
     context "with no items" do
-      specify "is zero" do
-        @collection.save
-        expect(@collection.bytes).to eq 0
-      end
+      before { @collection.save }
+      it { is_expected.to eq 0 }
     end
 
-    context "with characterized GenericFiles" do
-      let(:file) { mock_model GenericFile, file_size: ["50"] }
-      before do
-        allow(@collection).to receive(:members).and_return([file, file])
-      end
-      specify "is the sum of the files' sizes" do
-        expect(@collection.bytes).to eq 100
-      end
+    context "with two 50 byte files" do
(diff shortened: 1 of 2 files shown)


mjgiarlo:
Code looks clean and test is good.

One question: do we have a sense of how `#bytes` will work performancefise?  Is this doing Solr queries or hitting the Fedora API a bunch of times, or what?  Would it help to memoize bytes?

awead:
@mjgiarlo it should do no better or worse than the current implementation. It's going to be hitting LDP for each file, but it's cached, so that could be okay. Ideally, I think we'd have a property on the collection that gets updated with the value of `bytes` each time a file is added or deleted from the collection.

phebal:
I ran into performance problems here with a large collection taking minutes to load because of size calculations (current sufia master and before). I see some cache hits but I get no perceivable benefit from caching -- page loads in about the same time on subsequent tires. I tried the collection-size branch and can confirm that load time is still roughly the same. Maybe it's better with the latest version of ActiveFedora, I haven't tried it yet.

mjgiarlo:
@phebal That's problematic. How might you like to see this implemented? I'm wondering if we shouldn't lean on Solr more here though I say that with not much confidence that we're just kicking the can down the road a ways.

awead:
@mjgiarlo @phebal the optimization should occur when calling `object.members` since that's where the collection's items are being assembled. We can't get the kind of info out of solr at the moment, unfortunately.

phebal:
Thanks for replying @mjgiarlo and @awead . I apologize but I'm very new to hydra/sufia/fedora world and can't give an informed opinion on using Solr vs having a size property on a collection.

cam156:
@mjgiarlo @phebal I wonder if we should put a ticket into the system for upgrading performance.

phebal:
@cam156 I selfishly vote yes, but in either case I might take a look at fixing the performance issue sometime soon based on @awead suggestion.

Issue #939, opened 2015-02-26
Metadata should not bump to the next line


mtribone:
When I view the show page for a collection, the metadata for resource type is bumping to the next line. When I view the source, I find an empty anchor element with a partial href and a break element. I don't expect these to be there. See screen shot.

![screen shot 2015-02-26 at 2 29 00 pm](https://cloud.githubusercontent.com/assets/4163828/6400158/24fb31d6-bdc5-11e4-83a8-064923f49b38.png)

mtribone:
Also in rights

![screen shot 2015-02-26 at 2 50 27 pm](https://cloud.githubusercontent.com/assets/4163828/6400438/daf8b70a-bdc6-11e4-8a16-958d3b3bb5c0.png)

mtribone:
Made adjustments in the style sheet to pull dt and dd closer, but extra line needs to be resolved.

![screen shot 2015-03-03 at 9 54 52 am](https://cloud.githubusercontent.com/assets/4163828/6464920/957f148a-c18b-11e4-9f33-34ab517471ac.png)

mtribone:
It appears that empty strings are being added when the collection is created or edited. Not likely an issue with a view. See Rails output below:

![screen shot 2015-03-11 at 10 26 25 am](https://cloud.githubusercontent.com/assets/4163828/6598319/417fbb9c-c7d9-11e4-88db-e3b5b6a897f5.png)

mtribone:
So I think that I know how it is happening. When I create a new collection and add metadata by entering text and hitting the return key, another field appears underneath. If I decide to not add more metadata and leave the new field empty and hit save, I get an empty string.

mtribone:
I created a new collection, entered a name for creator and hit the return key. New field appears under the one I just filled out. Hit save and an empty string appears as the first item.

![screen shot 2015-03-12 at 10 49 29 am](https://cloud.githubusercontent.com/assets/4163828/6620150/8a52895e-c8a5-11e4-8581-fb040d60e7e1.png)

mjgiarlo:
I wonder if this is a regression related to https://github.com/projecthydra/sufia/pull/948

mtribone:
Good question @mjgiarlo . Perhaps because if I look at my original screen shots they were affecting the select list and rights dropdown, too. I don't directly see how the use of a return key on other metadata in form would add empty strings there?

mjgiarlo:
@mtribone Maybe there are two behaviors at play here and I just conflated them:

[...]

PR #940, opened 2015-02-27
Adds CSS two column for collection show to match file show

mtribone:
We go from this:

screen shot 2015-02-25 at 11 15 38 am

To this:

screen shot 2015-02-27 at 9 18 28 am

### DIFF
--- a/app/assets/stylesheets/sufia/_collections.scss
+++ b/app/assets/stylesheets/sufia/_collections.scss
@@ -7,9 +7,27 @@
   letter-spacing: .10em;
 }
 
-.metadata-collections dt {
-  float: left;
-  margin-right: 1em;
+.metadata-collections  {
+  dt {
+    color: $file-show-term-color;
+  }
+  dd {
+    padding-bottom: .35em;
+    word-wrap: break-word;
+  }
+}
+
+.metadata-collections dt::after {
+  content: ':';
+}
+
+.metadata-collections {
+  -moz-column-count: 2;
+  -webkit-column-count: 2;
+  column-count: 2;
+  -moz-column-width: 20em;
+  -webkit-column-width: 20em;
+  column-width: 20em;
 }
 
 .collection-icon, .collection-icon-search, .collection-icon-small {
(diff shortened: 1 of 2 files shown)


awead:
@mtribone my only beef with this is that the "Rights:" label is at the bottom of the first column and the content is at the top of the second. Is there a way to ensure the columns only break at labels, so the the label is always directly adjacent to the content?

jcoyne:
@awead that's the joy of definition lists.

mtribone:
@awead I had this conversation with @cam156 and perhaps saving that fix for another ticket. @jcoyne it's the CSS that breaks the dt and dd apart. The code is fluid enough to "evenly" distribute content into two columns, but breaks where it sees fit. I could add coulmn-height to see if there is an improvement, but I can't control the amount of user content. Or we'd have to break out all of the metadata individually and manually put them into Bootstrap columns. Thoughts?

awead:
@mtribone I'm good with leaving this as is, or putting in another ticket to address this later -- if that's even possible. I was thinking:
```
<div class="dont-break-here">
  <dt>
  <dd>
</div>
```
Don't know...

jcoyne:
@mtribone have you considered `break-before: avoid` on the `<dd>`

mtribone:
@awead @jcoyne :clap:  Nice. I'll give them both a shot.

jcoyne:
@awead you can't stick `div` in a `dl` http://www.w3.org/TR/html5/grouping-content.html#the-dl-element

mtribone:
I also wonder if it is because there is an empty anchor element that is the first line. https://github.com/projecthydra/sufia/issues/939

awead:
@jcoyne then a class on `dt` that disallows breaking might work.

jcoyne:
@cam156 why is this `In Progress`? Is there any reason we can't merge it now?

mtribone:
We can push the button and then work on 939 to help alleviate the issue. It doesn't happen on the file show page and I suspect it is because of the empty anchor.

mjgiarlo:
@jcoyne waffle.io did that.

jcoyne:
@mjgiarlo So what's the implication of that? Does that mean "do not merge because I'm working on it"?

mjgiarlo:
I haven't touched waffle.io yet, so I'd defer to Carolyn.  Hey, @cam156?

cam156:
@jcoyne @mjgiarlo No the implication is that someone assigned them self.  I'll see what I can do to change waffle so it does not add labels...

mjgiarlo:
:ok: :+1:

Issue #942, opened 2015-02-27
Auto linking urls in the description field


awead:
As a user
When I add "http://my.link.com/" to the description field
Then I should see a link to my.link.com

awead:
heh. Github does it...

jcoyne:
Also, when I link to `javascript:alert('foo')` It should not be a link.

mjgiarlo:
#943 was already merged, it looks like.  Is this a concern we should test, @jcoyne?

jcoyne:
Maybe waiting for other institutions to review PRs is a good idea.

jcoyne:
Anyways, it uses auto_link, so it' probably okay.

jcoyne:
Should this be closed?

awead:
@jcoyne This specific issue is fixed, but if there are concerns with the javascript, we should create a new issue.

PR #943, opened 2015-02-27
Add auto-link to description, in #942

awead:
Addresses issue with links appearing in the description field. Note that the link is absent in the twitter meta tag.


mjgiarlo:
:+1: 

Not at all concerned about Twitter meta tag.

PR #944, opened 2015-03-02
Updating to use the new blacklight partial instead of defining in-line

cam156:
Using the new blacklight partial (https://github.com/projectblacklight/blacklight/blob/master/app/views/catalog/_search_results.html.erb) instead of overriding the entire contents to change the zero response.


mjgiarlo:
:+1:

PR #947, opened 2015-03-03
Replaces icon font with default SVG graphic

mtribone:
Current icon is the same as the icon for gallery view and could be confusing. Moves default graphic out of ScholarSphere and into Sufia.


mjgiarlo:
:+1: Makin' it :bangbang:

PR #948, opened 2015-03-03
Remove blank strings from files and collections

### LINKED ISSUE #941: GenericFiles and Collections are saved with blank strings for empty form fields
When updating or creating GenericFiles or Collections, attributes that stem from blank fields in forms are saved as either `""` or `[""]` depending on the type. I'm not sure if this change is intentional but Sufia 4.x preserved them as either `nil` or `[]`. One side effect of this is that it breaks the `Sufia::CollectionPresenter#terms_with_values` for multi-value attributes because `[""].present?` evaluates to `true`.

phebal:
Currently blank strings are saved in place of empty arrays for null values as described in  #941 

Problems with `Collection` were only fixed in Sufia and not hydra-collections, because Sufia defines its own `collection_params` method and because `model_attributes` is available to remove the blank strings.

I'm not sure that the fix for `GenericFile` blanks removal is correct. I don't fully understand what the implications of removing a part of the `xml_template` from `FitsDatastream` are, but many terms were already missing from it so I just removed the rest of them. I couldn't pass few of the tests before and after the change because of incompatibilities with ImageMagic and ffmpeg but the numbers remained constant.

Closes #941

### DIFF
--- a/spec/models/generic_file_spec.rb
+++ b/spec/models/generic_file_spec.rb
@@ -8,6 +8,14 @@ describe GenericFile, :type => :model do
     @file.apply_depositor_metadata(user.user_key)
   end
 
+  context "when it is initialized" do
+    it "has empty arrays for all the properties" do
+      subject.attributes.each do |k,v|
+        expect(Array(v)).to eq([])
+      end
+    end
+  end
+
   describe "created for someone (proxy)" do
     before do
       @transfer_to = FactoryGirl.find_or_create(:jill)
--- a/sufia-models/app/models/datastreams/fits_datastream.rb
+++ b/sufia-models/app/models/datastreams/fits_datastream.rb
@@ -144,42 +144,7 @@ class FitsDatastream < ActiveFedora::OmDatastream
     http://hul.harvard.edu/ois/xml/xsd/fits/fits_output.xsd",
                version: "0.6.0",
                timestamp: "1/25/12 11:04 AM") {
-        xml.identification {
-          xml.identity(format: '', mimetype: '',
-                       toolname: 'FITS', toolversion: '') {
-            xml.tool(toolname: '', toolversion: '')
-            xml.version(toolname: '', toolversion: '')
-            xml.externalIdentifier(toolname: '', toolversion: '')
-          }
-        }
-        xml.fileinfo {
-          xml.size(toolname: '', toolversion: '')
-          xml.creatingApplicatioName(toolname: '', toolversion: '',
-                                     status: '')
-          xml.lastmodified(toolname: '', toolversion: '', status: '')
-          xml.filepath(toolname: '', toolversion: '', status: '')
-          xml.filename(toolname: '', toolversion: '', status: '')
-          xml.md5checksum(toolname: '', toolversion: '', status: '')
-          xml.fslastmodified(toolname: '', toolversion: '', status: '')
-        }
-        xml.filestatus {
-          xml.tag! "well-formed", toolname: '', toolversion: '', status: ''
-          xml.valid(toolname: '', toolversion: '', status: '')
-        }
-        xml.metadata {
-          xml.document {
-            xml.title(toolname: '', toolversion: '', status: '')
-            xml.author(toolname: '', toolversion: '', status: '')
-            xml.pageCount(toolname: '', toolversion: '')
-            xml.isTagged(toolname: '', toolversion: '')
-            xml.hasOutline(toolname: '', toolversion: '')
-            xml.hasAnnotations(toolname: '', toolversion: '')
-            xml.isRightsManaged(toolname: '', toolversion: '',
-                                status: '')
-            xml.isProtected(toolname: '', toolversion: '')
-            xml.hasForms(toolname: '', toolversion: '', status: '')
-          }
-        }
+        xml.identification { xml.identity(toolname: 'FITS') }
       }
     end
     builder.doc
--- a/app/controllers/concerns/sufia/collections_controller_behavior.rb
+++ b/app/controllers/concerns/sufia/collections_controller_behavior.rb
@@ -44,10 +44,12 @@ module Sufia
     end
 
     def collection_params
-      params.require(:collection).permit(:title, :description, :members, part_of: [],
-        contributor: [], creator: [], publisher: [], date_created: [], subject: [],
-        language: [], rights: [], resource_type: [], identifier: [], based_near: [],
-        tag: [], related_url: [])
+      form_class.model_attributes(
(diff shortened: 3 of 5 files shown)


spec/models/generic_file_spec.rb
```diff
@@ -8,6 +8,16 @@
     @file.apply_depositor_metadata(user.user_key)
   end
 
+  it 'is initialized with blanks' do
+    subject.attributes.each do |k,v|
+      if v.is_a?(Array)
```
mjgiarlo: How about replacing lines 13-17 with:

```ruby
expect(Array(v)).to eq([])
```

:question:

phebal: It seems much cleaner, will fix. Thanks for the suggestion.

phebal: @mjgiarlo that brings a question on form, should I rebase or make a new commit and then expect someone else to rebase before merging? I don't see that addressed in the guidelines for contributing. Thanks.

mjgiarlo:
Hi, @phebal,

Thanks for the high-quality pull request with a detailed description. (This helps. A lot!)

In order for us to merge this code, we'll need to have signed contributor license agreements on file for you as an individual and your employer. You can read more about our CLA processes here:
https://wiki.duraspace.org/display/hydra/Hydra+Project+Intellectual+Property+Licensing+and+Ownership
The process itself is fairly simple -- you just need to print out the Hydra Project individual Contributor License Agreement (iCLA) and sign it and then send it to legal@projecthydra.org. Same for the corporate Contributor License Agreement (cCLA) but that document must be signed with someone who has signatory authority for your institution. (This can be more time-consuming, but folks at the Northwestern University Library have already done so, so there is precedent for this on campus.)

In the meantime, I am looking into whether Northwestern's current cCLA covers you or not. I will respond when I have a clear answer to that.

Let me know if y'all have any questions about this process, and I'd be happy to help.

spec/models/generic_file_spec.rb
mjgiarlo: I think you should:

1. Fix this in a new commit
2. Squash all your commits in this branch
2a. (OPTIONAL) if your branch has diverged substantially from master, rebase on the latest in master. I suspect you don't need to do this now, but re: my comment about CLAs, you may need to do that eventually depending on how quickly you can get CLAs signed.
3. Force-push your branch, and this PR will be automatically updated.

How does that sound?

phebal:
@mjgiarlo thanks for the explanation about handling the legal stuff, and for checking on Northwestern's cCLA. I'll send in the iCLA while I'm waiting on the results of your inquiry.

spec/models/generic_file_spec.rb
phebal: Sounds good, I fixed per https://github.com/projecthydra/sufia/pull/948#discussion_r25710327 and squashed. Thanks.

mjgiarlo:
Got it.  The Steering Group will process your iCLA in the meantime.  It may also be prudent to inquire with your supervisor about how you would deal with the cCLA should that be needed.  I'll drop back in when I get more information about the coverage of NU's existing cCLA. Thank you for your patience.

mjgiarlo:
And thanks for the re-push.  Code looks :+1: to me!

Any comments, @jcoyne @cam156 @awead others?

spec/models/generic_file_spec.rb
```diff
@@ -8,6 +8,12 @@
     @file.apply_depositor_metadata(user.user_key)
   end
 
+  it 'is initialized with blanks' do
```
jcoyne: I'm not sure why this test is necessary. Seems like this is taken care of by ActiveTriples.

sufia-models/app/models/datastreams/fits_datastream.rb
```diff
@@ -144,42 +144,7 @@ def self.xml_template
     http://hul.harvard.edu/ois/xml/xsd/fits/fits_output.xsd",
                version: "0.6.0",
                timestamp: "1/25/12 11:04 AM") {
-        xml.identification {
```
jcoyne: How is this change relevant to the goal of the PR? This seems like a backwards incompatible change and we should deprecate all these methods.

phebal: The intended purpose of this change is to get rid of the blank strings (`[""]` and `""`) from `GenericFiles` on initialization that are put into some attributes because of the content of this template. You can see what I mean by running `::GnericFile.new` (in current `master`) and looking at the content of the characterization related attributes. As I've indicated above, I'm not sure about the correctness of my solution or its backward compatibility. I'm not even sure what you mean by `we should deprecate all these methods`? I'm new to the project so you'll have to bear with me.

spec/models/generic_file_spec.rb
phebal: It's to illustrate the changes made to the `FitsDatastream`, this test doesn't pass against the current master but passes against this branch.

sufia-models/app/models/datastreams/fits_datastream.rb
jcoyne: This is not a template. It is a directive for parsing xml files.  If the attributes are returning empty strings, it's because the XML tags are empty.  People may be depending on these directive to say, get the md5checksum, now that you've remove them, they'll get a NoMethodError. So if you want to remove them, you should make a replacement method that does exactly what these do and log an error about the method being deprecated and give an alternative method for getting at each thing.

jcoyne: Whoops! I misread the commit. You are right. This is the template.  Yeah, I don't see any reason for us to generate a fully formed template. :+1:

spec/models/generic_file_spec.rb
jcoyne: Can we rephrase to:
```ruby
context "when it is initialized" do
  it "has empty arrays for all the properties" do
    subject.attributes.each do |k,v|
      expect(Array(v)).to be_empty
    end
  end
end
```

jcoyne:
yeah, this makes sense now. Sorry I read it all wrong the first pass through.

spec/models/generic_file_spec.rb
phebal: This seems better @jcoyne will fix. Thanks.

cam156:
@mjgiarlo Are we waiting in CLA's?  Looks good to me.  I think this was required when we were using OM, but we are no longer dependent on OM.

mjgiarlo:
Hi, yes, we are waiting on a clarification re: Northwestern's cCLA.  Moving that along as quickly as we can.

mjgiarlo:
All, @phebal's CLAs are now in order.

Issue #915, opened 2015-02-18
Empty Batches


cam156:
When a user goes to the upload page a batch gets create
When the user refreshes the page a new batch gets created
When the user refreshes the page 100 times 100 batches get created
Then the system should not have 100 empty batches forever

This could be tackled in two ways: 1) periodically remove batches 2) change the process too not create the empty batches

We will eventually have empty batches in the system is a user uploads files and then deletes them, but that is a different scenario (and far less likely to occur regularly).

hectorcorrea:
PR #961 delays the creation of batch. With this PR the batch will be created only if the user actually clicks on the "upload files" button rather than by just visiting the upload page. This will reduce the frequency of users leaving empty batches.

PR #949, opened 2015-03-05
Rake task for deleting empty batches

awead:
Addresses #915. While it's unlikely that users will find batches without any associated files, it can happen and a simple, default task for locating and deleting them is provided.

### DIFF
--- a/sufia-models/lib/tasks/batch_cleanup.rake
+++ b/sufia-models/lib/tasks/batch_cleanup.rake
@@ -1,0 +1,19 @@
+namespace :sufia do
+
+  desc "Reports on and optionally removes empty batches that contain no associated files"
+  task :empty_batches, [:remove] => :environment do |t, args|
+    option = args.to_hash.fetch(:remove, "keep")
+    Batch.all.each do |batch|
+      if batch.generic_files.empty?
+        print "#{batch.id} contains no files - "
+        if option == "remove"
+          batch.destroy
+          puts "deleted"
+        else
+          puts "to delete, rerun with the remove option: rake sufia:empty_batches[remove]"
+        end
+      end
+    end
+  end
+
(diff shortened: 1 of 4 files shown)


sufia-models/lib/tasks/batch_cleanup.rake
```diff
@@ -0,0 +1,19 @@
+namespace :sufia do
+
+  desc "Reports on and optionally removes empty batches that contain no associated files"
+  task :empty_batches, [:remove] => :environment do |t, args|
+    option = args.to_hash.fetch(:remove, "keep")
+    Batch.all.each do |batch|
+      if batch.generic_files.empty?
+        print "#{batch.id} contains no files - "
```
mjgiarlo: `puts` ?

mjgiarlo:
Travis is :unamused: and could you add spec coverage?

sufia-models/lib/tasks/batch_cleanup.rake
awead: `print` omits the line ending and keeps it all on the same line.

awead:
How would you test the rake task?

sufia-models/lib/tasks/batch_cleanup.rake
mjgiarlo: Ah!  Didn't realize that was intentional.  OK then!

mjgiarlo:
We have an example rake task spec in ScholarSphere you can look at: https://github.com/psu-stewardship/scholarsphere/blob/develop/spec/rake/sitemap_spec.rb

awead:
@mjgiarlo yo :dog2: I head like you spec tests.

mjgiarlo:
:+1:

PR #952, opened 2015-03-06
Blacklight's generator doesn't use a hashrocket

awead:
This was breaking `rake engine_cart:generate` because Blacklight's original root path wasn't getting removed. The tests would then fail because there were two root paths.

### DIFF
--- a/lib/generators/sufia/install_generator.rb
+++ b/lib/generators/sufia/install_generator.rb
@@ -77,7 +77,7 @@ module Sufia
 
     # The engine routes have to come after the devise routes so that /users/sign_in will work
     def inject_routes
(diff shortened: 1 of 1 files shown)


jcoyne:
How about:
```ruby 
gsub_file 'config/routes.rb',  /root (:to =>|to:) "catalog#index"/, ''
```

awead:
`gsub_file 'config/routes.rb', /root .+ "catalog#index"$/, ''`

jcoyne:
@awead are you planning on updating this to use a regex?

awead:
This should be superseded by the fix for #953

jcoyne:
@awead I don't see any fix for #953

PR #954, opened 2015-03-09
Fix generator for blacklight 5.10

jcoyne:


mjgiarlo:
Travis is :unamused:

jcoyne:
@mjgiarlo yeah, I know. I believe it's due to blacklight_advanced_search doing some nasty monkey patching.

awead:
This uses blacklight_advanced_search from master to avoid lots of deprecation warnings. if we want, we can wait and merge this once new versions of that and hydra-head have been released,

jcoyne:
@awead thanks for tackling this, but I'd say wait until versions of hydra-head and blacklight_advanced_search are out.

awead:
@jcoyne agreed. Do you want to cut another release of hydra-head? I can ask the Blacklighters about the advanced search gem.

spec/views/batch/edit.html.erb_spec.rb
```diff
@@ -1,8 +1,12 @@
 require 'spec_helper'
 
 describe 'batch/edit.html.erb' do
-  let(:batch) { stub_model(Batch, id: '123') }
-  let(:generic_file) { stub_model(GenericFile, id: nil, depositor: 'bob', rights: ['']) }
+  let(:batch) { Batch.create }
+  let(:generic_file) do
+    GenericFile.new(title: ['some title']).tap do |f|
```
mjgiarlo: Are you intentionally not saving this file?

jcoyne: the tap is unnecessary here. AF::Base.new takes a block in AF 9

awead: @mjgiarlo it doesn't need to be save. We just need to have an actual batch in hand for the test.

mjgiarlo: :ok:

PR #957, opened 2015-03-10
Making the stats import a bit more resilient.

hectorcorrea:
There are two main changes in this PR. 

One is the addition of a new parameter to the `UserStatImporter` to allow for a delay between the calls to Google Analytics. We use this in our overwrite in ScholarSphere and it makes sense to make it available in Sufia.

The second change is to allow the `UserStatImporter` to process first the users with the oldest stats.

### DIFF
--- a/spec/lib/sufia/user_stat_importer_spec.rb
+++ b/spec/lib/sufia/user_stat_importer_spec.rb
@@ -10,6 +10,7 @@ describe Sufia::UserStatImporter do
 
   let(:bilbo) { FactoryGirl.create(:user, email: 'bilbo@example.com') }
   let(:frodo) { FactoryGirl.create(:user, email: 'frodo@example.com') }
+  let!(:gollum) { FactoryGirl.create(:user, email: 'gollum@example.com') }
 
   let!(:bilbo_file_1) do
     GenericFile.new(id: 'bilbo1').tap do |f|
@@ -143,7 +144,7 @@ describe Sufia::UserStatImporter do
     end
 
     it "doesn't duplicate entries for existing dates" do
-      expect(User.count).to eq 2
+      expect(User.count).to eq 3
       expect(UserStat.count).to eq 3
 
       Sufia::UserStatImporter.new.import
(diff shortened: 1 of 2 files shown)


spec/lib/sufia/user_stat_importer_spec.rb
```diff
@@ -10,6 +10,7 @@
 
   let(:bilbo) { FactoryGirl.create(:user, email: 'bilbo@example.com') }
   let(:frodo) { FactoryGirl.create(:user, email: 'frodo@example.com') }
+  let!(:gollum) { FactoryGirl.create(:user, email: 'gollum@example.com') }
```
mjgiarlo: Just curious why you used `let!` here and `let` for the other hobbitses.

spec/lib/sufia/user_stat_importer_spec.rb
```diff
@@ -164,6 +165,13 @@
 
       expect(frodos_stats[0].file_downloads).to eq(frodo_file_1_download_stats[0].totalEvents.to_i)
     end
+
+    it "processes the oldest records first" do
+      # Since Gollum has no stats it will be the first one processed. 
+      # Followed by Frodo and Bilbo.
+      sorted_ids = Sufia::UserStatImporter.new.users_sorted.map {|u| u.id}
```
mjgiarlo: Minor points:
1. I'd expect the method to be named `sorted_users` rather than `users_sorted`.
1. Add spacing inside the braces: `.map { |u| u.id }`

spec/lib/sufia/user_stat_importer_spec.rb
hectorcorrea: Since gollum is only used in the expect condition (gollum.id) I had to use `let!` to force RSpec to execute the statement. Otherwise the record wouldn't be created. I spent like 20 minutes scratching my head on why the record was there.

sufia-models/lib/sufia/models/stats/user_stat_importer.rb
```diff
@@ -25,9 +30,24 @@ def import
...
+      ::User.find_each do |user|
+        users.push(UserRecord.new(user.id, user.user_key, date_since_last_cache(user)))
+      end
+      users.sort! {|a, b| a.last_stats_update <=> b.last_stats_update}
+    end
 
 private
 
+    def delay 
+      return if @delay_secs.nil?
```
mjgiarlo: You can also coerce `@delay_secs` to an integer on line 9, and strike this line, since if `options[:delay_secs]` is nil, that will ultimately tell Ruby to sleep for 0 seconds, which has the effect of not sleeping.

spec/lib/sufia/user_stat_importer_spec.rb
mjgiarlo: Ha!  Well done, sir.

spec/lib/sufia/user_stat_importer_spec.rb
hectorcorrea: Done.

sufia-models/lib/sufia/models/stats/user_stat_importer.rb
hectorcorrea: Done. I am forcing to float, though, because I want fractions of a second to be allowed.

mjgiarlo:
:+1:

PR #961, opened 2015-03-13
Delay creation of batch until the user actually submits files. 

hectorcorrea:
This will reduce the number of orphan batches created since we won't create a new batch when the user issues an HTTP GET on the page, rather we will wait until the user issues an HTTP POST to the page which is more inline with the way HTTP calls are expected to behave.

### DIFF
--- a/app/controllers/concerns/sufia/files_controller_behavior.rb
+++ b/app/controllers/concerns/sufia/files_controller_behavior.rb
@@ -48,7 +48,7 @@ module Sufia
 
     # routed to /files/new
     def new
-      @batch_id = Batch.create.id
+      @batch_id = Sufia::IdService.mint
     end
 
     # routed to /files/:id/edit
@@ -211,6 +211,9 @@ module Sufia
     end
 
     def process_file(file)
+
+      Batch.find_or_create(params[:batch_id])
+
       update_metadata_from_upload_screen
       actor.create_metadata(params[:batch_id])
       if actor.create_content(file, file.original_filename, file_path, file.content_type)
--- a/sufia-models/app/models/batch.rb
+++ b/sufia-models/app/models/batch.rb
@@ -10,13 +10,27 @@ class Batch < ActiveFedora::Base
   property :status, predicate: ::RDF::DC.type
 
   def self.find_or_create(id)
-    # FIXME potential race condition in this method. Consider that `find' may raise
-    # ObjectNotFound in multiple processes. However, Fedora should raise an error
-    # if we try to create two objects with the same id.
     begin
       Batch.find(id)
     rescue ActiveFedora::ObjectNotFoundError
-      Batch.create(id: id)
+      safe_create(id)
     end
   end
+
+  private
+
+    # This method handles most race conditions gracefully. 
+    # If a batch with the same ID is created by another thread
+    # we fetch the batch that was created (rather than throwing
+    # an error) and continute.
+    def self.safe_create(id)
+      begin      
+        Batch.create(id: id)
+      rescue ActiveFedora::IllegalOperation => ex
+        # This is the exception thrown by LDP when we attempt to 
+        # create a duplicate object. If we can find the object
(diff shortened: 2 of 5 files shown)


awead:
Seems pretty clever to me!

app/controllers/concerns/sufia/files_controller_behavior.rb
```diff
@@ -48,7 +48,7 @@ module FilesControllerBehavior
 
     # routed to /files/new
     def new
-      @batch_id = Batch.create.id
+      @batch_id = Sufia::IdService.mint
```
mjgiarlo: Hmmm, shouldn't `Sufia::IdService.mint` be already getting called behind the scenes?

mjgiarlo:
:+1: 

Travis is sad.

jcoyne:
I'm a tiny bit concerned about potential race conditions here. When you press "start upload" on the form, a new process starts for each file you upload all trying to call "process_file" simultaneously.  I would think that some might raise an exception if they couldn't find the batch, but also couldn't create it (because one was just created).

hectorcorrea:
@jcoyne I've updated Batch.find_or_create to be a bit more resilient to race conditions. This plus the fact that browsers limit the number of requests to the same host should make the race conditions very unlikely to happen.

sufia-models/app/models/batch.rb
```diff
@@ -10,13 +10,29 @@ class Batch < ActiveFedora::Base
...
   end
+
+  private
+
+    # This method handles most race conditions gracefully. 
+    # If a batch with the same ID is created by another thread
+    # we fetch the batch that was created (rather than throwing
+    # an error) and continute.
+    def self.safe_create(id)
+      begin
```
mjgiarlo: You could also use method-level `rescue` here and eliminate a couple LOC.

sufia-models/app/models/batch.rb
```diff
@@ -10,13 +10,29 @@ class Batch < ActiveFedora::Base
...
+    # we fetch the batch that was created (rather than throwing
+    # an error) and continute.
+    def self.safe_create(id)
+      begin      
+        Batch.create(id: id)
+      rescue ActiveFedora::IllegalOperation => ex
+        if ex.message.starts_with?("Attempting to recreate existing ldp_source")
+          # This is OK. The batch was created in between the checks. 
+          Batch.find(id)
+        else
```
mjgiarlo: Do you want this else only to occur when `ActiveFedora::IllegalOperation` is raised or when any other exception is raised?

hectorcorrea: @mjgiarlo Exceptions that are not `ActiveFedora::IllegalOperation` won't make it to the rescue block. The `else` in this case is for when I get an `ActiveFedora::IllegalOperation` that is not the specific one that I am after. Eventually, I would like to update `ActiveFedora` to raise a more specific exception (e.g. DuplicateRecordFound) and update this code.

mjgiarlo: :OK:, just checking!

sufia-models/app/models/batch.rb
```diff
@@ -10,13 +10,29 @@ class Batch < ActiveFedora::Base
...
+
+    # This method handles most race conditions gracefully. 
+    # If a batch with the same ID is created by another thread
+    # we fetch the batch that was created (rather than throwing
+    # an error) and continute.
+    def self.safe_create(id)
+      begin      
+        Batch.create(id: id)
+      rescue ActiveFedora::IllegalOperation => ex
+        if ex.message.starts_with?("Attempting to recreate existing ldp_source")
```
jcoyne: This message is likely to change, which would break this code. I would suggest skipping this step.

hectorcorrea: @jcoyne Good point. Since we are doing a `find` inside the rescue it seems pretty safe to remove the check (and code will be less brittle as you indicated)

PR #964, opened 2015-03-16
Hide Buttons When No Permission

terrellt:
This hides the create file/collection buttons when there's no permission to access them. It also lets you configure whether or not to show the "Share Your Work" button.

### DIFF
--- a/spec/models/ability_spec.rb
+++ b/spec/models/ability_spec.rb
@@ -10,7 +10,7 @@ describe Sufia::Ability, :type => :model do
     it { is_expected.not_to be_able_to(:create, ContentBlock) }
     it { is_expected.not_to be_able_to(:update, ContentBlock) }
     it { is_expected.to     be_able_to(:read,  ContentBlock) }
-    it { is_expected.to     be_able_to(:view_upload, GenericFile) }
+    it { is_expected.to     be_able_to(:view_share_work, GenericFile) }
   end
 
   describe "a registered user" do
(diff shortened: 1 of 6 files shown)


spec/models/ability_spec.rb
```diff
@@ -10,6 +10,7 @@
     it { is_expected.not_to be_able_to(:create, ContentBlock) }
     it { is_expected.not_to be_able_to(:update, ContentBlock) }
     it { is_expected.to     be_able_to(:read,  ContentBlock) }
+    it { is_expected.to     be_able_to(:view_upload, GenericFile) }
```
cam156: @terrellt  Why view_upload here?  You are removing the upload button without the create ability and upload ~= create.
Should this be only view?

terrellt: > Why view_upload here?

This is actually a permission to hide the "Share Your Work" button on the home page. Maybe a different name? :view_share_work ?

spec/views/homepage/_home_header.html.erb_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+  let(:ability) { instance_double("Ability") }
+  describe "share your work button" do
+    before do
+      allow(controller).to receive(:current_ability).and_return(ability)
+      allow(ability).to receive(:can?).with(:view_share_work, GenericFile).and_return(can_view_share_work)
+      stub_template "homepage/_marketing" => "marketing"
+      render
+    end
+    context "when the user can view" do
+      let(:can_view_share_work) { true }
```
mjgiarlo: What magic is this?  Is this a thing?  If so, cool!

mjgiarlo:
Thanks for your contribution to Sufia, @terrellt!

PR #978, opened 2015-03-18
Replaces redundant title attribute with more beneficial text for all ...

mtribone:
users. Instead of just repeating the text that is the main link text in the title attribute, replace it with advisory information that informs users that they can change the permissions or visibility of the resource by selecting the link.


jcoyne:
Can you put this text in the locale file?

mtribone:
Ah! Can do. Forgot. Thanks @jcoyne

mtribone:
Okay added the text to the locale file and updated additional files. Waiting for travis.

mtribone:
Found a couple more instances and fixed a spec test.

PR #980, opened 2015-03-20
Prevent from picking up Blacklight 5.10 

hectorcorrea:
Blacklight 5.10 does not work out of the box with this version of Sufia.


awead:
@hectorcorrea you should also pin blacklight_advanced_search to < 5.1.4. Are there any others? hydra-X ,etc?

jeremyf:
Ping @hectorcorrea

hectorcorrea:
@jeremyf pong. I pinned the version of blacklight_advanced_search in my local branch and I am still getting some errors but I haven't have time to work them out. This PR as-is makes a few things better but it is incomplete. Let me see if I can get back to this today or tomorrow and push a better version.

Issue #981, opened 2015-03-21
Long file names from Box


awead:
When integration with Box, there will be very long urls as the file name instead of the label which is the actual name of the file.

PR #983, opened 2015-03-24
Use the label instead of the file_name, #981

awead:
This fixes issue #981. We were seeing long urls when ingesting files via Box. The label property contained the correct file name, where as title was getting set to the `file_name` variable which was in fact the url to reach the file on Box, long, randomly generated one. This corrects the issue by setting title to the same as label.


cam156:
:+1:   Looks good to me.  @jcoyne?

PR #989, opened 2015-03-25
Upgrade to Blacklight 5.12.1

jcoyne:


Gemfile
```diff
@@ -4,7 +4,7 @@ source 'https://rubygems.org'
 gemspec
 
 # Required for doing pagination inside an engine. See https://github.com/amatsuda/kaminari/pull/322
-gem 'kaminari', github: 'harai/kaminari', branch: 'route_prefix_prototype'
+gem 'kaminari', github: 'jcoyne/kaminari', branch: 'sufia'
```
awead: How about a new gem called coyminari... which is exactly like kaminari, but better because it has this PR merged. :smiling_imp:

PR #1017, opened 2015-03-27
Set the RDF type of GenericFile to pcdm:Object

jcoyne:


awead:
@jcoyne I'm assuming this is to be for a 6.1 release?

cam156:
@awead I think jcoyne was think for 6.0.  That way when we move to PCDM the object is already typed correctly.

jcoyne:
@awead yeah, we should release this ASAP 6.0.1

mjgiarlo:
:+1: to a 6.0.1 release.

PR #1020, opened 2015-03-30
Noid service extracted out into its own gem

mjgiarlo:


jcoyne:
Great!

mjgiarlo:
You may also want to review the code upon which this depends, which is brand new in projecthydra-labs, since we will depend upon it:

https://github.com/projecthydra-labs/active_fedora-noid

PR #1021, opened 2015-03-30
Download full-text jars via HTTPS

mjgiarlo:
Or else the task winds up downloading a bunch of 161-byte HTML files with 302 FOUND in them.


mjgiarlo:
Did this work before for you, @awead?  The change from wget to Net::HTTP did not work for me, and I had a couple people on IRC verify it was also broken for them.  (And I fear Travis wouldn't have picked up these errors since the one test that checks this is set not to run on Travis.)

PR #1022, opened 2015-03-30
Add Arkivo API integration

mjgiarlo:
Integration with Zotero-managed publications is possible using [Arkivo](https://github.com/inukshuk/arkivo). Arkivo is a Node-based Zotero subscription service that monitors Zotero for changes and will feed those changes to your Sufia-based app. [Read more about this work.](https://www.zotero.org/blog/feeds-and-institutional-repositories-coming-to-zotero/)

To enable Zotero integration, first [register an OAuth client with Zotero](https://www.zotero.org/oauth/apps), then [install and start Arkivo-Sufia](https://github.com/inukshuk/arkivo-sufia) and then generate the Arkivo API in your Sufia-based application:

```
rails g sufia:models:arkivo_api
```

The generator does the following:

* Enables the API in the Sufia initializer
* Adds a database migration
* Creates a routing constraint that allows you to control what clients can access the API
* Copies a config file that allows you to specify the host and port Arkivo is running on
* Copies a config file for your Zotero OAuth client credentials

Update your database schema with `rake db:migrate`.

Add unique Arkivo tokens for each of your existing user accounts with `rake sufia:user:tokens`. (New users will have tokens created as part of the account creation process.)

Edit the routing constraint in `config/initializers/arkivo_constraint.rb` so that your Sufia-based app will allow connections from Arkivo. **Make sure this is restrictive as you are allowing access to an API that allows creates, updates and deletes.**

Tweak `config/arkivo.yml` to point at the host and port your instance of Arkivo is running on.

Tweak `config/zotero.yml` to hold your Zotero OAuth client key and secret. Alternatively, if you'd rather not paste these into a file, you may use the environment variables `ZOTERO_CLIENT_KEY` and `ZOTERO_CLIENT_SECRET`.

Restart your app and it should now be able to pull in Zotero-managed publications on behalf of your users. Each user will need to link their Sufia app account with their Zotero accounts, which can be done in the "Edit Profile" page. After the accounts are linked, Arkivo will create a subscription to that user's Zotero-hosted "My Publications" collection. When users add items to their "My Publications" collection via the Zotero client, they will automatically be pushed into the Sufia-based repository application. Updates to these items will trigger updates to item metadata in your app, and deletes will delete the files from your app.


app/controllers/api/items_controller.rb
```diff
@@ -0,0 +1,74 @@
...
+        return render plain: "id '#{params[:id]}' not found", status: :not_found
+      end
+
+      def actor
+        Sufia::Arkivo::Actor.new(user, item)
+      end
+
+      def token
+        (request.get? || request.delete?) ? params[:token] : item['token']
+      end
```
cam156: @mjgiarlo Not sure I follow why you would not also get the param[:token] with a put.  Does a token not get sent for a post and put?

mjgiarlo: On GET and DELETE, the token is passed as a parameter in the request (because GETs don't have a body, and Arkivo doesn't send DELETEs with a body). With PUT and POST, the token is parsed out of JSON that is in the body of the request.

cam156: Ahh! that explains where item is coming from

app/helpers/sufia/sufia_helper_behavior.rb
```diff
@@ -4,6 +4,14 @@ def orcid_label(style_class='')
       "#{image_tag 'orcid.png', { alt: t('sufia.user_profile.orcid.alt'), class: style_class }} #{t('sufia.user_profile.orcid.label')}".html_safe
     end
 
+    def zotero_label(style_class='')
+      "#{image_tag 'zotero.png', { alt: t('sufia.user_profile.zotero.alt'), class: style_class }} #{t('sufia.user_profile.zotero.label')}".html_safe
+    end
+
+    def zotero_profile_url(user_id)
```
cam156: Is this the zotero id, or Sufia user key?  It might be good to be more explicit

lib/sufia/arkivo.rb
```diff
@@ -0,0 +1,11 @@
+require 'sufia/arkivo/config'
```
cam156: @mjgiarlo Was there a decision to put this into lib instead of app?  It seems to be moving against the current trend

lib/sufia/arkivo/actor.rb
```diff
@@ -0,0 +1,96 @@
...
+      end
+
+      def create_file_from_item
+        batch = Batch.create
+        file_actor = Sufia::GenericFile::Actor.new(::GenericFile.new, user)
+        file_actor.create_metadata(batch.id)
+        store_checksum(file_actor.generic_file)
+        file_actor.create_content(file, item['file']['filename'], file_path, item['file']['contentType'])
+        ::Resque.inline = true
+        Sufia.queue.push(BatchUpdateJob.new(user.user_key, batch.id, item['metadata']['title'], attributes, default_visibility))
```
cam156: @mjgiarlo does this already run in the background, or in user context?  If it is already in the background you might just want to run the batch job.

lib/sufia/arkivo/actor.rb
```diff
@@ -0,0 +1,96 @@
...
+        def file_path
+          'content'
+        end
+
+        def file
+          extract_file_from_item
+        end
+
+        def extract_file_from_item
+          encoded = item['file']['base64']
```
cam156: @mjgiarlo  Do we really need a TempFile here. could we just use http://ruby-doc.org/stdlib-1.9.3/libdoc/stringio/rdoc/StringIO.html

lib/sufia/arkivo/actor.rb
cam156: @mjgiarlo NM I see the inline true.  Explicitly just calling run may be less "magic"

app/helpers/sufia/sufia_helper_behavior.rb
mjgiarlo: Zotero ID.

lib/sufia/arkivo.rb
mjgiarlo: Yes, I did decide to throw it in `lib/` intentionally, but I'm happy to revisit that decision.  In short, the reason was that I thought it might be easier to extract if I had the Zotero stuff all sitting in the same spot in `lib/` without much knowledge of the app within which it is contained. I made that decision early on, but then as I developed, I very well could have lost sight of this as I sprinted towards the finish line.

lib/sufia/arkivo/actor.rb
mjgiarlo: Ah, do we have an example of that syntax somewhere?  I briefly looked for an example of it, couldn't find it, and used inline and moved on.  Happy to rework this for clarity.

(The reason I run the job synchronously here is that the request that calls this method is sent from the Arkivo server, not a real human user, so I thought a slight delay would be acceptable.)

lib/sufia/arkivo/actor.rb
mjgiarlo: Good question.  I believe I used a TempFile because it's passed along to `add_file`, and I'm not sure that method deals with StringIO.  It expects an HttpUploadedFile, if memory serves, with an original_filename and a content_type.

app/controllers/api/items_controller.rb
```diff
@@ -0,0 +1,74 @@
+module API
+  class ItemsController < ApplicationController
```
jcoyne: This class seems specific to Arkivo, but "ItemsController" doesn't tell me much. How about adding some documentation about what this class is for.

app/controllers/api/zotero_controller.rb
```diff
@@ -0,0 +1,65 @@
+require 'oauth'
+
+module API
+  class ZoteroController < ApplicationController
```
jcoyne: Add class level documentation. What is this for.

app/helpers/sufia/sufia_helper_behavior.rb
```diff
@@ -4,6 +4,14 @@ def orcid_label(style_class='')
       "#{image_tag 'orcid.png', { alt: t('sufia.user_profile.orcid.alt'), class: style_class }} #{t('sufia.user_profile.orcid.label')}".html_safe
     end
 
+    def zotero_label(style_class='')
```
jcoyne: Have you considered putting these methods in a presenter instead of a helper?

app/views/users/_edit_primary.html.erb
```diff
@@ -7,25 +7,38 @@
       <%= f.file_field :avatar %>
       <%= f.hidden_field :avatar_cache %>
       <span class="help-block">JPG, GIF, or PNG (less than 2MB)</span>
+
```
jcoyne: Does this change have anything to do with Zotero? Why not make a separate PR?

app/views/users/_edit_primary.html.erb
```diff
@@ -7,25 +7,38 @@
...
-      <% end %>
-    </div>
-    <div class="checkbox">
-      <%= f.label :update_directory do %>
-        <%= f.check_box :update_directory %>
-        Refresh directory info
-        <a href="#" id="refresh_directory_help" rel="popover" data-content="The information displayed in your profile comes from <%=t('sufia.institution_name') %>'s central directory (LDAP) and is cached by <%=t('sufia.product_name') %>.  If you have updated that information and don't see those changes in your <%=t('sufia.product_name') %> profile, check the box and save your profile." data-original-title="Refresh Directory Info"><i class="glyphicon glyphicon-question-sign"></i></a>
-      <% end %>
+  <% if Sufia.config.arkivo_api %>
+  <div class="form-group">
```
jcoyne: I think it would be a good idea to put this whole block in a partial

app/helpers/sufia/sufia_helper_behavior.rb
jcoyne: Making this argument take a hash of options would make it easier to see that the argument is a class. e.g. `zotero_label(html_class: 'orcid')`

sufia-models/lib/generators/sufia/models/templates/config/arkivo_constraint.rb
```diff
@@ -0,0 +1,8 @@
+module Sufia
+  class ArkivoConstraint
+    def self.matches?(request)
+      # Add your own logic here to authorize trusted connections to the API
```
jcoyne: Please add an example here.

sufia-models/app/actors/sufia/generic_file/actor.rb
```diff
@@ -105,7 +105,7 @@ def save_and_record_committer
     end
 
     def push_characterize_job
-      Sufia.queue.push(CharacterizeJob.new(@generic_file.id))
+      Sufia.queue.push(CharacterizeJob.new(generic_file.id))
```
jcoyne: This seems unrelated to the declared purpose of the pull request. Can you separate this to a different PR?

app/controllers/api/items_controller.rb
hectorcorrea: Kudos for using proper HTTP status codes.

spec/spec_helper.rb
```diff
@@ -32,9 +33,6 @@
 Capybara.default_driver = :rack_test      # This is a faster driver
 Capybara.javascript_driver = :poltergeist # This is slower
 Capybara.default_wait_time = ENV['TRAVIS'] ? 30 : 15
-# HttpLogger.logger = Logger.new(STDOUT)
-# HttpLogger.ignore = [/localhost:8983\/solr/]
-# HttpLogger.colorize = false
```
jcoyne: I think leaving these here are helpful for debugging. And show developers how they could do it. You just don't want them activated.

spec/spec_helper.rb
```diff
@@ -1,13 +1,14 @@
 ENV["RAILS_ENV"] ||= 'test'
 require "bundler/setup"
+require 'coveralls'
 
+Coveralls.wear!('rails')
```
jcoyne: This seems unrelated to the Zotero PR. Please separate it into a different PR.

jcoyne:
Has there been any thought to making this a plugin/add-on gem for Sufia rather than something that ships with Sufia by default?  It's not something that everyone would use, and I think Sufia is already pretty complex and hard to test.

app/controllers/api/items_controller.rb
```diff
@@ -0,0 +1,74 @@
...
+
+      def user
+        User.find_by(arkivo_token: token)
+      end
+
+      def validate_item
+        return render plain: 'no item parameter', status: :bad_request if params[:item].blank?
+        Sufia::Arkivo::SchemaValidator.new(params[:item]).call
+      rescue Sufia::Arkivo::InvalidItem => exception
+        return render plain: "invalid item parameter: #{exception.message}", status: :bad_request
```
hectorcorrea: I like that you return a friendly error to the caller, but should we be also logging the details of the exception (e.g. message, stack trace)?

lib/sufia/arkivo/actor.rb
hectorcorrea: Wait...does the `Sufia.queue.push(BatchUpdateJob.new(...))` not run in the background because of the `::Resque.inline = true` above it? Why push it to the queue in this case? 

Why not just call `BatchUpdateJob.new(...)` directly?

mjgiarlo: That's what @cam156 just asked, too!  Heh. :)

lib/sufia/arkivo.rb
cam156: @mjgiarlo Seems like a good reason.  It is not integral to the app...

lib/sufia/arkivo/actor.rb
cam156: @mjgiarlo  BatchUpdateJob.new(...).run() should do the job

app/controllers/api/items_controller.rb
```diff
@@ -0,0 +1,74 @@
+module API
+  class ItemsController < ApplicationController
+    before_filter :validate_item, only: [:create, :update]
```
inukshuk: Mike, I had to disable `protect_from_forgery` in the demo app's (spec/instance) ApplicationController -- is this enabled by default in Sufia? If so, maybe we could add `skip_before_action :verify_authenticity_token` for JSON requests to `[:create, :update, :destroy]`.

lib/sufia/arkivo/schema_validator.rb
```diff
@@ -0,0 +1,57 @@
...
+      properties: {
+        token: { type: 'string', required: true },
+        metadata: {
+          type: 'object',
+          required: true,
+          properties: {
+            title: { type: 'string', required: true },
+            creators: { type: 'array', required: true },
+            rights: { type: 'string', required: true },
+            tags: { type: 'array', required: true },
```
inukshuk: @mjgiarlo by default Rails converts empty arrays in a JSON request to `nil` (see [here](https://github.com/rails/rails/issues/13766)) — with 4.1 and later there is a way to disable this, but I guess the easier solution is to permit nil values for `tags` and `creators` since the empty array sent by Arkivo arrives as nil. What do you think?

app/controllers/api/items_controller.rb
```diff
@@ -0,0 +1,74 @@
...
+        User.find_by(arkivo_token: token)
+      end
+
+      def validate_item
+        return render plain: 'no item parameter', status: :bad_request if params[:item].blank?
+        Sufia::Arkivo::SchemaValidator.new(params[:item]).call
+      rescue Sufia::Arkivo::InvalidItem => exception
+        return render plain: "invalid item parameter: #{exception.message}", status: :bad_request
+      else
+        @item = JSON.parse(params[:item])
```
inukshuk: @mjgiarlo this fails for me -- does JSON.parse work with Parameters as input? I worked around it with `params[:item].to_unsafe_hash` for the time being.

app/controllers/api/items_controller.rb
```diff
@@ -0,0 +1,74 @@
...
+    before_filter :authorize_token
+    before_filter :my_load_and_authorize_resource, only: [:update, :destroy, :show]
+
+    attr_reader :item
+
+    def create
+      head :created, location: sufia.api_item_path(actor.create_file_from_item)
+    end
+
+    def update
```
inukshuk: @mjgiarlo since Arkivo needs to keep track of the Zotero key to Sufia id mapping, can we return 404 on `update` and `delete` if the given id does not exist? The current response is 500 if I am not mistaken, but it will be important for Arkivo to know, for example, that an `update` failed because the item did not exist (as opposed to other errors) in order for it to send a `create` request instead. Similarly, that a `delete` failed because the item was already removed (hence the error can be ignored).

app/controllers/api/items_controller.rb
mjgiarlo: Will do.

app/controllers/api/zotero_controller.rb
mjgiarlo: Will do.

app/helpers/sufia/sufia_helper_behavior.rb
mjgiarlo: Will change to accept a hash of options instead of now.

app/views/users/_edit_primary.html.erb
mjgiarlo: Nothing to do with Zotero other than this living in the UI right next to the Zotero widget, which is why I spruced it up in this PR.  Happy to remove if you think that's best.

app/views/users/_edit_primary.html.erb
mjgiarlo: Will do.

sufia-models/lib/generators/sufia/models/templates/config/arkivo_constraint.rb
mjgiarlo: Will do.

sufia-models/app/actors/sufia/generic_file/actor.rb
mjgiarlo: I'll revert this change, :ok_hand:

spec/spec_helper.rb
mjgiarlo: I'll restore those lines.

spec/spec_helper.rb
mjgiarlo: OK.

mjgiarlo:
Quoth @jcoyne:
> Has there been any thought to making this a plugin/add-on gem for Sufia rather than something that
> ships with Sufia by default? It's not something that everyone would use, and I think Sufia is already
> pretty complex and hard to test.

Good question. Yes, in fact that was my starting point, but I was under the gun to get something running very quickly and I opted not to create that layer of abstraction rather than created a rushed, wrong layer of abstraction.  I spent a couple days trying to make that work and just couldn't, so I moved on and put it right in Sufia.

app/controllers/api/items_controller.rb
mjgiarlo: I'm on the fence about that, @hectorcorrea, mostly because our log files are already so laden with noise. I also suspect that Arkivo will already be logging errors like this. Is there any truth to that, @inukshuk?

app/controllers/api/items_controller.rb
mjgiarlo: I haven't run into this myself, @inukshuk.  Do you have any opinions about this, @jcoyne?

lib/sufia/arkivo/schema_validator.rb
mjgiarlo: OK, I'll relax the requirements for those elements, @inukshuk

app/controllers/api/items_controller.rb
mjgiarlo: @inukshuk `params[:item]` here should be JSON in a string, if memory serves. What error are you seeing? Can you post a traceback or error message?

app/controllers/api/items_controller.rb
mjgiarlo: @inukshuk I thought we were tossing a 404 in responses to `update` and `delete` if the specified Sufia ID it not found.  See here: https://github.com/projecthydra/sufia/pull/1022/files#diff-b2d6cad020de1805df8ddd050e8e159fR40  If that's not catching the exception and instead a 500 is being raised, that'd be good to know. If so, can you tell me how I can reproduce this?

mjgiarlo: @inukshuk If you check out `spec/controllers/api/items_controller_spec.rb` you'll see three tests that make sure a 404 is returned when if a specified ID is not found on `show`, `update`, and `destroy`.

app/controllers/api/items_controller.rb
inukshuk: @mjgiarlo you can configure Arkivo to log a plugin's output and we can obviously add all the information you want to the Sufia plugin. Having said that, this wouldn't print the server side stack trace obviously (though in this particular case of invalid items the exception message is probably sufficient).

app/controllers/api/items_controller.rb
inukshuk: @mjgiarlo have you tested posting to the API from the 'outside'? The Rails  test environment sets `config.action_controller.allow_forgery_protection = false`, but in production this defaults to true if I am not mistaken.

lib/sufia/arkivo/schema_validator.rb
inukshuk: Is there a way to specify type is 'array' or 'null'? Just setting required to false did not fix it for me, because Rails actually inserts a null value there.

app/controllers/api/items_controller.rb
inukshuk: @mjgiarlo Here is the error I'm getting (probably using a different JSON gem by default):

    TypeError (no implicit conversion of ActionController::Parameters into String):
      json (1.8.2) lib/json/common.rb:155:in `initialize'
      json (1.8.2) lib/json/common.rb:155:in `new'
      json (1.8.2) lib/json/common.rb:155:in `parse'
      /home/vagrant/sufia/app/controllers/api/items_controller.rb:63:in `validate_it
em'

app/controllers/api/items_controller.rb
inukshuk: @mjgiarlo you're right; when trying to reproduce I'm getting 404s now... I guess I was running into an unrelated issue earlier, but will keep an eye open for this.

lib/sufia/arkivo/schema_validator.rb
mjgiarlo: I don't believe json-schema has the concept of a disjunctive type, unfortunately, no.

app/controllers/api/items_controller.rb
mjgiarlo: I haven't, no!  Thanks.

lib/sufia/arkivo/actor.rb
mjgiarlo: OK!

app/helpers/sufia/sufia_helper_behavior.rb
mjgiarlo: Will clarify.

README.md
```diff
@@ -6,7 +6,6 @@
 [![API Docs](http://img.shields.io/badge/API-docs-blue.svg)](http://rubydoc.info/gems/sufia)
 [![Build Status](https://travis-ci.org/projecthydra/sufia.png?branch=master)](https://travis-ci.org/projecthydra/sufia)
 [![Dependency Status](https://gemnasium.com/projecthydra/sufia.png)](https://gemnasium.com/projecthydra/sufia)
-[![Coverage Status](https://img.shields.io/coveralls/projecthydra/sufia.svg)](https://coveralls.io/r/projecthydra/sufia?branch=master)
```
cam156: Think this is a mistake...

app/controllers/api/items_controller.rb
```diff
@@ -0,0 +1,78 @@
...
+    end
+
+    def show
+      head :no_content
+    end
+
+    private
+
+      def my_load_and_authorize_resource
+        @file = GenericFile.find(params[:id])
```
cam156: May want a load instance from solr here instead of a find

README.md
mjgiarlo: @jcoyne pointed out that coveralls is orthogonal to this branch, so I removed it.

app/controllers/api/items_controller.rb
mjgiarlo: Will do.

mjgiarlo: Er, strike that. I need this instance variable to be loaded from Fedora.

app/helpers/sufia/sufia_helper_behavior.rb
```diff
@@ -4,6 +4,15 @@ def orcid_label(style_class='')
       "#{image_tag 'orcid.png', { alt: t('sufia.user_profile.orcid.alt'), class: style_class }} #{t('sufia.user_profile.orcid.label')}".html_safe
     end
 
+    def zotero_label(opts={})
+      html_class = opts[:html_class] || ''
+      "#{image_tag 'zotero.png', { alt: t('sufia.user_profile.zotero.alt'), class: html_class }} #{t('sufia.user_profile.zotero.label')}".html_safe
+    end
+
+    def zotero_profile_url(zotero_user_id)
+      "https://www.zotero.org/users/#{zotero_user_id}"
```
cam156: Should this url be configurable? or dependent on the options in the controller?

app/views/users/_user_info.html.erb
```diff
@@ -5,6 +5,11 @@
   <dd class="col-xs-7"><%= link_to user.orcid, user.orcid, { target: '_blank' } %></dd>
 <% end %>
 
+<% if user.zotero_userid.present? %>
+  <dt class="col-xs-5"><%= zotero_label(html_class: 'orcid') %></dt>
```
cam156: Should there be a separate class here?  or rename orcid?

app/views/users/_zotero.html.erb
```diff
@@ -0,0 +1,12 @@
+  <div class="form-group">
+    <%= f.label :zotero, class: 'col-xs-4 control-label' do %>
+      <%= zotero_label %>
+    <% end %>
+    <div class="col-xs-8">
+    <% if user.zotero_userid.present? %>
+       <%= link_to "Connected!", zotero_profile_url(user.zotero_userid), class: "form-control" %>
+    <% else %>
+       <%= link_to "Link with Zotero", sufia.api_zotero_initiate_path, id: "zotero", class: "btn btn-mini btn-primary" %>
```
cam156: I18N the label here if you please...

lib/sufia/arkivo/actor.rb
```diff
@@ -0,0 +1,95 @@
...
+      end
+
+      def destroy_file(gf)
+        gf.destroy
+      end
+
+      private
+
+        def reset_metadata(actor)
+          actor.generic_file.tap do |gf|
```
cam156: I wonder if this is an action that should be pushed to the GF actor

app/helpers/sufia/sufia_helper_behavior.rb
mjgiarlo: Nope. This is how all Zotero profile URLs work.

lib/sufia/arkivo/create_subscription_job.rb
```diff
@@ -0,0 +1,58 @@
...
+        subscription_path = response.headers['Location']
+        # create subscription
+        user.arkivo_subscription = subscription_path
+        user.save
+      end
+
+      private
+
+      def validate_user!
+        raise SubscriptionError.new('User not found') if user.blank?
```
cam156: how are these errors getting fed back into the system?

lib/sufia/arkivo/metadata_munger.rb
```diff
@@ -0,0 +1,51 @@
...
+        # Then, rename the url key to related_url
+        munged['related_url'] = munged.delete('url') if munged['url']
+
+        # Then, rename the tags key to tag
+        munged['tag'] = munged.delete('tags') if munged['tags']
+
+        # Then, normalize creator names
+        munged_creators = munged['creators'].each do |entry|
+          next if entry['name']
+          entry['name'] = "#{entry.delete('firstName')} #{entry.delete('lastName')}".strip
```
cam156: should we stick with last, first as recommended in SS?

app/views/users/_user_info.html.erb
mjgiarlo: Will rename orcid.

app/views/users/_zotero.html.erb
mjgiarlo: Done.

lib/sufia/arkivo/actor.rb
mjgiarlo: Until we have another use for wiping out metadata *en masse*, I think we should keep this in the Arkivo actor.

lib/sufia/arkivo/create_subscription_job.rb
mjgiarlo: I'll log them.

lib/sufia/arkivo/metadata_munger.rb
mjgiarlo: Done.

README.md
jcoyne: Yes, but this is already in master. You shouldn't be changing it.

lib/sufia/arkivo/actor.rb
cam156: :+1:

sufia-models/lib/tasks/sufia-models_tasks.rake
```diff
@@ -8,6 +8,16 @@ namespace :solr do
 end
 
 namespace :sufia do
+  namespace :user do
+    desc 'Populate user tokens'
+    task tokens: :environment do
```
cam156: Should this task just exit in arkivo_api is false

spec/lib/sufia/arkivo/schema_validator_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+  it 'ensures a file section is included' do
+    expect {
+      described_class.new(item.except('file')).call
+    }.to raise_error(Sufia::Arkivo::InvalidItem, /required property of 'file'/)
+  end
+
+  it 'ensures the metadata has a title' do
+    item['metadata'].delete('title')
+    expect {
+      described_class.new(item).call
```
cam156: Would item.except(title) work here.  Just asking for consistency

README.md
mjgiarlo: I suspect this went into master after I first created the PR.  So originally the request was to remove Coveralls-related stuff, and now it's to put this line back because it's in master, right?  If so, makes sense.

spec/lib/sufia/arkivo/schema_validator_spec.rb
mjgiarlo: I don't believe it would.

sufia-models/lib/tasks/sufia-models_tasks.rake
mjgiarlo: OK

grosscol:
I tested this with my sufia based apps.  Without it being enabled, it didn't affect the applications.  Enabling it functioned as expected.

mjgiarlo:
:clap: @grosscol 

Thanks much.

PR #1029, opened 2015-04-07
Prevent user profile page from crashing if a file listed as a trophy 

hectorcorrea:
does not exist in Fedora. We log the error instead.


jcoyne:
:+1:

PR #1034, opened 2015-04-10
Update message when seeing who is following a user or who is the user 

hectorcorrea:
following to differentiate between the current user checking their profile versus another user's profile

### DIFF
--- a/app/views/users/_follower_modal.html.erb
+++ b/app/views/users/_follower_modal.html.erb
@@ -18,7 +18,11 @@
               <% end %>
             </ul>
           <% else %>
-            <h3> <%= t("sufia.user_profile.no_followers") %></h3>
+            <% if @user.id == current_user.id %> 
+                <h3> <%= t("sufia.user_profile.no_followers") %></h3>
+            <% else %>
+                <h3> <%= t("sufia.user_profile.no_followers_other") %></h3>
+            <% end %>
           <% end %>
         </div>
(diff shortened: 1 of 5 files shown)


app/views/users/_follower_modal.html.erb
```diff
@@ -18,7 +18,11 @@
               <% end %>
             </ul>
           <% else %>
-            <h3> <%= t("sufia.user_profile.no_followers") %></h3>
+            <% if @user.id == current_user.id %> 
+                <h3> <%= t("sufia.user_profile.no_followers") %></h3>
+            <% else %>
+                <h3> <%= t("sufia.user_profile.no_followers_other") %></h3>
+            <% end %>
```
awead: I would put this in a helper to avoid the duplication.

app/views/users/_following_modal.html.erb
```diff
@@ -18,7 +18,11 @@
               <% end %>
             </ul>
           <% else %>
-            <h3><%= t("sufia.user_profile.no_following") %></h3>
+            <% if @user.id == current_user.id %> 
+                <h3> <%= t("sufia.user_profile.no_following") %></h3>
+            <% else %>
+                <h3> <%= t("sufia.user_profile.no_following_other") %></h3>
+            <% end %>
```
awead: ah. nm. it's not duplicated... it just looks very similar.

PR #1037, opened 2015-04-10
Use ActiveFedora::Noid 1.0 and surface its config in the Sufia config

mjgiarlo:


mjgiarlo:
@jcoyne @terrellt others?

terrellt:
@jechols ?

sufia-models/lib/sufia/models/engine.rb
```diff
@@ -30,6 +28,14 @@ class Engine < ::Rails::Engine
       config.max_notifications_for_dashboard = 5
       config.activity_to_show_default_seconds_since_now = 24*60*60
 
+      # Noid identifiers
+      config.enable_noids = true
+      config.noid_template = '.reeddeeddk'
+      config.minter_statefile = '/tmp/minter-state'
```
jechols: I'd recommend someplace more permanent for people wanting noid minting to be fairly guaranteed to not repeat

jechols:
My only concern is the problem with the treeifier - unless noids are truly random (on our system they appeared to be pretty predictable), using id causes directories to fill up before new ones are created, which is pretty awful if Fedora wants fairly well-distributed assets.  But as I mentioned in the noid project (https://github.com/projecthydra-labs/active_fedora-noid/issues/5), there are problems with the treeifier "fix" I pushed up.

If this is the definitive turnkey project, I think that issue needs to be figured out :-/

sufia-models/lib/sufia/models/engine.rb
mjgiarlo: This PR does not change the default `minter_statefile`; it just reorganizes Noid-related configs into a contiguous block. I'm reluctant to change the default value because `/tmp` is a well known location that is likely to exist in most environments and that is writeable by all users. If you have a different default location to recommend, I am cool with that, but I'm thinking you should do that in a separate PR since it's unrelated to what I'm trying to do with this PR which is to update to AF::Noid 1.0 and retain the previous treeifying behavior.

mjgiarlo:
@jechols I agree that this needs to be worked out!  Until we have a workable treeifier solution, and I would welcome your thoughts and code since you've done the most work on this so far, I'm aiming here to update to AF::Noid 1.0 and retain the previous treeifying behavior. Could this other treeifier work follow #1037 once it's merged?

sufia-models/sufia-models.gemspec
```diff
@@ -30,10 +30,12 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'activeresource', "~> 4.0" # No longer a dependency of rails 4.0
 
   spec.add_dependency "hydra-head", "~> 9.0"
-  spec.add_dependency "active-fedora", "~> 9.0"
+  # AF 9.0.7 was yanked and I can't get bundler to understand this
+  # TODO: Remove this line when 9.0.8 is released
+  spec.add_dependency "active-fedora", "< 9.0.7"
```
mjgiarlo: @jcoyne Is this :ok: for now, or should I relax this?

jechols:
Yeah, there's really nothing stopping this from merging.  The real issue is that once a particular algorithm is chosen for treeification, there's no changing it without a ton of work.  Anybody using the current algorithm is kind of stuck with it.

I wish I had a good solution to the problem, I just haven't come up with anything yet.  It's something we have to solve for ourselves for sure.

mjgiarlo:
OK, thanks, @jechols.

@jcoyne @terrellt Sounds like we're good to merge, other input notwithstanding.

jcoyne:
@jechols I'm not too concerned with the unbalanced distribution because the default 2-digit treeifier should guarantee that there are no more than 100 child nodes in any one directory. The problem as I understand it is with directory that contain a large number of children, not an unbalanced tree.

jechols:
@jcoyne That's true of the bottom-most directory when using the default template.

If the template is short, but ends in 'ee', the bottom-most directory contains about 900 items.  Is that a problem?  I have no idea, but it's certainly not optimal.

If the template is long, say `reeddeeddeedd`, the bottom-most directory can end up with over 90,000 items.  The longer the template, the more assets will go into a single place.

Even with the default template, the second-to-bottom directory will be based on extended digits, which would make for about 900 sub-directories per third-level directory.

Note that this could theoretically be addressed by simply making a noid minter that reverses the minted value.  I imagine that breaks the NOID spec, but it would make for moderately balanced asset distribution.

sufia-models/sufia-models.gemspec
awead: @mjgiarlo 9.1 just went out today, so you can update this.

mjgiarlo:
Pls do not merge yet, folks.  Discussion still ongoing, and haven't had time to respond yet. Thx.

mjgiarlo:
Finally coming back to this, and I'll rebase.  I'm in favor of moving forward with this PR which keeps Sufia up to date with AF::Noid and preserves existing behavior.

jcoyne:
I don't think it's possible to use the default af-noid 1.0 to work with Sufia because not all resources have NOIDs as identifiers. Some are just using Fedora's built in treeifier which operates much like the treeifier in af-noid 0.2. ActiveFedora currently only allows a single mapping to and from id <-> uri.

mjgiarlo:
After discussing this with @jcoyne on IRC, we think we should tackle https://github.com/projecthydra-labs/active_fedora-noid/issues/5 first and then revisit this PR.

mjgiarlo:
Ping @jechols @terrellt re: :arrow_up:

jechols:
Sorry, didn't notice this until today.

So the way we solved the issue for OregonDigital was in fact by reversing the NOID and using that as the identifier.  Yes, the official and final ID for an asset is a reversed NOID :(  From there, we just letting the default treeifier do its thing.  It feels wrong, but it ensures that buckets work well no matter the pid length, and it's backwards-compatible with any possible identifier that has the same general bucketing strategy.

I hesitate to recommend that as a general strategy for obvious reasons.  I'd love to hear a better solution, especially since we still haven't gone live with OD 2 :)

Somebody shoot me if I say "bucket" one more time.

jechols:
Actually, I'll put that comment into the referenced issue, too

mjgiarlo:
I'm working on a related fix upstream. Stay tuned. https://github.com/microservices/noid/pull/13

mjgiarlo:
Also related: I have yanked AF::Noid 1.0.0.

mjgiarlo:
This is ready for review now, @jcoyne @terrellt @jechols. (Travis is still running... :watch:)

mjgiarlo:
While troubleshooting this issue, I stumbled into https://github.com/projecthydra-labs/active_fedora-noid/issues/16.

mjgiarlo:
As soon as https://github.com/projecthydra-labs/active_fedora-noid/pull/17 is merged and I cut the AF::Noid 1.0.2 release I'll push updated commits here.

mjgiarlo:
:arrow_up: is done. Ready for review.

jechols:
I don't see any reason this should wait any longer at this point.  My only concern initially was the treeifier situation, which has been fixed nicely via your noid fix.

:+1:

PR #1052, opened 2015-04-24
Uses correct file date when collecting analytics

hectorcorrea:
This handles the situation in which the "created date" for a file is more recent than the "date uploaded" which happens for data migrated from Fedora 3 to Fedora 4. For those files, the "created date" reflects the date the file was migrated to Fedora 4 rather than the date the file was originally uploaded to the system (in Fedora 3).

This PR makes sure we prefer the "date uploaded" if available.


awead:
:clap: awesome!

Issue #996, opened 2015-03-26
Create a Generic Work


cam156:
It should have many GenericFiles

* Does not need to use PCDM so that we can move forward with sufia development and views

cam156:
blah

mtribone:
Is this related to #998?

cam156:
Use has_members and belnongs_to relationships

cam156:
This is a model level implementation of a Work.  This may be an opportunity to come close to worthwhile by grabbing their implementation of a work at the model level and add it into sufia: https://github.com/projecthydra-labs/worthwhile/blob/master/worthwhile-models/app/models/generic_work.rb

PR #1077, opened 2015-05-01
Adding in work concept from hydra-work before fresh start commit

cam156:
See #996 

This is pulled from the hydra-works code in this state: https://github.com/projecthydra-labs/hydra-works/tree/a9ed2e003f35403d4b03b738e36d33c5ca7642ef

I grabbed a test out of worthwhile also, which lead to including a couple more curation concerns.

### DIFF
--- a/sufia-models/app/models/sufia/works/curation_concern/with_basic_metadata.rb
+++ b/sufia-models/app/models/sufia/works/curation_concern/with_basic_metadata.rb
@@ -1,0 +1,35 @@
+# Basic metadata for all Works
+# Required fields:
+#   dc:title
+#   dc:rights
+#
+# Optional fields:
+#   dc:contributor
+#   dc:coverage
+#   dc:creator
+#   dc:date
+#   dc:description
+#   dc:format
+#   dc:identifier
+#   dc:language
+#   dc:publisher
+#   dc:relation
+#   dc:source
+#   dc:subject
+#   dc:type
+module Sufia::Works
+  module CurationConcern
+    module WithBasicMetadata
+      extend ActiveSupport::Concern
+
+      included do
+        include GenericWorkRdfProperties
+        # Validations that apply to all types of Work AND Collections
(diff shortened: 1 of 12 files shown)


awead:
are you proposing this go into the pcdm branch?

sufia-models/app/models/sufia/works/curation_concern/with_basic_metadata.rb
```diff
@@ -0,0 +1,35 @@
...
+#   dc:source
+#   dc:subject
+#   dc:type
+module Sufia::Works
+  module CurationConcern
+    module WithBasicMetadata
+      extend ActiveSupport::Concern
+
+      included do
+        include GenericWorkRdfProperties
```
awead: This ought to metadata.rb

sufia-models/app/models/sufia/works/work.rb
```diff
@@ -0,0 +1,7 @@
+# The base class of all works
```
awead: I'm unclear why you're distinguishing between Work and GenericWork.

cam156: I closed this PR since it was against the wrong branch.  That said WorthWhile and Curate both have a concept of many work types.  I left that in.

sufia-models/app/models/sufia/works/curation_concern/with_basic_metadata.rb
cam156: Unclear what you mean.  The files should be renamed, or sufia already has the concept?

awead: I meant we could use the existing GenericFile::Metadata class.

sufia-models/app/models/sufia/works/work.rb
jcoyne: @awead Work is an abstract class. GenericWork is the simplest concrete type of Work.

Issue #1007, opened 2015-03-26
Convert GenericWork to PCDM


cam156:
Convert the GenericWork (created in ticket #996) to model the Hydra::PCDM::Object

Use the new structure with indirect containers so the correct LDP will get generated.

dependent on projecthydra-labs/hydra-pcdm#4

This ticket includes converting the mock generic work (ticket #996) to the new official class. We will assume that the hydra class will have been implemented and tested.

mjgiarlo:
Done.

PR #1078, opened 2015-05-01
Adding in work concept from hydra-work before fresh start commit. refs #...

cam156:
...996

This is code taken from hydra-works pre reset: https://github.com/projecthydra-labs/hydra-works/tree/a9ed2e003f35403d4b03b738e36d33c5ca7642ef

Added a test from worthwhile also, which caused me to add a couple more concerns.

### DIFF
--- a/sufia-models/app/models/sufia/works/curation_concern/with_basic_metadata.rb
+++ b/sufia-models/app/models/sufia/works/curation_concern/with_basic_metadata.rb
@@ -1,0 +1,32 @@
+# Basic metadata for all Works
+# Required fields:
+#   dc:title
+#   dc:rights
+#   dc:relation (tag/ key term)
+#   dc:creator
+#
+#
+# Optional fields:
+#   dc:contributor
+#   dc:coverage
+#   dc:date
+#   dc:description
+#   dc:format
+#   dc:identifier
+#   dc:language
+#   dc:publisher
+#   dc:source
+#   dc:subject
+#   dc:type
+module Sufia::Works
+  module CurationConcern
+    module WithBasicMetadata
+      extend ActiveSupport::Concern
+
+      included do
+        include GenericWorkRdfProperties
+      end
+
+    end
+  end
+end
--- a/sufia-models/app/models/sufia/works/curation_concern/with_generic_files.rb
+++ b/sufia-models/app/models/sufia/works/curation_concern/with_generic_files.rb
@@ -1,0 +1,18 @@
+module Sufia::Works
+  module CurationConcern
+    module WithGenericFiles
+      extend ActiveSupport::Concern
+
+      included do
+        # This used to have a hasFile relation when in hydra-works.  That does not seem to exist so I am using hasPart instead
+        has_many :files, predicate: ActiveFedora::RDF::Fcrepo::RelsExt.isPartOf, class_name: "GenericFile"
+        #has_many :generic_files, predicate: ActiveFedora::RDF::Fcrepo::RelsExt.isPartOf
+        before_destroy :before_destroy_cleanup_generic_files
+      end
+
+      def before_destroy_cleanup_generic_files
+        files.each(&:destroy)
+      end
+    end
+  end
+end
--- a/sufia-models/app/models/sufia/works/curation_concern/work.rb
+++ b/sufia-models/app/models/sufia/works/curation_concern/work.rb
@@ -1,0 +1,13 @@
+module Sufia::Works
+  module CurationConcern
+    module Work
+      extend ActiveSupport::Concern
+      include WithGenericFiles
+      include HumanReadableType
+      include Sufia::Noid
+      include Sufia::ModelMethods
+      include Hydra::Collections::Collectible
+      include Solrizer::Common
+    end
+  end
+end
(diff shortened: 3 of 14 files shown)


awead:
What about also including the relationship between GenericFile and GenericWork, as here:
https://github.com/projecthydra-labs/hydra-works/blob/a9ed2e003f35403d4b03b738e36d33c5ca7642ef/app/models/hydra/works/generic_file.rb
Or is that out of scope for this ticket?

cam156:
@awead  Good idea.  I will add it in

awead:
@cam156 :clap: looking good! I'm starting to have more questions now about what functions should get moved from GenericFile to GenericWork, such as trophies, featured, etc. but that's probably for other tickets. Do we want to merge this now and move on?

cam156:
@awead Yes those items get moved.  I thin we have tickets for each of those, but it would be good to double check the list.

Yes I think merging and moving on would be good if there are no other changes to what is there.

cam156:
@awead I double checked.  We had Featured Work, but not Highlighted Work.  I added a tciket.

sufia-models/app/models/sufia/works/curation_concern/with_basic_metadata.rb
```diff
@@ -0,0 +1,35 @@
+# Basic metadata for all Works
+# Required fields:
```
mjgiarlo: Are we saying that keywords and creator will no longer be required?  The documentation here is in conflict with a longstanding Sufia feature.

sufia-models/app/models/sufia/works/curation_concern/with_basic_metadata.rb
```diff
@@ -0,0 +1,35 @@
...
+#   dc:type
+module Sufia::Works
+  module CurationConcern
+    module WithBasicMetadata
+      extend ActiveSupport::Concern
+
+      included do
+        include GenericWorkRdfProperties
+        # Validations that apply to all types of Work AND Collections
+        validates_presence_of :title,  message: 'Your work must have a title.'
```
mjgiarlo: This is OK for now, but we should square this with our other approach to metadata validation in Sufia (by using hydra-editor forms rather than model validations).

sufia-models/app/models/sufia/works/curation_concern/with_generic_files.rb
```diff
@@ -0,0 +1,17 @@
+module Sufia::Works
+  module CurationConcern
+    module WithGenericFiles
+      extend ActiveSupport::Concern
+
+      included do
+        # This used to have a hasFile relation when in hydra-works.  That does not seem to exist so I am using hasPart instead
```
mjgiarlo: Is this comment worthy of a TODO or a new ticket, or do we already have a ticket that wraps this up?

sufia-models/app/models/sufia/works/curation_concern/with_generic_files.rb
```diff
@@ -0,0 +1,17 @@
...
+        has_and_belongs_to_many :files, predicate: ActiveFedora::RDF::Fcrepo::RelsExt.hasPart, class_name: "GenericFile"
```
mjgiarlo: Works belong to files?  Shouldn't this be `has_many`?

sufia-models/app/models/sufia/works/curation_concern/with_generic_files.rb
cam156: There are already tickets for changing the object to PCDM (#1007).  I figured that was enough...

sufia-models/app/models/sufia/works/curation_concern/with_generic_files.rb
jcoyne: Nope, this is correct. Works point at their files.

sufia-models/app/models/sufia/works/curation_concern/with_generic_files.rb
mjgiarlo: Yep, that's fine!

sufia-models/app/models/sufia/works/curation_concern/with_editors.rb
```diff
@@ -0,0 +1,45 @@
+module Sufia::Works
+  module CurationConcern
+    module WithEditors
```
jcoyne: I don't think we want this. Can we throw it out?

sufia-models/app/models/sufia/works/curation_concern/work.rb
```diff
@@ -0,0 +1,14 @@
+module Sufia::Works
+  module CurationConcern
+    module Work
+      extend ActiveSupport::Concern
+      include WithGenericFiles
+      include WithEditors
```
jcoyne: Please remove WithEditors. I don't think we need that.

sufia-models/app/models/sufia/works/curation_concern/with_generic_files.rb
mjgiarlo: Oh, right, the semantics of AF habtm are not identical to AR habtm.

sufia-models/app/models/sufia/works/curation_concern/with_basic_metadata.rb
cam156: Added the other required fields.  Thanks for noting my copy paster error!

Issue #995, opened 2015-03-26
Filter Catalog to show Generic Work and not Generic File


cam156:
Modify catalog controller to return GenericWorks

This is advanced because there will be many specs that break and will need to be rewritten to display GenericFile vs GenericWork
This is a good opportunity to make our specs faster... rewrite feature tests as view test maybe? and more stubbing of things that are not tested

PR #1083, opened 2015-05-05
Filter catalog to show GenericWorks instead of GenericFiles. Story #995

val99erie:
Ping @cam156 for code review, since you are working on the controller story.

Some notes about my changes:
* I marked the full-text search spec as 'skip'.  I felt like that work was out-of-scope for this story.

* I removed some of the tests from spec/features/catalog_search_spec.rb because they seemed to be testing the same functionality as the tests in spec/controllers/catalog_controller_spec.rb.

* The "many tags" spec in spec/features/catalog_search_spec.rb seemed to be testing the same behavior as spec/features/browse_catalog_spec.rb, so I removed the duplicate spec from the catalog_search_spec.rb.

* I had to add some (temporary) placeholder code in order to get specs passing, for example, I added a show route and empty controller class for GenericWorks, and a placeholder view template for _generic_work.html.erb.


cam156:
@val99erie I added #1084 to cover full text searching
Looks good!

PR #1085, opened 2015-05-05
Update README with version information for setting up the dependencies.

rotated8:


cam156:
+1  Waiting on Travis (although that seems silly for readme)

rotated8:
The last error we have is because of imagemagick's JPEG-2000 support, it seems. It can be fixed by uninstalling imagemagick and `brew install graphicsmagick --with-jasper`, but I don't want to recommend that.

rotated8:
You can see the difference in FITS' output between 0.6.2 and 0.8.4 here: https://gist.github.com/rotated8/09afa8de929d245d6677

PR #1087, opened 2015-05-06
Add worthwhile work controllers in sufia

cam156:
This is an initial cut at grabbing the controller from worthwhile for the works model in sufia.

Feel free to hammer it.

If there are additional test I should have pulled over please let me know.


scherztc:
Should we hold off on merging this and instead use it as a reference for our routing/controller/model needs in our own fix branches?

scherztc:
And possibly use it as a conversation place for issues around these topics?

cam156:
@scherztc I think merging it may help with the other tickets, but we should discuss it.

scherztc:
Glen and I have been discussing it this morning and rather than cut/paste code from your branch we came to the agreement that it would be better to merge this in, also.  I will merge it.

PR #1088, opened 2015-05-06
Add Create Work Button to Dashboard Fixes #1069

### LINKED ISSUE #1069: Add Create Work Button To Dashboard
Routes to #1067 Work Creation page

See wireframe: https://wiki.duraspace.org/display/hydra/Sufia+-+Dashboard

scherztc:
screen shot 2015-05-06 at 5 10 21 pm

### DIFF
--- a/app/views/dashboard/_index_partials/_heading_actions.html.erb
+++ b/app/views/dashboard/_index_partials/_heading_actions.html.erb
@@ -1,30 +1,39 @@
 <div class="col-xs-12 heading-row">
-  <% if can?(:create, GenericFile) %>
-    <div class="col-xs-6 col-sm-3 heading-tile">
-      <%= link_to sufia.new_generic_file_path do %>
+  <% if can?(:create, Sufia::Works::GenericWork) %>
+    <div class="col-xs-6 col-sm-2 heading-tile">
+      <%= link_to sufia.new_generic_work_path do %>
           <span class="glyphicon glyphicon-upload"></span>
-          <%= t("sufia.dashboard.upload") %>
+          <%= I18n.t("sufia.dashboard.create_work") %>
       <% end %>
     </div>
   <% end %>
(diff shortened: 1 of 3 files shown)


scherztc:
Running full suite locally now.  Will clean up.  Sorry.

scherztc:
"Upload" is the starting point for various feature test.  I am working through each of the features, but we may need to get the Create Work page complete before these feature test will pass.

app/views/dashboard/_index_partials/_heading_actions.html.erb
```diff
@@ -1,9 +1,9 @@
 <div class="col-xs-12 heading-row">
-  <% if can?(:create, GenericFile) %>
+  <% if can?(:create, Sufia::Works::GenericWork) %>
     <div class="col-xs-6 col-sm-3 heading-tile">
-      <%= link_to sufia.new_generic_file_path do %>
+      <%= link_to sufia.new_generic_work_path do %>
           <span class="glyphicon glyphicon-upload"></span>
-          <%= t("sufia.dashboard.upload") %>
+          <%= t("sufia.dashboard.create_work") %>
```
cam156: Can you add a default I18N translation in for this.

cam156:
@scherztc I wonder if the simpler step would be to Add the button and leave the upload button in place.

scherztc:
I can get the test to pass once we get the Create Work page with the Batch Upload link.  If we remove the upload button from the dashboard then this link is the only path to the traditional upload.  I guess this is a good thing and bad thing.  I would defer the decision about two buttons (Create Work and Upload) to  the current implementors of sufia.

cam156:
@scherztc I think leave in both button now since it is simpler and allows this work to be merged without the other work happening.  We let the users look at it and see if the want to ditch the first button.

PR #1090, opened 2015-05-06
Fixes #993

### LINKED ISSUE #993: Descriptive Metadata for Generic Works
Copy the descriptive metadata from GenericFile into the new GenericWork class.

yinlinchen:
Copy the descriptive metadata from GenericFile into the new GenericWork class.

### DIFF
--- a/spec/models/works/generic_work_spec.rb
+++ b/spec/models/works/generic_work_spec.rb
@@ -41,5 +41,29 @@ describe Sufia::Works::GenericWork do
       expect { work.destroy }.to change{ GenericFile.count }.by(-2)
     end
   end
+
+  describe "metadata" do
+    it "should have descriptive metadata" do
+      expect(subject).to respond_to(:relative_path)
+      expect(subject).to respond_to(:depositor)
+      expect(subject).to respond_to(:related_url)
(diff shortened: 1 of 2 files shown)


sufia-models/app/models/sufia/works/generic_work.rb
```diff
@@ -3,5 +3,6 @@ module Sufia::Works
   class GenericWork < Work
     include Sufia::Works::CurationConcern::WithBasicMetadata
     include Sufia::GenericFile::Permissions
+    include Sufia::GenericFile::Metadata
```
cam156: I think we need only this line or 4, but not both.  I would suggest removing one and see if the tests still pass.

spec/models/works/generic_work_spec.rb
```diff
@@ -41,5 +41,29 @@ class TestWork < Sufia::Works::GenericWork
       expect { work.destroy }.to change{ GenericFile.count }.by(-2)
     end
   end
+
+  describe "metadata" do
+    it "should have descriptive metadata" do
```
cam156: Nice :clap:

cam156:
:+1: Just waiting on travis

Issue #1079, opened 2015-05-01
Change Highlighted Files to Highlighted Works


cam156:
Show Highlighted works in the My Highlights List and allow a Highlight action on Works.

This will include moving the highlight action from files to works in the controllers.

Additionally Highlights on the user profile will need to show works.

cam156:
This is still be in the code (not at the UI) as trophy.  So instance of trophies or highlighted files both need to move to works.

tingtingjh:
Is there a separate ticket involving remodeling trophy in generic file? Then this ticket will have dependency on that. Or this ticket involves both UI and remodeling?

tingtingjh:
Need to make a new migration for trophies to be included in ./sufia_models/lib/generators/sufia/models/upgrade700_generator.rb, as pcdm migration is likely to be in sufia 7.0

Issue #1050, opened 2015-04-23
Remove Batch Edit from Dashboard


hectorcorrea:
Remove the option to perform a Batch Edit from the Dashboard (and other file listing places) until we decide how this is going to work once the PCDM has been fully implemented. See also #1003

Issue #1011, opened 2015-03-26
Change Transfer to transfer Works instead of files


cam156:
Change the transfer to allow transferring of works not files

cam156:
dependent on #996

cam156:
Transferring all files and the work to the new user.   This means depositor would be changed on all files and the work.

val99erie:
In PR #1093, I added skips for several specs because the specs were failing and the transfer work wasn't done yet.

For this story please either fix these specs, or delete them if they are no longer appropriate.  The specs are in this file:
spec/features/ownership_transfer_spec.rb

grosscol:
I fixed the ownership transfer specs in the course of this work.  However, the specs that deal with the UI of transfering a work have all but gutted and replaced with pending as the UI was not implemented.  I plan on undertaking the UI of transferring works next.

Issue #1003, opened 2015-03-26
How does Batch Edit behave with works


cam156:
How will Batch edit work.  Will it edit GenericFiles or will it now edit works?

cam156:
Batch editing of the work should be configurable on what level the metadata cascades down to.
Title and a configurable list should be filtered from being editable

hectorcorrea:
A possible outcome would be not to allow Batch Edit on works.

hectorcorrea:
This ticket is for the planning meeting, not an implementation.

mjgiarlo:
Working together to define a new batch editing UI is going to be a high priority in 2016. Since this issue is a high-level question, `How will Batch edit work. Will it edit GenericFiles or will it now edit works?`, I'm inclined to answer that question and close this ticket. Batch-edit will modify descriptive metadata on works in bulk.

PR #1093, opened 2015-05-06
Show works instead of files on 'My Files' dashboard page. Story #999

val99erie:
Here is how the page looks now, including "TODO" placeholders:

my_files_dashboard

I assumed that the following things were out-of-scope for this story.  I put TODO placeholders in the code, and I'll write new stories for these things if stories don't already exist for them:

  * Transfer of ownership of a work (Story #1011)
  * Visibility link
  * The 'Click for more details' toggle
  * Batch Controls
      * Add to Collection
      * Batch delete
      * Batch edit (Story #1050 & #1003)
  * Highlight work on profile (Story #1079)
  * Download a work
  * 'Single Use Link' for a work
  * List collections that the work belongs to (depends on how solr indexing is done)

### DIFF
--- a/spec/actors/curation_concern/generic_work_actor_spec.rb
+++ b/spec/actors/curation_concern/generic_work_actor_spec.rb
@@ -147,7 +147,6 @@ describe Sufia::CurationConcern::GenericWorkActor do
 
       it "should add to collections" do
         reload = Sufia::Works::GenericWork.find(curation_concern.id)
-        puts curation_concern.collection_ids
         expect(reload.collections).to eq [collection1]
 
         expect(subject.update).to be true
--- a/app/controllers/my/files_controller.rb
+++ b/app/controllers/my/files_controller.rb
@@ -3,7 +3,7 @@ module My
 
     self.search_params_logic += [
       :show_only_resources_deposited_by_current_user,
-      :show_only_generic_files
+      :show_only_generic_works
     ]
 
     def index
--- a/app/search_builders/sufia/search_builder.rb
+++ b/app/search_builders/sufia/search_builder.rb
@@ -24,6 +24,13 @@ module Sufia::SearchBuilder
     ]
   end
 
+  def show_only_generic_works(solr_parameters)
+    solr_parameters[:fq] ||= []
+    solr_parameters[:fq] += [
+      ActiveFedora::SolrQueryBuilder.construct_query_for_rel(has_model: Sufia::Works::GenericWork.to_class_uri)
+    ]
+  end
+
   def show_only_shared_files(solr_parameters)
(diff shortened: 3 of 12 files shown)


spec/actors/curation_concern/generic_work_actor_spec.rb
```diff
@@ -147,7 +147,6 @@
 
       it "should add to collections" do
         reload = Sufia::Works::GenericWork.find(curation_concern.id)
-        puts curation_concern.collection_ids
```
cam156: :clap: I have been meaning to put a PR in to get rid of that puts.

spec/factories/generic_files.rb
```diff
@@ -11,39 +12,5 @@
       read_groups ["public"]
     end
 
-    factory :fixture do
```
cam156: Why remove this?

spec/support/features.rb
```diff
@@ -1,6 +1,5 @@
 # spec/support/features.rb
 require File.expand_path('../features/session_helpers', __FILE__)
-require File.expand_path('../fixture_helpers', __FILE__)
```
cam156: Is there a reason to remove these.  We will likely need them in the future.

val99erie: I removed them just because they were no longer being used anywhere, but I'll add them back in if you think they are still useful.

cam156: @val99erie I'm guessing we might still want to test the different thumbnail types in the future.  Right now a work has no thumbnail, but #1073 adds them back in.

cam156:
@val99erie add #1096 for the batch delete todo.  I wonder if adding to a collection would just "work".  I do not think hydra-collections cares about it's model.

cam156:
:+1:  Waiting on travis

val99erie:
@cam156  Adding works to collections does "just work" (in the rails console), but there will need to be some view/javascript work to make it work on the dashboard.

Also, I noticed that in the old code, the "my files" page was displaying the collection names by loading the actual GenericFiles out of fedora and then calling the gf.collection method, which seems like it must be very slow.  This PCDM work might be a good opportunity to change it so that the collection IDs and names are indexed on the GenericWork, and then the "my files" page can display that info from the solr document instead of having to load objects out of fedora.  Do you think that is feasible, or are there other reasons that the objects are loaded out of fedora for the my files / dashboard views?

cam156:
@val99erie Single use links for a Work are out of scope for the moment.  No need to add a ticket for them.

val99erie:
@cam156:  Ok, I'll remove the TODO for single-use links from the code.  I'll update this PR a little later.

PR #1099, opened 2015-05-07
Pcdm 1086 GenericWork as top level class.

grosscol:
Create a GenericWork class in the primary modules directory.  Move Sufia::Works::Work and Sufia::Works::GenericWork to Concerns.  This should bring the GenericWork inline with the existing paradigm used in Sufia.

This is likely to cause a bit of an issue with rebasing for the other devs working concurrently.


cam156:
@grosscol This should be against the pcdm branch: https://github.com/projecthydra/sufia/compare/pcdm...grosscol:pcdm_1086

PR #1100, opened 2015-05-07
Move highlighted files and trophies into works including a new migrat…

tingtingjh:
The UI changes will occur upon the merging of generic_work controller/views.


app/controllers/concerns/sufia/users_controller_behavior.rb
```diff
@@ -70,18 +70,18 @@ def update_directory?
...
   def toggle_trophy
-     unless current_user.can? :edit, params[:file_id]
-       redirect_to root_path, alert: "You do not have permissions to the file"
+     unless current_user.can? :edit, params[:work_id]
+       redirect_to root_path, alert: "You do not have permissions to the work"
        return false
      end
-     # TODO  make sure current user has access to file
-     t = current_user.trophies.where(generic_file_id: params[:file_id]).first
+     # TODO  make sure current user has access to work
```
cam156: I think you can just remove the todo here.  Looks like the current user is being checked above.

tingtingjh: Will check on all the redundant TODO items when working on my next ticket.

Issue #1058, opened 2015-04-24
Users proxies partial include table rows without a surrounding table


cam156:
app/views/users/_proxies.html.erb (https://github.com/projecthydra/sufia/blob/master/app/views/users/_proxies.html.erb#L2-L6) has table rows, but no surrounding table.

We fixed this in scholarsphere by removing the table rows and using an unordered list: https://github.com/psu-stewardship/scholarsphere/commit/6749caa92c41a29bcec58fc28f2b5666934e7e89

PR #1105, opened 2015-05-07
Replaces table rows with an unordered list to make html valid, fixes …

tingtingjh:
…#1058


cam156:
:+1: waiting on Travis

PR #1113, opened 2015-05-08
fixes #1107

### LINKED ISSUE #1107: My Files in search drop down needs to be My Works
My Files in Search drop down needs to be labeled My Works.

This may be just a change to the default I18N

yinlinchen:
- Change to "My Works"
screen shot 2015-05-08 at 10 56 13 am 3

### DIFF
--- a/config/locales/sufia.en.yml
+++ b/config/locales/sufia.en.yml
@@ -19,9 +19,9 @@ en:
           all:
             label_short:  "All"
             label_long:   "All of Sufia"
-          my_files:
-            label_short:  "My Files"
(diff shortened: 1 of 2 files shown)


config/locales/sufia.en.yml
```diff
@@ -20,8 +20,8 @@ en:
             label_short:  "All"
             label_long:   "All of Sufia"
           my_files:
```
cam156: @yinlinchen It might be best to change the translation key to works as well.  That way the code will make more sense.

PR #1114, opened 2015-05-08
Adding works metadata to batch job processing.  fixes #1095

### LINKED ISSUE #1095: Batch edit after upload should modify work metadata
For the original upload stream we create a work for each file.  When the batch edit of metadata is applied after the upload completes each work should get the metadata applied along with the file.  

The batch metadata is applied to the file in this job: https://github.com/projecthydra/sufia/blob/master/sufia-models/app/jobs/batch_update_job.rb

cam156:


sufia-models/app/jobs/batch_update_job.rb
```diff
@@ -41,22 +42,17 @@ def update_file(gf, user)
...
     end
+    # update the file using the actor after setting the title
     gf.title = title[gf.id] if title[gf.id]
-    gf.attributes = file_attributes
-    gf.visibility= visibility
+    Sufia::GenericFile::Actor.new(gf, user).update_metadata(file_attributes, visibility)
+
+    # update the work to the same metadata as the file.
+    # NOTE: For the moment we are assuming copied metadata.  This is likely to change.
+    unless gf.generic_work.blank?
```
awead: Under what circumstances would the work be blank? Doesn't a GenericFile have to have a GenericWork?

cam156: @awead  In our old upload scenario there is a possibility the work gets in without a work.

Aka I upload a bunch of files, the original way.  No work is specified for them to go in. I should get a work, but I am just being careful.  You can create a file from the rails console without specifying a work at this point too.

awead: :+1:

Issue #1104, opened 2015-05-07
Display the collections that a work belongs to


val99erie:
On the "My Files" section of the dashboard display links for the collections that a GenericWork belongs to.

Here is a screenshot of how it used to look when the "My Files" page displayed GenericFiles instead of GenericWorks.  The collection links for GenericWorks should look similar to this:

![collection_links](https://cloud.githubusercontent.com/assets/298865/7524698/66728d2c-f4c9-11e4-9806-28b9e8916b6c.png)

val99erie:
One possible way to implement this would be to index the collection ID and the collection name on the GenericWork so that the link to the collection can be created using information out of the solr document (instead of having to load the GenericWork from fedora).

PR #1117, opened 2015-05-08
Links to collections from 'My Files' dashboard. Story #1104

val99erie:


app/helpers/generic_work_helper.rb
```diff
@@ -0,0 +1,11 @@
+module GenericWorkHelper
+
+  def render_collection_links(solr_doc)
+    return if solr_doc.collections.empty?
+    links = solr_doc.collections.map do |collection|
+      link_to collection.title_or_label, collections.collection_path(collection.id)
+    end
+    content_tag :span, ('Is part of: ' + links.join(', ')).html_safe
```
cam156: @val99erie  If you have a chance to change the "Is part of:" to an I18N label that would be great!

Issue #1049, opened 2015-04-23
Mark a Work as Featured Work


hectorcorrea:
Implement workflow to mark a Work as a "Featured Work" (this would be to replace the "Featured Files" feature once we implement PCDM)

cam156:
This is dependent on #1065.

cam156:
The feature option is currently on the show page of a file in the action list:

![screen shot 2015-04-29 at 10 44 35 am](https://cloud.githubusercontent.com/assets/1599081/7393618/da95e794-ee5c-11e4-9257-5de3b3977cd4.png)

gordonleacock:
Should the feature/unfeature action show on the works list under "My works" or on each work's individual page?

gordonleacock:
Or on the work's edit page?

mjgiarlo:
@gordonleacock I incline towards doing what 6.x does, and leaving it on the show page. Once it's working and 7.0.0 is out, we can chat about changing the behavior.

hectorcorrea:
@gordonleacock @mjgiarlo For now I've added the Feature/Unfeature buttons on the Work Show page. Only one of them is displayed at a time (if at all) 

![screen shot 2016-01-07 at 11 54 31 am](https://cloud.githubusercontent.com/assets/568286/12176486/9c1aa2ac-b535-11e5-94b4-0f31a54aa2a2.png)

gordonleacock:
Thanks for the heads up Hector!


----------------
Gordon Leacock (cell phone)

PR #1118, opened 2015-05-08
Featured works to use GenericWork.

grosscol:
Remove featured work from GenericFile.
Remove Trophies from GenericFile Concern.
Remove featured_work from generic file actor spec.
Remove featured from generic file specs.
Sub generic_work_id for generic_file_id.
Sub generic_work for generic_file.
Add demo GenericWork into spec setup.


sufia-models/app/models/concerns/sufia/generic_file/featured.rb
```diff
@@ -1,11 +0,0 @@
-module Sufia
```
cam156: @grosscol I'm thinking we may want to add a module to GenericWork so it knows if it is featured.  This is not required for this ticket, but will be required for #1049.  This does not block this PR from being merged.

cam156:
@grosscol Looks good.  One minor tweak if you do not mind.

cam156:
@grosscol I'm going to merge this.  The person whi picks up #1049 should just look at the code removed in this PR.

PR #1136, opened 2015-05-11
Characterize and Create Derivatives as services

### LINKED ISSUE #1127: ExtractMetadata on GenericFile should be moved into the service based archtecture
Currently FileContentDatastream is mixing in Hydra::Derivatives::ExtractMetadata.

This will allow us to do away with FileContentDatastream and allow us to treat all Files as PCDM::Files

jcoyne:
Fixes #1127


sufia-models/app/services/sufia/characterization_service.rb
```diff
@@ -0,0 +1,67 @@
+module Sufia
+  class CharacterizationService
+    include Hydra::Derivatives::ExtractMetadata
+
+    delegate :mime_type, :uri, to: :@generic_file
+    attr_reader :generic_file
+
+    def self.run(generic_file)
```
mjgiarlo: I've seen service objects use different methods as their primary interface, usually `run`, `call`, and `create`. Is it worth converging on a convention here?

spec/models/fits_datastream_spec.rb
```diff
@@ -3,9 +3,10 @@
 describe FitsDatastream, type: :model, unless: $in_travis do
   describe "image" do
     before(:all) do
+      ActiveFedora::Cleaner.clean!
```
cam156: @jcoyne Is this needed?  I thought this line would do that: https://github.com/projecthydra/sufia/blob/master/spec/spec_helper.rb#L112

sufia-models/app/services/sufia/characterization_service.rb
```diff
@@ -0,0 +1,67 @@
...
+    end
+
+    protected
+
+      def store_fulltext(extracted_text)
+        generic_file.full_text.content = extracted_text if extracted_text.present?
+      end
+
+      def extract_fulltext
+        FullTextExtractionService.run(@generic_file)
```
cam156: Should this be stored in a class variable so that it only gets run once.  I can see you are only running it once now, but in the future the method could get called multiple times.  Some one wanting to add logging ect.

sufia-models/app/services/sufia/create_derivatives_service.rb
```diff
@@ -0,0 +1,17 @@
+module Sufia
+  class CreateDerivativesService
+    include Hydra::Derivatives
+
+    def self.run(generic_file)
+      new(generic_file).create_derivatives
```
cam156: @jcoyne why new here?  I would have expected just create_derivatives

sufia-models/app/services/sufia/full_text_extraction_service.rb
```diff
@@ -0,0 +1,37 @@
+module Sufia
+  class FullTextExtractionService
+    def self.run(generic_file)
+      new(generic_file).extract
```
cam156: Similar question to above.  Why put new()?

sufia-models/app/services/sufia/full_text_extraction_service.rb
```diff
@@ -0,0 +1,37 @@
+module Sufia
+  class FullTextExtractionService
```
cam156: @jcoyne The other services modify the generic_file internally.  This one just returns the extracted text.  A comment about what the service does might be helpful in the future.

spec/models/fits_datastream_spec.rb
jcoyne: I can remove this.

sufia-models/app/services/sufia/characterization_service.rb
jcoyne: YAGNI

sufia-models/app/services/sufia/create_derivatives_service.rb
jcoyne: You need `new` to create an instance. You can't create the derivatives on a class.

sufia-models/app/services/sufia/full_text_extraction_service.rb
jcoyne: To create an instance.

sufia-models/app/services/sufia/characterization_service.rb
jcoyne: I really don't see that being a problem, do you?

sufia-models/app/services/sufia/create_derivatives_service.rb
cam156: @jcoyne I was under the impression that it is expecting an instance of generic_file as the parameter passed to the initializer. Is that an invalid impression?  If so some comments would be helpful.

sufia-models/app/services/sufia/characterization_service.rb
cam156: Ok

sufia-models/app/services/sufia/create_derivatives_service.rb
jcoyne: That's correct. But the `new` creats an instance of the CreateDerivativesService.

cam156: Now that makes sense.  I need more caffeine... Sorry!
:ok_hand:

sufia-models/app/services/sufia/characterization_service.rb
mjgiarlo: Problem would put it too strongly.

PR #1137, opened 2015-05-11
Resque jobs should be namespaced appropriately under Passenger. Fixes #1131

### LINKED ISSUE #1131: Resque jobs sometimes not queueing
Apparently this is fixed by https://github.com/ClevelandArtGIT/cma-archives/commit/177751083753bab18aa94ed2918f9463e6c6e1f6#diff-47dbc8225025d7a1c65fb7f3fba5ee7eR14.

Probably just need to adjust the generators

mjgiarlo:
This ensures that Redis keys for Resque queues receive the expected namespace when applications are run under Passenger. A number of Sufia implementations over the years have found that workers are namespaced with sufia:ENV: and queues receive the default Resque namespace of resque:, which meants that workers never pick up jobs. Resolving this issue requires adding the namespace-setting magic to the Redis config initializer and adding a new setup rake task for Resque (since jobs can be worked off via rake tasks).

Will fix new installations, and not affect existing ones.

Thanks too to @narogers, @hackmasterA, @terrellt and others over the months for poking at this and helping us arrive at a workable solution.  This is a longstanding bug that many of us have worked around for a long time!


mjgiarlo:
:clap: @cam156 :smiley_cat:

PR #1144, opened 2015-05-12
Extract a VersioningService. Fixes #1130

### LINKED ISSUE #1130: When saving uploaded files, the service should record a version in Fedora
Currently this is done by https://github.com/projecthydra/sufia/blob/master/sufia-models/lib/sufia/models/file_content/versions.rb, but we want to move that functionality to the service based architeture.

This will allow use to get rid of the FileContentDatastream and treat all attached files the same.

Related to #1127

jcoyne:


mjgiarlo:
Look at all those gorgeous deletions!

spec/controllers/generic_files_controller_spec.rb
```diff
@@ -478,12 +472,12 @@
...
           let(:versions)         { restored_content.versions }
-          let(:latest_version)   { restored_content.latest_version }
+          let(:latest_version)   { Sufia::VersioningService.latest_version_of(restored_content) }
 
-          it "should restore the first versions's content and metadata" do
+          it "restores the first versions's content and metadata" do
             expect(restored_content.mime_type).to eq file1_type
             expect(restored_content.original_name).to eq file1
-            expect(versions.all.count).to eq 3
+            expect(versions.all.count).to eq 4
```
cam156: @jcoyne I wonder why we now get 4 versions instead of 3?

spec/jobs/audit_job_spec.rb
```diff
@@ -12,24 +12,23 @@
...
     end
-  end
 
-  describe "audit on an invalid version of the content" do
-    let(:uri) { file.content.versions.first.uri + "bogus" }
-    it "should fail" do
-      expect(job.run).to eq(false)
+    context "on an invalid version of the content" do
+      before { Sufia::VersioningService.create(file.content) }
+      let(:uri) { Sufia::VersioningService.latest_version_of(file.content).uri + 'bogus' }
```
cam156: Is there any reason to actually run the version service in this test?  Couldn't we just pas any string as the URI?

sufia-models/app/services/sufia/versioning_service.rb
```diff
@@ -0,0 +1,27 @@
...
+    # @param [ActiveFedora::File] content
+    # @param [User] user
+    def self.create(content, user=nil)
+      content.create_version
+      record_committer(content, user) if user
+    end
+
+    # @param [ActiveFedora::File] file
+    def self.latest_version_of(file)
+      file.versions.last
```
cam156: @jcoyne I am seeing two ways of getting at the version file.content.versions.first and file.versions. Should these all be file.versions now, since content is implied, or is there a version for the file and a version for the content?

spec/controllers/generic_files_controller_spec.rb
jcoyne: I don't know. I do know the versioning service is only called three times.

sufia-models/app/services/sufia/versioning_service.rb
jcoyne: file.content.versions only works when file is a GenericFile. file.versions is when file is an ActiveFedora::File

spec/jobs/audit_job_spec.rb
jcoyne: I'm just trying to keep similar to the spirit of the test that was already there.

PR #1148, opened 2015-05-13
Add a collection size service that queries solr for collect size info…

### LINKED ISSUE #1120: Collection Size
Collection Size must include the size of a work.

At this line: https://github.com/projecthydra/sufia/blob/master/sufia-models/app/models/concerns/sufia/collection_behavior.rb#L21

Works do not have content to determine size from bytes.  Need to add in a work size, and allow the collection to get the size.

cam156:
…rmation. fixes #1120

### DIFF
--- a/app/search_builders/collection_search_builder.rb
+++ b/app/search_builders/collection_search_builder.rb
@@ -1,7 +1,9 @@
 class CollectionSearchBuilder < Blacklight::Solr::SearchBuilder
   include Hydra::Collections::SearchBehaviors
 
-  attr_reader :item
+  def item
+    scope.item
+  end
 
   # include filters into the query to only include the collection memebers
   def include_item_ids(solr_parameters)
@@ -9,9 +11,9 @@ class CollectionSearchBuilder < Blacklight::Solr::SearchBuilder
     solr_parameters[:fq] << "hasCollectionMember_ssim:#{item.id}"
   end
 
-  def initialize(item, processor_chain, scope)
-    super(processor_chain, scope)
-    @item = item
+  def include_contained_files(solr_parameters)
+    solr_parameters[:fq] ||= []
+    solr_parameters[:fq] << "{!join from=generic_files_tesim to=id}{!join from=hasCollectionMember_ssim to=id}id:#{collection.id}"
   end
 
 end
--- a/app/services/sufia/collection_member_service.rb
+++ b/app/services/sufia/collection_member_service.rb
@@ -23,7 +23,7 @@ module Sufia
     end
 
     def collection_search_builder
-      @collection_search_builder ||= CollectionSearchBuilder.new(item, [:include_item_ids, :add_paging_to_solr], self)
+      @collection_search_builder ||= CollectionSearchBuilder.new([:include_item_ids, :add_paging_to_solr], self)
     end
 
   end
--- a/app/services/sufia/collection_size_service.rb
+++ b/app/services/sufia/collection_size_service.rb
@@ -1,0 +1,37 @@
+# returns a list of solr documents for collections the item is a part of
+module Sufia
+  class CollectionSizeService
+    include Blacklight::Configurable
+    include Blacklight::SearchHelper
+
+    attr_reader :collection
+
+    self.copy_blacklight_config_from(CatalogController)
+
+    def self.run(collection)
+      new(collection).collection_size
+    end
+
+    def initialize(collection)
+      @collection = collection
+    end
+
+    def collection_size
+      query = collection_search_builder.rows(max_collection_size).merge({ fl:[size_field] })
+      resp = repository.search(query)
+      resp.documents.reduce(0) { |total, doc| total += doc[size_field].blank? ? 0 : doc[size_field][0].to_f }
+    end
+
+    def collection_search_builder
+      @collection_search_builder ||= CollectionSearchBuilder.new([:include_contained_files, :add_paging_to_solr], self)
+    end
+
+    def size_field
+      Solrizer.solr_name(:file_size, :symbol)
+    end
+
+    def max_collection_size
+      1000
+    end
+  end
+end
--- a/spec/presenters/sufia/collection_presenter_spec.rb
+++ b/spec/presenters/sufia/collection_presenter_spec.rb
@@ -8,7 +8,7 @@ describe Sufia::CollectionPresenter do
                             :language, :identifier, :based_near, :related_url] }
   end
 
-  let(:collection) { Collection.new(description: 'a nice collection', title: 'A clever title') }
+  let(:collection) { Collection.new(id: '111', description: 'a nice collection', title: 'A clever title') }
   let(:presenter) { described_class.new(collection) }
 
   describe "#terms_with_values" do
--- a/spec/services/collection_size_service_spec.rb
+++ b/spec/services/collection_size_service_spec.rb
@@ -1,0 +1,78 @@
+require 'spec_helper'
+
+describe Sufia::CollectionSizeService do
+
+
+  let(:coll1_attrs) {{ id: 'col1', title_tesim: ['A Collection 1'], hasCollectionMember_ssim: works }}
+
+  let(:coll1) { SolrDocument.new(coll1_attrs) }
+
+  before do
+    ActiveFedora::SolrService.add(coll1_attrs)
+    ActiveFedora::SolrService.commit
+  end
+
+  describe "#run" do
+    subject { Sufia::CollectionSizeService.run(coll1) }
+
+    context "a collection with works" do
+      let(:file1_attrs) {{ id: 'file_1111', title_tesim: ['A first generic file'], file_size_ssim: [50] }}
+      let(:file2_attrs) {{ id: 'file_2222', title_tesim: ['A second generic file'], file_size_ssim: [55] }}
+
+      let(:work1_attrs) {{ id: 'work_1111', title_tesim: ['A first generic work'], generic_files_tesim: files1 }}
+      let(:work2_attrs) {{ id: 'work_2222', title_tesim: ['A second generic work'], generic_files_tesim: files2 }}
+
+      let (:work1) { SolrDocument.new(work1_attrs) }
+      let (:work2) { SolrDocument.new(work2_attrs) }
+      let (:file1) { SolrDocument.new(file1_attrs) }
+      let (:file2) { SolrDocument.new(file2_attrs) }
+
+      let(:works) {[work1.id, work2.id]}
+
+      before do
+        ActiveFedora::SolrService.add(file1_attrs)
+        ActiveFedora::SolrService.add(work1_attrs)
+        ActiveFedora::SolrService.add(file2_attrs)
+        ActiveFedora::SolrService.add(work2_attrs)
+        ActiveFedora::SolrService.commit
+      end
+
+      context "that have files" do
+        let(:files1) {[file1.id]}
+        let(:files2) {[file2.id]}
+
+        specify "returns the correct size" do
+          expect(subject).to eq(105.0)
+        end
+      end
+
+      context "that do not have files" do
+        let(:files1) {[]}
+        let(:files2) {[]}
+
+        specify "returns zero size" do
+          expect(subject).to eq(0.0)
+        end
+      end
+
+      context "that some have files" do
+        let(:files1) {[file1.id]}
+        let(:files2) {[]}
+
+        specify "returns the correct size" do
+          expect(subject).to eq(50.0)
+        end
+      end
+    end
+
+    context "a collection without works" do
+      let(:works) {[]}
+
+      specify "returns zero size" do
+        expect(subject).to eq(0.0)
+
+      end
+    end
+  end
+
+end
--- a/spec/views/collections/_show_descriptions.html.erb_spec.rb
+++ b/spec/views/collections/_show_descriptions.html.erb_spec.rb
@@ -2,17 +2,19 @@ require 'spec_helper'
 
 describe 'collections/_show_descriptions.html.erb', :type => :view do
   context 'displaying a custom collection' do
+    let(:collection_size) { 123456678 }
     let(:collection) do
       mock_model(Collection,
                   resource_type: [], creator: [], contributor: [], tag: [],
                   description: '', title: 'hmm',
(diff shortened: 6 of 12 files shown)


app/search_builders/collection_search_builder.rb
```diff
@@ -9,9 +11,9 @@ def include_item_ids(solr_parameters)
     solr_parameters[:fq] << "hasCollectionMember_ssim:#{item.id}"
   end
 
-  def initialize(item, processor_chain, scope)
-    super(processor_chain, scope)
-    @item = item
+  def include_contained_files(solr_parameters)
+    solr_parameters[:fq] ||= []
+    solr_parameters[:fq] << "{!join from=generic_files_tesim to=id}{!join from=hasCollectionMember_ssim to=id}id:#{collection.id}"
```
mjgiarlo: I'm not that well versed on Solr joins. Could someone like @atz or @jcoyne or @cbeer comment on this line?

app/search_builders/collection_search_builder.rb
```diff
@@ -9,9 +11,9 @@ def include_item_ids(solr_parameters)
     solr_parameters[:fq] << "hasCollectionMember_ssim:#{item.id}"
   end
 
-  def initialize(item, processor_chain, scope)
```
mjgiarlo: It's not clear to me why `#initialize` is going away in favor of `#include_contained_files`. Could you say a bit about this change?

app/services/sufia/collection_member_service.rb
```diff
@@ -23,7 +23,7 @@ def list_collections()
     end
 
     def collection_search_builder
-      @collection_search_builder ||= CollectionSearchBuilder.new(item, [:include_item_ids, :add_paging_to_solr], self)
+      @collection_search_builder ||= CollectionSearchBuilder.new([:include_item_ids, :add_paging_to_solr], self)
```
mjgiarlo: Related to the above: can you say a bit about this change?

app/services/sufia/collection_size_service.rb
```diff
@@ -0,0 +1,34 @@
...
+
+    def self.run(collection)
+      new(collection).collection_size
+    end
+
+    def initialize(collection)
+      @collection = collection
+    end
+
+    def collection_size()
```
mjgiarlo: Remove parens.

app/services/sufia/collection_size_service.rb
```diff
@@ -0,0 +1,34 @@
...
+      query = collection_search_builder.rows(1000).merge({ fl:[size_field]})
```
mjgiarlo: What does the hard-coded `1000` represent?  Should this be a constant with a descriptive name, so folks know what it is and can override it?

mjgiarlo: Also, trivial spacing issue:

...`.merge({ f1: [size_field] })`

app/services/sufia/collection_size_service.rb
```diff
@@ -0,0 +1,34 @@
...
+      resp = repository.search(query)
+      resp.documents.reduce(0) { |total, doc| total += doc[size_field].blank? ? 0: doc[size_field][0].to_f }
```
mjgiarlo: Add a space between the zero and the colon.

app/search_builders/collection_search_builder.rb
cam156: @mjgiarlo I used the same mechanics as Hydra::Collections::SearchBehaviors instead of passing in an item to the initializer.  The only reason to override the initializer was to allow for item to be passed in.

see: https://github.com/projecthydra/hydra-collections/blob/master/lib/hydra/collections/search_behaviors.rb#L3-L5

When I added the additional method that did not use item it did not make sense to me to force a nil to be passed in.

app/services/sufia/collection_size_service.rb
```diff
@@ -0,0 +1,34 @@
...
+    end
+
+    def collection_search_builder
+      @collection_search_builder ||= CollectionSearchBuilder.new([:include_contained_files, :add_paging_to_solr], self)
+    end
+
+
```
mjgiarlo: Remove this blank line.

spec/presenters/sufia/collection_presenter_spec.rb
```diff
@@ -8,7 +8,7 @@
                             :language, :identifier, :based_near, :related_url] }
   end
 
-  let(:collection) { Collection.new(description: 'a nice collection', title: 'A clever title') }
+  let(:collection) { Collection.new(id: '111', description: 'a nice collection', title: 'A clever title') }
```
mjgiarlo: Can you say why you need to add an ID for these tests to pass?

spec/services/collection_size_service_spec.rb
```diff
@@ -0,0 +1,78 @@
+require 'spec_helper'
+
+describe Sufia::CollectionSizeService do
+
+
+  let(:coll1_attrs) {{ id: 'col1', title_tesim: ['A Collection 1'], hasCollectionMember_ssim: works }}
```
mjgiarlo: Sorry if I missed this, but where is "works" defined at the end of this line?

spec/services/collection_size_service_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+
+  let (:coll1) { SolrDocument.new(coll1_attrs) }
```
mjgiarlo: Remove space between `let` and the opening paren.

spec/services/collection_size_service_spec.rb
mjgiarlo: nm, found it.

app/services/sufia/collection_size_service.rb
mjgiarlo: Just curious why you chose floats vs. fixnums for size? Is the size returning # bytes or what?

spec/views/collections/_show_descriptions.html.erb_spec.rb
```diff
@@ -13,6 +13,7 @@
     end
     before do
       assign(:presenter, presenter)
+      allow(Sufia::CollectionSizeService).to receive(:run).and_return(123456678)
```
mjgiarlo: Is this just here so that things don't blow up?  If so, maybe just end at `receive(:run)`?  Having the return value there suggests it's relevant for the tests but I'm guessing it's a garbage value.

sufia-models/app/models/concerns/sufia/works/curation_concern/with_generic_files.rb
```diff
@@ -5,7 +5,7 @@ module WithGenericFiles
 
       included do
         # This used to have a hasFile relation when in hydra-works.  That does not seem to exist so I am using hasPart instead
-        has_many :generic_files, predicate: ActiveFedora::RDF::Fcrepo::RelsExt.isPartOf, class_name: "GenericFile"
+        has_many :generic_files, predicate: ActiveFedora::RDF::Fcrepo::RelsExt.hasPart, class_name: "GenericFile"
```
mjgiarlo: I see the comment on line 7 but i'm not clear on why Sufia should be doing this if Hydra::Works is.  Is it not?  Is this one of those areas where we have temporary code that will go away when we're more fully based on HW?  `hasPart` is clearly preferable to `isPartOf` here, but I would think HW already defines this relationship for us.

spec/services/collection_size_service_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+  before do
+    ActiveFedora::SolrService.add(coll1_attrs)
+    ActiveFedora::SolrService.commit
+  end
+
+  describe "#run" do
+    subject { Sufia::CollectionSizeService.run(coll1) }
+
+    context "a collection with works" do
+      let(:file1_attrs) {{ id: 'file_1111', title_tesim: ['A first generic file'], file_size_ssim: [50] }}
```
mjgiarlo: Why is file_size an array?

sufia-models/app/services/sufia/generic_file_indexing_service.rb
```diff
@@ -5,6 +5,7 @@ def generate_solr_document
         solr_doc[Solrizer.solr_name('label')] = object.label
         solr_doc[Solrizer.solr_name('file_format')] = object.file_format
         solr_doc[Solrizer.solr_name('file_format', :facetable)] = object.file_format
+        solr_doc[Solrizer.solr_name(:file_size, :symbol)] = object.file_size[0]
```
mjgiarlo: Why is file_size returning an enumeration?

app/search_builders/collection_search_builder.rb
mjgiarlo: :ok_hand:

sufia-models/app/services/sufia/generic_file_indexing_service.rb
cam156: @mjgiarlo I do not know.  The file_size field was defined previously to this commit.  I am just using the existing field.

spec/services/collection_size_service_spec.rb
cam156: Not sure.  I has been one, I did not change the field in this PR.

sufia-models/app/services/sufia/generic_file_indexing_service.rb
mjgiarlo: OK, orthogonal to this PR, then.

sufia-models/app/models/concerns/sufia/works/curation_concern/with_generic_files.rb
cam156: We have not yet integrated HydraWorks into sufia.  This line will be removed once that happens, but for now the stop gap was invalid.

spec/services/collection_size_service_spec.rb
mjgiarlo: :ok: best not to change that in this PR then.

sufia-models/app/models/concerns/sufia/works/curation_concern/with_generic_files.rb
mjgiarlo: Ahh, very good.

spec/views/collections/_show_descriptions.html.erb_spec.rb
cam156: The test was checking to be certain the value was returned and show, so it is not just to keep it from blowing up but also to give the correct information to the view.

spec/services/collection_size_service_spec.rb
cam156: :ok_hand:

spec/views/collections/_show_descriptions.html.erb_spec.rb
mjgiarlo: Ah, I now see the value up above in "bytes".  Perhaps put this number into a let statement and use that variable (e.g., "bytes") in both places?

`let(:bytes) { 123456678 }`

spec/presenters/sufia/collection_presenter_spec.rb
cam156: The solr query now uses the ID to find the size.  The solr query fails without the id

app/services/sufia/collection_size_service.rb
cam156: :ok_hand:

app/services/sufia/collection_size_service.rb
cam156: I did not choose it, it was what solr returned, and how the file size was defined currently.

app/services/sufia/collection_size_service.rb
cam156: @mjgiarlo good idea.  I will create a method.

spec/presenters/sufia/collection_presenter_spec.rb
mjgiarlo: And in practice, in our code we shouldn't have any Collections without IDs, so no problem?

app/services/sufia/collection_size_service.rb
mjgiarlo: Okie dokie. For another PR at another time,then. :)

app/services/sufia/collection_size_service.rb
mjgiarlo: :+1:

spec/presenters/sufia/collection_presenter_spec.rb
cam156: Yes, that is the case.  id is a required field.  Even if it is not set it is passed to solr as "DONOTUSE" or something like that.

mjgiarlo: :cool:

spec/views/collections/_show_descriptions.html.erb_spec.rb
cam156: :ok_hand: I can do that.

mjgiarlo: Also, don't ever let our users see this number or they'll get funny ideas about how much they can upload. :wink:

app/search_builders/collection_search_builder.rb
atz: `_tesim` would be the wrong kind of field to store an ID in.  We don't want english text stemming on IDs.  (This kind of imprecision is common in our legacy hydra apps, too.)  As for whether the joins do exactly what you want them to in this context, I don't know enough about what is going into those fields to say.

atz: In general, Solr joins are not entirely like SQL joins.  They have stronger limitations: 
- only one "table" (solr core), so more like a self-join
- no fields returned out of the "from" docs

But JOINs are typically a place Solr performance goes to die.  Given this community's tendency to treat Solr as a far-away thing to be abstracted, encapsulated and otherwise misunderstood, I would regard the introduction of joins with skepticism.  Basically, it ain't a RDBS.  

Clues from Lucid: https://lucidworks.com/blog/solr-and-joins/

If I understand correctly that the current modeling has object-documents and collection-documents (read "document" here as "Solr document") in the same core, i.e. that we are joining on UNIQUE_ID, then we should expect worst case performance similar to the post's example.  My interpretation of that is that it would be acceptable on the scale of say, collections (~100's of IDs), but unacceptable on the scale of every-object-in-repo (~1M).

PR #1152, opened 2015-05-14
Remove single use link and download work link from work action list i…

tingtingjh:
fixes-1132_before
fixes-1132_after


cam156:
:+1: Just waiting on travis

PR #1155, opened 2015-05-14
Implement Work Creation Page. Fixes #1067

### LINKED ISSUE #1067: Implement Work Creation Page
As described by the wireframe: https://wiki.duraspace.org/pages/viewpage.action?pageId=68070083#mockup_action_create_work

We may want to skip the upload into a collection for the initial implementation.  A collection is not required for a work to be uploaded.

dependent on #998

hortongn:
blacklight

Since a work only has 1 title, this issue required me to change the generic work model's `title` from an array to a scalar value.  That broke many tests, but all have been fixed except for 3 tests in spec/jobs/batch_update_job_spec.rb.  It looks like batch update is broken now that generic works and generic files have different types of values for the `title` field.  It looks like a simple fix, but should probably be it's own ticket.  

As suggested in this ticket's description, this implements the creation of a work without attaching files.  The next step is to allow the user to upload files during the work's creation.

### DIFF
--- a/app/views/curation_concern/base/_form.html.erb
+++ b/app/views/curation_concern/base/_form.html.erb
@@ -1,17 +1,17 @@
-<%= simple_form_for [:curation_concern, curation_concern], html: { multipart: true } do |f| %>
+<%= simple_form_for [sufia, @form], html: { multipart: true } do |f| %>
   <div id="works_edit">
     <h2 class="non lower">Describe Your Work</h2>
     <p>The more descriptive information you provide the better we can serve your needs.</p>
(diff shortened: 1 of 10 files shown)


cam156:
@hortongn mark those job specs as pending and we can create a ticket to fix them

hortongn:
@cam156 will do

hortongn:
@cam156 I've marked the failing spec/jobs/batch_update_job_spec.rb tests as pending.

cam156:
@hortongn looks like sopmeone added a test that will need to be updated.  I will take care of rebasing this PR tomorrow morning.  Thanks for all your hard work!

PR #1156, opened 2015-05-14
When clicking on a work in Collection Show page, it should redirect t…

tingtingjh:
fixes-1129_worktitle
fixes-1129_thumbnail

### DIFF
--- a/spec/features/show_work_in_collection_spec.rb
+++ b/spec/features/show_work_in_collection_spec.rb
@@ -1,0 +1,52 @@
+require 'spec_helper'
+
+describe "In Show Collection", :type => :feature do
+  let(:user) { create :jill }
+
+  let!(:work) {
+    GenericWork.new.tap do |work|
+      work.title = ['work title abc']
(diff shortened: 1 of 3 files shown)


spec/features/show_work_in_collection_spec.rb
```diff
@@ -0,0 +1,52 @@
+require 'spec_helper'
+
+describe "In Show Collection", :type => :feature do
+  let(:user) { create :jill }
+
+  let!(:work) {
```
cam156: @tingtingjh you can create a public work using the Factory: let(work) { FactoryGirl.create(:public_generic_work) }

cam156:
@tingtingjh the comment above was for informational purposes only.  :+1: waiting on travis

spec/features/show_work_in_collection_spec.rb
tingtingjh: Will make sure to write specs later on in this way.

PR #1166, opened 2015-05-20
Include the correct Metadata namespace. 

hectorcorrea:
The previous code referenced a non-existing namespace (but Rails masked the issue and loaded the correct one)


awead:
:+1:

jcoyne:
WithBasicMetadata is including the same stuff as Metadata, so one or the other should be removed.

See: 
https://github.com/projecthydra/sufia/blob/pcdm/sufia-models/app/models/concerns/sufia/works/curation_concern/with_basic_metadata.rb
https://github.com/projecthydra/sufia/blob/pcdm/sufia-models/app/models/concerns/sufia/works/generic_work_rdf_properties.rb

PR #1167, opened 2015-05-20
Removes extra includes from works.

hectorcorrea:
This addresses https://github.com/projecthydra/sufia/pull/1166#issuecomment-103974822 that @jcoyne made


jcoyne:
:+1:

cam156:
@hectorcorrea & @jcoyne This was discussed during the sprint.  The additional layers were left in from worthwhile so that when @flyingzumwalt refactors the code it will be easier.  Not sure we really want to remove them for that reason.

jcoyne:
If we don't merge this PR, then we should make another PR to revert https://github.com/projecthydra/sufia/commit/6df5a195f094e1a36a37c6e62994029d20a4174f. There's two modules doing largely the same thing (defining metadata).

PR #1168, opened 2015-05-20
Use a version of Hydra::Collections that uses PCDM

jcoyne:


spec/actors/curation_concern/generic_work_actor_spec.rb
```diff
@@ -170,6 +166,7 @@
         reload = GenericWork.find(curation_concern.id)
         expect(reload.identifier).to be_blank
         expect(reload).to be_persisted
+        pending "aggregation delete is not working so it's not removing from old collections. Possibly fixed by https://github.com/projecthydra-labs/activefedora-aggregation/pull/33"
```
cam156: Can you create a ticket for this pending so it does not get lost?

spec/views/my/_list_works.html.erb_spec.rb
```diff
@@ -4,15 +4,17 @@
 
   let(:work_title) { 'Work Title' }
   let(:coll_title) { 'Collection Title' }
-  let(:collection) { FactoryGirl.create(:collection, title: coll_title) }
-  let!(:work) { FactoryGirl.create(:work, title: [work_title], collections: [collection]) }
+  let(:collection) { FactoryGirl.build(:collection, id: '3197z497t', title: coll_title) }
+  let!(:work) { FactoryGirl.build(:work, id: '3197z511f', title: [work_title]) }
```
cam156: @jcoyne How does the work get assigned to the collection here?  what am I missing?

sufia-models/app/actors/curation_concern/generic_work_actor.rb
```diff
@@ -46,7 +46,11 @@ def add_to_collections(new_collection_ids)
       end
 
       #add to new
-      curation_concern.collection_ids = new_collection_ids
+      new_collection_ids.each do |coll_id|
+        collection = Collection.find(coll_id)
+        collection.members << curation_concern
+        collection.save
```
cam156: @jcoyne Do we really need to save after each assign?  Couldn't this be moved outside the loop?

spec/actors/curation_concern/generic_work_actor_spec.rb
jcoyne: https://github.com/projecthydra/sufia/issues/1170

spec/views/my/_list_works.html.erb_spec.rb
jcoyne: It doesn't in this test. This is a view test, so I just stubbed out the `CollectionMemberService` to return the correct stuff.

sufia-models/app/actors/curation_concern/generic_work_actor.rb
jcoyne: Yes, we need to save each collection. There's no way to move it outside the loop.

PR #1169, opened 2015-05-21
Adding filter for catalog to find works that contain files that match…

### LINKED ISSUE #1151: Allow a GenericWork to be found by the GenericFile metadata and content
Allow a GenericWork to be found by the contents and metadata of the files contained by the work.

I believe this entails indexing the GenericFile metadata at the GenericWork level.

cam156:
… the query. fixes #1151 

In addition I refactored the catalog controller test to use only solr instead of fedora to speed up the tests by half even though I added 3 additional tests.

### DIFF
--- a/app/controllers/concerns/sufia/catalog.rb
+++ b/app/controllers/concerns/sufia/catalog.rb
@@ -2,7 +2,7 @@ module Sufia
   module Catalog
     extend ActiveSupport::Concern
     included do
-      self.search_params_logic += [:only_works_and_collections]
+      self.search_params_logic += [:only_works_and_collections, :show_works_or_works_that_contain_files]
     end
   end
 end
--- a/app/search_builders/sufia/search_builder.rb
+++ b/app/search_builders/sufia/search_builder.rb
@@ -54,4 +54,32 @@ module Sufia::SearchBuilder
     solr_parameters[:fq] << "#{Solrizer.solr_name("has_model", :symbol)}:(\"GenericWork\" \"Collection\")"
   end
 
+  # show both works that match the query and works that contain files that match the query
+  def show_works_or_works_that_contain_files(solr_parameters)
+    return if solr_parameters[:q].blank?
+    solr_parameters[:user_query] = solr_parameters[:q]
+    solr_parameters[:q] = new_query
+  end
+
+  protected
+
+  # the {!lucene} gives us the OR syntax
+  def new_query
+    "{!lucene}#{ interal_query(dismax_query) } #{ interal_query(join_for_works_from_files) }"
+  end
+
+  # the _query_ allows for another parser (aka dismax)
+  def interal_query( query_value)
+     "_query_:\"#{query_value}\""
+  end
+
+  # the {!dismax} causes the query to go against the query fields
+  def dismax_query
+    "{!dismax v=$user_query}"
+  end
+
+  # join from file id to work relationship solrized generic_work_id
+  def join_for_works_from_files
+    "{!join from=#{Solrizer.solr_name('generic_work_id', :symbol)} to=id}#{dismax_query}"
+  end
 end
--- a/lib/generators/sufia/templates/catalog_controller.rb
+++ b/lib/generators/sufia/templates/catalog_controller.rb
@@ -15,7 +15,7 @@ class CatalogController < ApplicationController
   # These before_filters apply the hydra access controls
   before_filter :enforce_show_permissions, only: :show
   # This applies appropriate access controls to all solr queries
-  CatalogController.search_params_logic += [:add_access_controls_to_solr_params, :add_advanced_parse_q_to_solr]
+  CatalogController.search_params_logic = [:add_advanced_parse_q_to_solr] + search_params_logic + [:add_access_controls_to_solr_params]
 
   skip_before_filter :default_html_head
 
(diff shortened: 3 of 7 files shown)


app/search_builders/sufia/search_builder.rb
```diff
@@ -54,4 +54,30 @@ def only_works_and_collections(solr_parameters)
...
 
+  # show both works that match the query and works that contain files that match the query
+  def show_works_or_works_that_conatin_files(solr_parameters)
+    return if scope.params[:q].blank?
+    solr_parameters[:q] = new_query
+    solr_parameters[:user_query] = scope.params[:q]
+  end
+
+  # the {!lucene} gives us the OR syntax
+  def new_query
```
jcoyne: Should this be a protected method?

app/search_builders/sufia/search_builder.rb
```diff
@@ -54,4 +54,30 @@ def only_works_and_collections(solr_parameters)
...
+    "{!lucene}#{ interal_query(dismax_query) } #{ interal_query(join_for_works_from_files) }"
+  end
+
+  # the _query_ allows for another parser (aka dismax)
+  def interal_query( query_value)
```
jcoyne: protected method?

app/search_builders/sufia/search_builder.rb
```diff
@@ -54,4 +54,30 @@ def only_works_and_collections(solr_parameters)
...
+     "_query_:\"#{query_value}\""
+  end
+
+  # the {!dismax} causes the query to go against the query fields
+  def dismax_query
```
jcoyne: protected?

app/search_builders/sufia/search_builder.rb
```diff
@@ -54,4 +54,30 @@ def only_works_and_collections(solr_parameters)
...
+    "{!dismax v=$user_query}"
+  end
+
+  # join from file id to work relationship solrized generic_work_id
+  def join_for_works_from_files
```
jcoyne: protected?

app/search_builders/sufia/search_builder.rb
```diff
@@ -54,4 +54,30 @@ def only_works_and_collections(solr_parameters)
     solr_parameters[:fq] << "#{Solrizer.solr_name("has_model", :symbol)}:(\"GenericWork\" \"Collection\")"
   end
 
+  # show both works that match the query and works that contain files that match the query
+  def show_works_or_works_that_conatin_files(solr_parameters)
```
jcoyne: typo `contain`

app/controllers/concerns/sufia/catalog.rb
```diff
@@ -2,7 +2,7 @@ module Sufia
   module Catalog
     extend ActiveSupport::Concern
     included do
-      self.search_params_logic += [:only_works_and_collections]
+      self.search_params_logic += [:only_works_and_collections, :show_works_or_works_that_conatin_files]
```
jcoyne: typo `contain`

lib/generators/sufia/templates/catalog_controller.rb
```diff
@@ -15,7 +15,7 @@ class CatalogController < ApplicationController
   # These before_filters apply the hydra access controls
   before_filter :enforce_show_permissions, only: :show
   # This applies appropriate access controls to all solr queries
-  CatalogController.search_params_logic += [:add_access_controls_to_solr_params, :add_advanced_parse_q_to_solr]
+  CatalogController.search_params_logic = [:add_advanced_parse_q_to_solr] + search_params_logic + [:add_access_controls_to_solr_params]
```
jcoyne: What's the reason for putting the `add_access_controls_to_solr_params` at the end and the `add_advanced_parse_q_to_solr` at the beginning?

spec/controllers/catalog_controller_spec.rb
```diff
@@ -12,34 +12,36 @@
   describe "#index" do
 
     let!(:rocks) {
-      GenericWork.new.tap do |work|
+      GenericWork.new(id: 'rock123').tap do |work|
```
jcoyne: why is this necessary?

jcoyne: Ah, I see, NM.

jcoyne:
Can you please add a test for search_builder.rb so we can see that the correct queries are being created given different input?

lib/generators/sufia/templates/catalog_controller.rb
cam156: @jcoyne add_advanced_parse_q_to_solr changes the query and my :show_works_or_works_that_conatin_files also changes the query.  Add add_advanced_parse_q_to_solr just overrides any changes I make, while my change keeps the original string.

cam156:
@jcoyne the reason I went away from search_builder.rb test is you are really just looking at strings.  The real test is if they do what you want them to do in the controller.  I'm not sure I see the value to adding those test.

jcoyne:
@cam156 if these tests in the controller start failing, I'd want to know "Are we sending the correct query to solr?" But presently we don't have tests that can help us answer that question.

cam156:
@jcoyne Ok, but I'm just really worried about test bloat.  Our test is extremely long as it is...

jcoyne:
@cam156 Fortunately testing the SearchBuilder should be really fast since there are no external API calls.

app/search_builders/sufia/search_builder.rb
cam156: added protected above all  the methods you mentioned

cam156:
@jcoyne added tests for the new search builder method and move the internal methods to protected.

jcoyne:
:+1:  Looks good to me.

PR #1172, opened 2015-05-26
Update work action menu with transfer link.

grosscol:
Update new transfers view for GenericWork.
Update locales strings for transfers of works
Move routes for transfers from GenericFile to GenericWork
Update proxy deposit routes spec for GenericWorks.
Add ownership transfer feature specs for Works.
Refactor Transfer Work Action link_to as block.

### DIFF
--- a/app/views/my/_work_action_menu.html.erb
+++ b/app/views/my/_work_action_menu.html.erb
@@ -28,13 +28,11 @@
         <span> TODO: Highlight Work on Profile </span>
       <% end %>
     </li>
-
     <li role="menuitem" tabindex="-1">
-      <%= link_to('#') do %>
-        <i class="glyphicon glyphicon-transfer" aria-hidden="true"></i>
-        <span> TODO: Transfer Ownership of Work </span>
+      <%= link_to( sufia.new_generic_work_transfer_path(document.id), class: 'itemicon itemtransfer', title: 'Transfer Ownership of Work') do %>
+        <i aria-hidden="true" class='glyphicon glyphicon-transfer'></i>
+        <span> Transfer Ownership of Work </span>
       <% end %>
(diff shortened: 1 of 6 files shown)


app/views/my/_work_action_menu.html.erb
```diff
@@ -28,13 +28,13 @@
         <span> TODO: Highlight Work on Profile </span>
       <% end %>
     </li>
-
     <li role="menuitem" tabindex="-1">
-      <%= link_to('#') do %>
-        <i class="glyphicon glyphicon-transfer" aria-hidden="true"></i>
-        <span> TODO: Transfer Ownership of Work </span>
+      <% gw = document.to_model %>
```
cam156: @grosscol Do we need the to_model here?  That requires the work being loaded from fedora.

spec/features/ownership_transfer_spec.rb
```diff
@@ -32,32 +32,66 @@
...
       context 'If the new owner accepts it' do
-      pending 'The transfer option is not available' do
-        fail
-      end
+        before do
+          new_owner.proxy_deposit_requests.last.transfer!
+          user_utility_toggle.click
+          click_link 'transfer requests'
+        end
+        specify 'I should see it was accepted' do
```
cam156: @grosscol is there a reason you are using specify here and it above?

grosscol: Probably the flavor that came along with the copy pasta.  Correcting specify calls to it calls.

app/views/my/_work_action_menu.html.erb
grosscol: @cam156 It does not appear that the model is required.  Removing.

PR #1173, opened 2015-05-28
Adding Jasmine for testing java script and a few small tests

cam156:
I intentionally made this two commits to show what code was created by jasmine and what code I added after the fact.

Jasmine is a testing suite for javascript (https://github.com/jasmine/jasmine-gem).  I have integrated it in with the spec tests within sufia in this PR, so no additional steps are needed by the developer or Travis to run the jasmine tests (https://github.com/psu-stewardship/sufia/blob/jasmine/spec/javascripts/jasmine_spec.rb).

Here's an example of a spec: https://github.com/psu-stewardship/sufia/blob/jasmine/spec/javascripts/notify_update_link_spec.js

Any additional *spec.js files in the spec/javascript directory will be picked up by jasmine and run in the future.

When rspec is run the jasmine rake task is started (rake jasmine:ci).  If there are no failures the test passes.  If there are failures the output from jasmine is reported and the test fails.

### DIFF
--- a/spec/javascripts/helpers/jasmine-jquery.js
+++ b/spec/javascripts/helpers/jasmine-jquery.js
@@ -1,0 +1,832 @@
+/*!
+Jasmine-jQuery: a set of jQuery helpers for Jasmine tests.
+
+Version 2.0.7
+
+https://github.com/velesin/jasmine-jquery
+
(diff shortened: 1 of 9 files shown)


jcoyne:
:+1:

PR #1175, opened 2015-06-01
Moving where the edit permissions are so that the my controller will …

### LINKED ISSUE #1061: Determine why the discovery permissions set in the my_controller_behavior have no affect on blacklight
Why does blacklight now ignore the discovery_permissions set here:
https://github.com/projecthydra/sufia/blob/master/app/controllers/concerns/sufia/my_controller_behavior.rb#L59-L62

I have a feeling that it is related to the new search builder architecture and that needs to be set at the search builder level.  

I did some crazy overrides in ScholarSphere to get us into production: https://github.com/psu-stewardship/scholarsphere/commit/28287c1c8637178e99c6a7a8fdfca4aa5fbee1f6

I doubt this would be the recommended course

cam156:
…filter for only items I have edit access to.  fixes #1061

### DIFF
--- a/app/search_builders/sufia/my_search_builder_behavior.rb
+++ b/app/search_builders/sufia/my_search_builder_behavior.rb
@@ -1,0 +1,20 @@
+# Behavior for the my controller search builder to allow the my controller to only get access to
+# items I have edit access to.
+#
+# Separating as a behavior allows for easier overriding of the gem behavior.
+module Sufia::MySearchBuilderBehavior
+  extend ActiveSupport::Concern
+
+  included do
+    include Blacklight::Solr::SearchBuilderBehavior
+  end
+
+  include Hydra::AccessControlsEnforcement
+  include Sufia::SearchBuilder
+
+  # show only files with edit permissions in lib/hydra/access_controls_enforcement.rb apply_gated_discovery
+  def discovery_permissions
+    ["edit"]
+  end
+
+end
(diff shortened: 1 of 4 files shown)


app/search_builders/sufia/my_search_builder_behavior.rb
```diff
@@ -0,0 +1,14 @@
...
+#
+# Separating as a behavior allows for easier overriding of the gem behavior.
+module Sufia::MySearchBuilderBehavior
+  include Blacklight::Solr::SearchBuilderBehavior
+  include Hydra::AccessControlsEnforcement
+  include Sufia::SearchBuilder
+
+  # show only files with edit permissions in lib/hydra/access_controls_enforcement.rb apply_gated_discovery
+  def discovery_permissions
+    @discovery_permissions = ["edit"]
```
jcoyne: What's the point of the ivar?

cam156: @jcoyne I was matching up with how the behavior works in hydra-access-controls. https://github.com/projecthydra/hydra-head/blob/master/hydra-access-controls/lib/hydra/access_controls_enforcement.rb#L81

jcoyne: But that only sets the ivar if it's not already set (default value). There's also a setter for discovery_permissions in there, which allows you to change the value of the ivar. In the class you provided, I think the ivar is unnecessary.

cam156: @jcoyne ok I changed it to remove it.

jcoyne: It doesn't appear that the change got pushed.

PR #1176, opened 2015-06-01
Adding a configurable sleep to retry unless.  default is set to zero …

### LINKED ISSUE #1059: Retry unless may need to sleep before retrying
Before you re call the block allow for a configurable pause would be nice https://github.com/projecthydra/sufia/blob/master/sufia-models/lib/sufia/models/utils.rb#L16

In certain situations when you want to retry giving the system a break before the retry is nice.

We overrode this in ScholarSphere here to allow for that: https://github.com/psu-stewardship/scholarsphere/commit/c6c4262f0ca20190e8e6b789ecab24f04289dfd4

cam156:
…or no sleep to not change the original behavior. fixes #1059

### DIFF
--- a/sufia-models/lib/sufia/models/engine.rb
+++ b/sufia-models/lib/sufia/models/engine.rb
@@ -30,6 +30,7 @@ module Sufia
       config.max_notifications_for_dashboard = 5
       config.activity_to_show_default_seconds_since_now = 24*60*60
       config.arkivo_api = false
+      config.retry_unless_sleep = 0.0
 
       # Defaulting analytic start date to whenever the file was uploaded by leaving it blank
       config.analytic_start_date = nil
--- a/sufia-models/lib/sufia/models/utils.rb
+++ b/sufia-models/lib/sufia/models/utils.rb
@@ -1,12 +1,28 @@
+# This module includes utility methods
 module Sufia
   module Utils
     extend ActiveSupport::Concern
 
+    # retry the block if the conditional call is true unless we hit the maximum tries
+    #
+    # @param number_of_tries [enumerator] maximum number of times to retry the block (eg. 7.times)
+    # @param condition [#call] conditional to call and see if we SHOULD retry
+    # @yeild block [] code you want to run and retry
+    #
+    # @return result of the block call
     def retry_unless(number_of_tries, condition, &block)
       self.class.retry_unless(number_of_tries, condition, &block)
     end
 
     module ClassMethods
+
+      # retry the block if the conditional call is true unless we hit the maximum tries
+      #
+      # @param number_of_tries [enumerator] maximum number of times to retry the block
+      # @param condition [#call] conditional to call and see if we SHOULD retry
+      # @yeild block [] code you want to run and retry
+      #
+      # @return result of the block call
       def retry_unless(number_of_tries, condition, &block)
         raise ArgumentError, "First argument must be an enumerator" unless number_of_tries.is_a? Enumerator
         raise ArgumentError, "Second argument must be a lambda" unless condition.respond_to? :call
@@ -14,6 +30,7 @@ module Sufia
         number_of_tries.each do
           result = block.call
           return result unless condition.call
+          sleep(Sufia.config.retry_unless_sleep) if Sufia.config.retry_unless_sleep > 0
         end
         raise RuntimeError, "retry_unless could not complete successfully. Try upping the # of tries?"
       end


sufia-models/lib/sufia/models/engine.rb
```diff
@@ -30,6 +30,7 @@ class Engine < ::Rails::Engine
       config.max_notifications_for_dashboard = 5
       config.activity_to_show_default_seconds_since_now = 24*60*60
       config.arkivo_api = false
+      config.rety_unless_sleep = 0.0
```
mjgiarlo: `s/rety/retry/`

sufia-models/lib/sufia/models/utils.rb
```diff
@@ -1,12 +1,26 @@
+# This module includes a utility methods
```
mjgiarlo: `s/a //`

sufia-models/lib/sufia/models/utils.rb
```diff
@@ -1,12 +1,26 @@
...
       self.class.retry_unless(number_of_tries, condition, &block)
     end
 
     module ClassMethods
+
+      # retry the block if the conditional call is true unless we hit the maximum tries
+      #
+      # @param number_of_tries [Fixnum] maximum number of times to retry the block
+      # @param condition [Boolean] conditional to call and see if we SHOULD retry
+      #
```
mjgiarlo: Would it be helpful to write anything about the block using something like https://gist.github.com/chetan/1827484#blocks?

sufia-models/lib/sufia/models/utils.rb
```diff
@@ -1,12 +1,26 @@
+# This module includes a utility methods
 module Sufia
   module Utils
     extend ActiveSupport::Concern
 
+    # retry the block if the conditional call is true unless we hit the maximum tries
+    #
+    # @param number_of_tries [Fixnum] maximum number of times to retry the block
+    # @param condition [Boolean] conditional to call and see if we SHOULD retry
```
mjgiarlo: For the condition param, we expect not to have a Boolean passed but rather something callable. Maybe this should be `s/Boolean/#call/`

sufia-models/lib/sufia/models/utils.rb
```diff
@@ -1,12 +1,26 @@
+# This module includes a utility methods
 module Sufia
   module Utils
     extend ActiveSupport::Concern
 
+    # retry the block if the conditional call is true unless we hit the maximum tries
+    #
+    # @param number_of_tries [Fixnum] maximum number of times to retry the block
```
mjgiarlo: This method doesn't expect a Fixnum for number_of_tries but rather an Enumerator, so perhaps `s/Fixnum/#each/`

sufia-models/lib/sufia/models/utils.rb
```diff
@@ -14,6 +28,7 @@ def retry_unless(number_of_tries, condition, &block)
         number_of_tries.each do
           result = block.call
           return result unless condition.call
+          sleep(Sufia.config.retry_unless_sleep) if Sufia.config.retry_unless_sleep > 0
```
mjgiarlo: What is the `if` protecting against here?

cam156: Making it backward compatible since no sleep was called previously

mjgiarlo: What value would that be set to other than a non-zero value?  Zero, the default?  If so, doesn't sleeping for zero preserve backward compatibility?

cam156: not necessarily.   Certainly people can do things like expect(sleep) and our new sleep might trip that up.  It's mostly for testing, but I do not see the harm...

mjgiarlo: OK, no biggie.

sufia-models/lib/sufia/models/utils.rb
cam156: @mjgiarlo  If you have something to say about it I can add it.  Otherwise it is a ruby concept people should know. Looking at the examples from the contributing doc, the one that passed a block did not comment on it.  In addition the getting started had no blocks.

cam156:
@mjgiarlo "fixed" up your comments on the parameters.  See if what I put fits better.

sufia-models/lib/sufia/models/utils.rb
```diff
@@ -1,12 +1,26 @@
...
     def retry_unless(number_of_tries, condition, &block)
       self.class.retry_unless(number_of_tries, condition, &block)
     end
 
     module ClassMethods
+
+      # retry the block if the conditional call is true unless we hit the maximum tries
+      #
+      # @param number_of_tries [enumerator] maximum number of times to retry the block
+      # @param condition [Boolean] conditional to call and see if we SHOULD retry
```
mjgiarlo: Similar to the above, `s/Boolean/#call/`

sufia-models/lib/sufia/models/utils.rb
```diff
@@ -1,12 +1,26 @@
+# This module includes utility methods
 module Sufia
   module Utils
     extend ActiveSupport::Concern
 
+    # retry the block if the conditional call is true unless we hit the maximum tries
+    #
+    # @param number_of_tries [Fixnum] maximum number of times to retry the block (eg. 7.times)
```
mjgiarlo: `s/Fixnum/enumerator/` for consistency with the method defined below

grosscol:
Per @cam156 in regards to retry_unless method, "It is used mostly in ScholarSphere to retry calls against ldap."  It is generic enough to warrant addition to Sufia.

PR #1177, opened 2015-06-02
Adding a coffescript test for testing the terms of service javascript…

cam156:
… and single use link javascript

Updated the spec runner to not fail on multiples of 10 and to output the number of jasmine tests run on passing to make it more transparant how many jasmine tests are running.

### DIFF
--- a/spec/javascripts/jasmine_spec.rb
+++ b/spec/javascripts/jasmine_spec.rb
@@ -1,16 +1,27 @@
 require 'spec_helper'
 
+# Run the jasmine tests by running the jasmine:ci rake command and capturing the output
+#   The spec will fail if any jasmine tests fails
+#
+#   If you have a jasmine syntax error the test will fail since there will not be zero failures
+#
+#   If you add a new jasmine test by adding a file to spec/javascripts/*spec.js* make sure the
+#   number of test run increments or you may have a syntax error inside your jasmine test
+#
 describe "Jasmine" do
   it "expects all jasmine tests to pass" do
     load_rake_environment ["#{jasmine_path}/lib/jasmine/tasks/jasmine.rake"]
     jasmine_out = run_task 'jasmine:ci'
     unless jasmine_out.include?  "0 failures"
-      puts "************************  Jasmine Output *************\n\n"
+      puts "\n\n************************  Jasmine Output *************"
       puts jasmine_out
       puts "************************  Jasmine Output *************\n\n"
+    else
+      js_specs_count = Dir['spec/javascripts/**/*_spec.js*'].count
+      puts "#{jasmine_out.match(/\n(.+) specs/)[1]} jasmine specs run (in #{js_specs_count} jasmine test files)"
     end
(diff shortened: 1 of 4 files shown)


spec/javascripts/jasmine_spec.rb
```diff
@@ -1,16 +1,26 @@
...
   it "expects all jasmine tests to pass" do
     load_rake_environment ["#{jasmine_path}/lib/jasmine/tasks/jasmine.rake"]
     jasmine_out = run_task 'jasmine:ci'
     unless jasmine_out.include?  "0 failures"
-      puts "************************  Jasmine Output *************\n\n"
+      puts "\n\n************************  Jasmine Output *************"
       puts jasmine_out
       puts "************************  Jasmine Output *************\n\n"
+    else
+      puts "#{jasmine_out.match(/\n(.+) specs/)[1]} jasmine specs run"
```
mjgiarlo: This is helpful. I wonder if this line might also surface how many spec files are actually there, per your comment on line 8. Something like:

```ruby
  js_specs_count = Dir['spec/javascript/**/*_spec.js*].count
  puts "#{jasmine_out.match(/\n(.+) specs/)[1]} jasmine specs run (out of #{js_specs_count})"
```

spec/javascripts/jasmine_spec.rb
```diff
@@ -1,16 +1,26 @@
 require 'spec_helper'
 
+# Run the jasmine tests by running the jasmine:ci rake command and capturing the output
+#   The spec will pass if any jasmine tests pass
```
mjgiarlo: Should this be the behavior, or should it be a failure if any fail?

spec/javascripts/jasmine_spec.rb
```diff
@@ -1,16 +1,26 @@
...
+#
+#   If you have a jasmine syntax error the test will fail since there will not be zero failures
+#
+#   If you add a new jasmine test by adding a file to spec/javascript/*spec.js* make sure the
```
mjgiarlo: `s/javascript/javascripts/`

spec/javascripts/jasmine_spec.rb
cam156: @mjgiarlo That only works if you have one test per file.  You can have many tests in one file.
I could change you line to say .. In <N> spec files

spec/javascripts/jasmine_spec.rb
cam156: @mjgiarlo It is just a mistake in the comments.  It fails if any test fails.

mjgiarlo: Ah, :ok: then!

spec/javascripts/jasmine_spec.rb
mjgiarlo: Oops, right.

cam156:
@mjgiarlo Thanks for the comments.  Fixed up the areas and added your suggestion.

mjgiarlo:
:shipit:

PR #1178, opened 2015-06-04
Allow the GenericFile show page to be loaded exclusively from solr

cam156:
Needed to adjust Audits and Characterization slightly to accommodate the solr backed GenericFile not having access to the same api as the fedora backed GenericFile.

### DIFF
--- a/app/controllers/concerns/sufia/files_controller_behavior.rb
+++ b/app/controllers/concerns/sufia/files_controller_behavior.rb
@@ -39,11 +39,22 @@ module Sufia
       before_filter :has_access?, except: [:show]
       before_filter :build_breadcrumbs, only: [:show, :edit, :stats]
       load_resource only: [:audit]
-      load_and_authorize_resource except: [:index, :audit]
+      load_and_authorize_resource except: [:index, :audit, :show]
+
+      # since we are only displaying data we are makingthe show action fast by loading the information from solr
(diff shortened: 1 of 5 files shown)


app/controllers/concerns/sufia/files_controller_behavior.rb
```diff
@@ -39,13 +39,20 @@ module FilesControllerBehavior
...
+      authorize_resource only: [:show]
 
       class_attribute :edit_form_class, :presenter_class
       self.edit_form_class = Sufia::Forms::GenericFileEditForm
       self.presenter_class = Sufia::GenericFilePresenter
     end
 
+    def load_resource_from_solr
+      @generic_file = ::GenericFile.load_instance_from_solr(params[:id])
+      @generic_file
```
jcoyne: This line can be omitted I think.

jcoyne:
Should there be a test for the show view?

cam156:
@jcoyne I was not changing the view, and made sure the controller did not find the GenericFile.  There is already a test for the show view here: https://github.com/projecthydra/sufia/blob/master/spec/views/generic_file/show.html.erb_spec.rb
Is there additional test that you are thinking about?

jcoyne:
@cam156 Ah, I see. You are still returning a GenericFile object, just one loaded from solr. I was thinking you were returning a SolrDocument. So never mind.

PR #1181, opened 2015-06-05
Adding admin stats to sufia by copying them from ScholarSphere

cam156:


app/models/stats_admin.rb
```diff
@@ -0,0 +1,8 @@
+# -*- coding: utf-8 -*-
+class StatsAdmin
```
mjgiarlo: Hmm, is this the right spot for this class?  For other routing constraint classes, we put them in ./sufia-models/lib/generators/sufia/models/templates/config/ and copy them into apps using generators.

app/controllers/admin/stats_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+    depositor_tuples = JSON.parse(depositors_json)['terms'][depositor_key] rescue []
+    # Change to hash where keys = logins and values = counts
+    @active_users = Hash[*depositor_tuples]
+
+    # Count of documents by permissions
+    ActiveFedora::SolrService.instance.conn.commit("expungeDeletes"=>true)
+    @files_count = {}
+    @files_count[:total] = GenericFile.count
+    @files_count[:public] = GenericFile.where(Solrizer.solr_name('read_access_group', :symbol) => 'public').count
+    @files_count[:psu] = GenericFile.where(Solrizer.solr_name('read_access_group', :symbol) =>'registered').count
```
mjgiarlo: I'm thinking we should `s/psu/registered/` across this PR.

config/routes.rb
```diff
@@ -131,6 +131,12 @@
     end
   end
 
+  namespace :admin do
```
mjgiarlo: I'm thinking this should follow the pattern above, both in terms of namespacing `StatsAdmin` as `Sufia::StatsAdmin` and in terms of checking to make sure that is defined first.

spec/controllers/admin_stats_controller_spec.rb
```diff
@@ -0,0 +1,86 @@
+require 'spec_helper'
+
+describe Admin::StatsController, type: :controller do
+  let(:user1) { FactoryGirl.find_or_create(:user) }
+  let(:user2) { FactoryGirl.find_or_create(:archivist) }
+  
+  before do
+    allow(user1).to receive(:groups).and_return(['umg/up.dlt.scholarsphere-admin-viewers'])
```
mjgiarlo: Should the groups returned here be `['admin']` to match up the the check in the `StatsAdmin` class, rather than the PSU-specific value?

spec/controllers/admin_stats_controller_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+    allow(user2).to receive(:groups).and_return(['umg/up.dlt.some-other-group'])
```
mjgiarlo: Similarly, maybe this should be something like `['notadmin']`

spec/controllers/admin_stats_controller_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+      end
+      it "should provide accurate files_count, ensuring that solr deletes have been expunged first" do
+        get :index
+        expect(assigns[:files_count][:total]).to eq(original_files_count - 1)
+      end
+    end
+
+    describe "counts" do
+      before do
+        GenericFile.new.tap do |f|
```
mjgiarlo: Simplify to `GenericFile.create { |f| f.apply_depositor_metadata(user1.user_key) }`

spec/controllers/admin_stats_controller_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+          f.apply_depositor_metadata(user1.user_key)
+          f.save
+        end
+        GenericFile.new.tap do |f|
```
mjgiarlo: Simplify to 

```ruby
GenericFile.create(read_groups = ['public']) do |f|
  f.apply_depositor_metadata(user1.user_key)
end
```

spec/controllers/admin_stats_controller_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+          f.apply_depositor_metadata(user1.user_key)
+          f.read_groups= ['public']
+          f.save
+        end
+        GenericFile.new.tap do |f|
```
mjgiarlo: Simplify to 

```ruby
GenericFile.create(read_groups: ['registered']) do |f|
  f.apply_depositor_metadata(user1.user_key)
end
```

spec/controllers/admin_stats_controller_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+          f.apply_depositor_metadata(user1.user_key)
+          f.read_groups= ['registered']
+          f.save
+        end
+        Collection.new.tap do |c|
```
mjgiarlo: Simplify to 

```ruby
Collection.create(title: 'test') do |c|
  c.apply_depositor_metadata(user1.user_key)
end
```

config/routes.rb
cam156: @mjgiarlo What do you mean in terms of namespacing?  admin is the same namespace isn't it?

cam156: @mjgiarlo nm.  I thought you were talking about route namespacing

app/models/stats_admin.rb
cam156: @mjgiarlo I'm not sure why we are generating it into the app.  For easier overriding?  With it being defined we do not need to check the route to see if it exists, so the logic is simpler.

mjgiarlo: Yep, for easier overriding.

app/controllers/admin/stats_controller.rb
```diff
@@ -0,0 +1,49 @@
+# -*- encoding : utf-8 -*-
+
+class Admin::StatsController < ApplicationController
+  def index
```
jcoyne: This method is really long. Can you break it up into a few smaller methods? Perhaps: `recent_users()`, `active_users()` and `files_count()`

app/views/admin/stats/index.html.erb
```diff
@@ -0,0 +1,66 @@
+<div class="stats">
+<h1>Statistics for <%= application_name %>
```
jcoyne: Indent

app/views/admin/stats/index.html.erb
```diff
@@ -0,0 +1,66 @@
...
+<%= Sufia::VERSION %></h1>
+<br />
+<h3>Total Files: <%= @files_count[:total] %> </h3>
```
jcoyne: Skipped `h2`. Not a big deal to me, but does that conform with your usability guidelines?

app/views/admin/stats/index.html.erb
```diff
@@ -0,0 +1,66 @@
...
+  <li><%= k  %> <span class="count">(<%= v %>)</span></li>
+<% end %>
+</ul>
+
+<br/>
+<h3>Total <%= application_name %> Users:&nbsp; <%= @users_count %> </h3>
+<br/>
+
+<h4>Newest Users</h4>
+  <%= form_for "users_stats", url: sufia.admin_stats_path, method:"GET" do |f| %>
```
jcoyne: Add space before `"GET"`

app/views/admin/stats/index.html.erb
```diff
@@ -0,0 +1,66 @@
...
+    <div>Total: <%= @recent_users.count %></div>
+  <%- else %>
+    Five most recent users:
+  <%- end %>
+
+<ul>
+  <% @recent_users.each do |usr| %>
+    <li>
+      <a href="<%= sufia.profile_path(usr.user_key) %>" title="View user's profile"><%= usr.name %></a>
+      <%- if usr.department %>
```
jcoyne: the dash is unnecessary.

config/routes.rb
```diff
@@ -131,6 +131,14 @@
     end
   end
 
+  if defined?(Sufia::StatsAdmin)
+    namespace :admin do
+      constraints Sufia::StatsAdmin do
```
jcoyne: Why are we doing it this way vs adding an ability?

app/controllers/admin/stats_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+    if @users_stats[:start_date]
+      #@user_stats[:start_date] ||= 7.days.ago
+      @recent_users = User.where('created_at >= ?',  @users_stats[:start_date])
+    else
+      @recent_users = User.order('created_at DESC').limit(5)
+    end
+
+    # Query Solr for top 5 depositors
+    depositor_key = Solrizer.solr_name('depositor', :stored_searchable, type: :string)
+    top_depositors_url = "#{ActiveFedora.solr_config[:url]}/terms?terms.fl=#{depositor_key}&terms.sort=count&terms.limit=#{top_depositors_count}&wt=json&omitHeader=true"
```
jcoyne: This API call really ought to be in a service class.

config/routes.rb
cam156: @jcoyne As far as I could tell by reading the CanCanCan docs the ability is added based on a model.  We do not have a model defined for the stats.  Additionally this matches with the resque admin routes, so it is not a new pattern

app/controllers/admin/stats_controller.rb
cam156: @jcoyne Broken up and refactored a bit.

config/routes.rb
jcoyne: Not saying you have to, but you can do `can :show, :stats if user.admin?` https://github.com/ryanb/cancan/wiki/Non-RESTful-Controllers

app/controllers/admin/stats_controller.rb
```diff
@@ -0,0 +1,70 @@
...
+    @files_count = document_by_permission
+
+    # Query Solr for top file formats
+    @top_formats = top_formats
+
+    render 'index'
+  end
+
+  private 
+  def top_depositors_count
```
jcoyne: Can you add an extra level of indent for private methods. http://guides.rubyonrails.org/contributing_to_ruby_on_rails.html#follow-the-coding-conventions

app/controllers/admin/stats_controller.rb
```diff
@@ -0,0 +1,70 @@
...
+    count = params[:dep_count].to_i 
+    count.in?(5..20) ? count : 5
+  end
+
+  def document_by_permission
+    # make sure everything is up to date in solr
+    ActiveFedora::SolrService.instance.conn.commit("expungeDeletes"=>true)
```
jcoyne: I've never seen `expungeDeletes` before. Is this something that is necessary?  Does it point to a bug elsewhere in the stack?

jcoyne: From what I can tell in the solr documentation, expungeDeletes has to do with how the index is structured on disk, but shouldn't change the returned results, so can we remove this?

cam156: @jcoyne I was copying the code from scholarsphere.  It came from this commit https://github.com/psu-stewardship/scholarsphere/commit/b3a531c6d61783858803adc37da4a4dabebed7ec#diff-eb2a74120fa8e97a8ce630c4a345ac0a

Not really sure if it can be removed, but since you believe it can I will see if the test still passes.

config/routes.rb
cam156: @jcoyne That is still at the controller level, so the route exists but they are not authorized.  It is a slightly different pattern than we are currently using for administrative routes.  I do not mind changing the pattern overall, but that is not within scope of this PR.  I can add a ticket that can be addressed later.

app/controllers/admin/stats_controller.rb
jcoyne: Question everything @flyingzumwalt does ;)

flyingzumwalt: I certainly do.

cam156:
@jcoyne removed the line and the tests still passes so I guess it it no longer need.  Good catch.

PR #1182, opened 2015-06-06
Restrict versions of bootstrap_form

awead:
Enables bundle install to finish under Sufia 5.0.


awead:
Travis fails intermittently, but the affected test passes locally always.

jcoyne:
Please say more about what happens before this change is made. Why is it a problem to have a newer version?

awead:
there was a conflict with hydra-editor. I needed to get a release out so I could use it with are upcoming ArchiveSphere release.

PR #1183, opened 2015-06-08
Fixes background color of facet X when filtering.

mtribone:
Collision with .remove class https://github.com/psu-stewardship/scholarsphere/issues/136

### DIFF
--- a/app/assets/stylesheets/sufia/_buttons.scss
+++ b/app/assets/stylesheets/sufia/_buttons.scss
@@ -47,6 +47,10 @@ ul.listing .add, ul.listing .remove {
   color: $remove-text-hover;
 }
 
+span.facet-label .remove {
+  background-color: $panel-background-color;
+}
+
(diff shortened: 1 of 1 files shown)


jcoyne:
What is it clashing with? Where is the `.remove` with the red background defined?

mtribone:
It uses the same class name for the X button on "You searched for"

screen shot 2015-06-08 at 4 11 41 pm

mtribone:
And also the remove button on the metadata page

screen shot 2015-06-08 at 4 13 36 pm

mtribone:
.remove is in _buttons.scss file in assets > stylesheets > sufia

jcoyne:
Why don't we make _buttons.css more specific to apply it's syles only where needed, rather than adding more specific styles that keep the defaults everwhere else?

mtribone:
Gotcha. I'll rework it.

PR #1184, opened 2015-06-08
Make direct query to Solr for the file size property which gets stored

narogers:
This should speed up the process of counting byte totals for a collection because it only pulls the file size and only from Solr. You do need to run ActiveFedora::Base.reindex_everything! to add the new file_size_isi field first.


sufia-models/app/models/concerns/sufia/collection_behavior.rb
```diff
@@ -15,10 +15,29 @@ def update_permissions
...
+      # process
+      args = {
+        :fq => "{!join from=hasCollectionMember_ssim to=id}id:#{self.id}",
+        :fl => "id,file_size_isi",
+        :rows => self.members.count
+      }
+
+      files = ActiveFedora::SolrService.query(query, args)
+
+      byte_count = 0
```
cam156: @narogers These 5 lines could become  files.reduce(0) { |sum, f| sum + f['file_size_isi'] }   Then you would not need the local variable.

cam156:
@narogers it looks like the id of you collection is null when you pass it to solr, so solr is not parsing it.  You can either save the collection in the test (which makes the test slower) or pass and id: to the new operator

cam156:
@narogers it look like there was some issue with the rebase.  Your commit has others in it.

narogers:
I'm not sure what happened so I am just going to close this PR and reopen a new one using the clean branch (without all the commits attached to it)

cam156:
@narogers you are closing this PR correct and creating a new one?

narogers:
Once I find time to rework the spec test I will create a new PR sans all the accidental merging of recent changes to the master branch

narogers:
Closed: See the updated version minus all the extra merges above.

PR #1192, opened 2015-06-10
Refactors CSS to keep defaults intact and more specifically targets remove

mtribone:
Metadata edit for remove is using btn-danger, so I just needed to add span.appliedFilter to .remove to get the desired result. Facet X no longer has red background.


mtribone:
Addresses https://github.com/projecthydra/sufia/pull/1183#issuecomment-110379984

jcoyne:
Thanks!

PR #1193, opened 2015-06-10
Adding user email list rake task.  fixes #1060

### LINKED ISSUE #1060: Add a user listing rake task
This task might be nice for others to have:
https://github.com/psu-stewardship/scholarsphere/blob/05d8448240a6f5aeb4a9e7df9a05e8cc665a24a7/lib/tasks/scholarsphere.rake#L359-L365

cam156:


tasks/sufia-user.rake
```diff
@@ -0,0 +1,14 @@
+namespace :sufia do
+  namespace :user do
+
+    desc "list user's email"
+    task "list_emails", [:file_name] => :environment do |cmd, args|
+      file_name = args[:file_name]
```
jcoyne: If you were to make the meat of this task be `UserEmaiReport.new(file_name).write` it would be easier to test. ;)

jcoyne:
:+1:

tasks/sufia-user.rake
cam156: @jcoyne since it is just a helper method it makes more sense to keep all the code in one place.

PR #1196, opened 2015-06-10
removing useless assignemnt to sanatize_text varriable

cam156:


jcoyne:
Heh. Copy and paste much?

PR #1197, opened 2015-06-10
Surrounding regular expressions being passed to a method by parens to…

cam156:
… distiguish them from division


jcoyne:
Doesn't the second slash on the line already distinguish them from division?  The Ruby grammer already must be able to distinguish this or you'd get a SyntaxError.

cam156:
rubocop is flagging them so I was adding them.  The grammer does not care about the whitespace, but we seem to...

Here is the rubocop output:  Ambiguous regexp literal. Parenthesize the method arguments if it's surely a regexp literal, or add a whitespace to the right of the / if it should be a division

jcoyne:
Okay (rubocop is dumb).

PR #1199, opened 2015-06-10
removing unecissary local variables identified by rubocop

cam156:


app/controllers/concerns/sufia/files_controller/browse_everything.rb
```diff
@@ -24,7 +24,7 @@ def create_from_browse_everything(params)
       # Generic utility for creating GenericFile from a URL
       # Used in to import files using URLs from a file picker like browse_everything 
       def create_file_from_url(url, file_name)
-        generic_file = ::GenericFile.new(import_url: url, label: file_name).tap do |gf|
+        ::GenericFile.new(import_url: url, label: file_name).tap do |gf|
```
jcoyne: No need to tap, new takes a block

app/controllers/concerns/sufia/files_controller/local_ingest_behavior.rb
```diff
@@ -54,7 +54,7 @@ def ingest_local_file
     def ingest_one(filename, unarranged)
       basename = File.basename(filename)
       # do not remove ::
-      generic_file = ::GenericFile.new(label: basename).tap do |gf|
+      ::GenericFile.new(label: basename).tap do |gf|
         gf.relative_path = filename if filename != basename
```
jcoyne: No need to tap, new takes a block

cam156:
@jcoyne Ok.  Is that new?  Our specs are riddled with the tap behavior.

PR #1203, opened 2015-06-11
Feature/unfeature from item page

### LINKED ISSUE #1202: Unfeature Item from File Page
We're working on this here, just putting the issue in ahead of time to make sure there's no complaints. This just makes it so there's an "Unfeature" link.

luisgreg99:
We added the ability to un-feature an item from the item page. We chose to add the link to improve the workflow to feature and un-feature an item. Fixes #1202


app/assets/javascripts/sufia/featured_works.js
```diff
@@ -31,20 +31,37 @@ function dragAndDrop(selector) {
 }
 
 Blacklight.onLoad(function() {
-  $('a[data-behavior="feature"]').on('click', function(evt) {
+  $('#content-wrapper').on('click', 'a[data-behavior="feature"]', function(evt) {
     evt.preventDefault();
+    evt.stopPropagation();
```
jcoyne: Can you please explain why stopPropagation is required here?

spec/features/featured_item_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+
+RSpec.describe "featuring items" do
+  context "when viewing a featured item" do
+    let(:user) do
+      u = FactoryGirl.build(:user)
+      u.group_list = "admin"
+      u.save
+      u
+    end
+    it "should have a working unfeature link", :js => true do
```
awead: `:js => true` isn't needed because Poltergeist in the default driver for all feature tests.

spec/features/featured_item_spec.rb
```diff
@@ -0,0 +1,26 @@
+require 'spec_helper'
+
+RSpec.describe "featuring items" do
```
awead: We usually just put `describe` instead of `RSpec.describe` although I'm not sure it matters.

app/assets/javascripts/sufia/featured_works.js
terrellt: We're not sure why. Without it, it runs the JS, fixes the link, and then redirects to the JSON despite preventDefault.

terrellt: The feature spec will fail without it too.

terrellt:
@luisgreg99 has a CLA on file.

app/assets/javascripts/sufia/featured_works.js
```diff
@@ -31,20 +31,37 @@ function dragAndDrop(selector) {
...
        success: function(data) {
-         anchor.before("Featured");
+         anchor.before('<a data-method="post" data-behavior="unfeature-page" href="'+anchor.attr('href')+'">Unfeature</a>');
          anchor.remove();
        }
     });
   });
 
-  $('a[data-behavior="unfeature"]').on('click', function(evt) {
+  $('#content-wrapper').on('click', 'a[data-behavior="unfeature-page"]', function(evt) {
```
jcoyne: `#content-wrapper` appears on every page, but we only need to register this handler for pages that have the button on them, so can you add a narrower selector to the `show_actions` partial?

spec/features/featured_item_spec.rb
terrellt: Ah - most of our projects have monkeypatching in rspec turned off. Apologies.

app/views/generic_files/_show_actions.html.erb
```diff
@@ -19,7 +19,7 @@
           <%= link_to "Feature", sufia.featured_work_path(@generic_file, format: :json), data: {behavior: 'feature'} %>
         <% elsif @generic_file.featured? %>
           &nbsp;|&nbsp;
-          Featured
+          <%= link_to "Unfeature", sufia.featured_work_path(@generic_file, format: :json), data: {behavior: 'unfeature-page'} %>
```
awead: It would be nice to "Unfeature" in the I18n config, but that impacts the JS that's getting appended, so I'm not sure how possible that is.

awead:
@luisgreg99 if you can get us a signed iCLA, then we can move forward with the merging process. You can email legal@projecthydra.org to get the process started.

terrellt:
@awead Should be on file already.

awead:
@terrellt @luisgreg99 Indeed it is. I just confirmed... It's my turn to offer apologies!

terrellt:
@awead In the future, is there a good way to confirm that an iCLA is on file?

awead:
@terrellt There's a link from this page: https://wiki.duraspace.org/display/hydra/The+Hydra+Project
to "Hydra Licensed Contributors" which lists all the folks that have their iCLAs on file, but it requires a login. If you don't have one, request one 'cause you all should have that.

spec/features/featured_item_spec.rb
jcoyne: I don't think that's true: 
https://github.com/projecthydra/sufia/blob/master/spec/spec_helper.rb#L31-L32

awead: @jcoyne I thought it was getting specified here:
https://github.com/projecthydra/sufia/blob/master/spec/support/poltergeist.rb#L10-11
but it's only for feature tests. Your above comment should allow specifying a different driver for view tests.

app/assets/javascripts/sufia/featured_works.js
luisgreg99: @jcoyne In my last commit, I updated the selector for the show_actions partial.

spec/features/featured_item_spec.rb
cam156: @jcoyne @awead was this determined?  Does @luisgreg99 need to remove the js:true?

app/views/generic_files/_show_actions.html.erb
cam156: You can create a js.erb file that will translate the i18n in the asset pipeline.  Here is an example of a js.erb file: https://github.com/projecthydra/sufia/blob/master/app/assets/javascripts/sufia/single_use_link.js.erb#L13

spec/features/featured_item_spec.rb
jcoyne: I think we should leave it because it communicates that the test depends on js. As another issue, I dont think we should always use the js driver, it's slower.

spec/features/featured_item_spec.rb
luisgreg99: @awead, @terrellt: `RSpec.describe` has been replaced with `describe`

spec/features/featured_item_spec.rb
awead: Yes, leave it in. You'll need it now that #1206 is in the queue. @luisgreg99 if you could remove the hashrocket, i.e. `js: true`, we're good to go. You know, because @mjgiarlo.  :wink:

cam156:
@luisgreg99 This PR needs to be rebased with the latest master.

PR #1204, opened 2015-06-16
Support collection size calculations using a Solr query instead of lo…

narogers:
Here is a clean version of the patch for making collection sizes calculate without being tied to the byte count. Other than passing tests it should be identical to the original pull request at https://github.com/projecthydra/sufia/pull/1184

### DIFF
--- a/spec/models/collection_spec.rb
+++ b/spec/models/collection_spec.rb
@@ -3,7 +3,7 @@ require 'spec_helper'
 describe Collection, :type => :model do
   before do
     @user = FactoryGirl.create(:user)
-    @collection = Collection.new(title: "test collection").tap do |c|
+    @collection = Collection.new(id: 'mock-collection-with-members', title: "test collection") do |c|
       c.apply_depositor_metadata(@user.user_key)
     end
   end
@@ -25,11 +25,26 @@ describe Collection, :type => :model do
       it { is_expected.to eq 0 }
     end
 
-    context "with two 50 byte files" do
-      let(:bitstream) { double("content", size: "50")}
+    context "with three 33 byte files" do
+      let(:bitstream) { double("content", size: "33")}
       let(:file) { mock_model GenericFile, content: bitstream }
-      before { allow(@collection).to receive(:members).and_return([file, file]) }
-      it { is_expected.to eq 100 }
+      let(:documents) do
+        [{ 'id' => 'file-1', 'file_size_is' => 33 }, 
+         { 'id' => 'file-2', 'file_size_is' => 33 }, 
+         { 'id' => 'file-3', 'file_size_is' => 33 }]
+      end
+      let(:query) { ActiveFedora::SolrQueryBuilder.construct_query_for_rel(has_model: ::GenericFile.to_class_uri) }
+      let(:args) do
+        { fq: "{!join from=hasCollectionMember_ssim to=id}id:#{@collection.id}",
+        fl: "id, file_size_is",
+        rows: 3 }
+      end
+
+      before do
+        allow(@collection).to receive(:members).and_return([file, file, file])
+        allow(ActiveFedora::SolrService).to receive(:query).with(query, args).and_return(documents)
+      end
+      it { is_expected.to eq 99 }
     end
 
   end
--- a/sufia-models/app/models/concerns/sufia/collection_behavior.rb
+++ b/sufia-models/app/models/concerns/sufia/collection_behavior.rb
@@ -15,10 +15,37 @@ module Sufia
       self.visibility = "open"
     end
 
-    # Compute the sum of each file in the collection
-    # Return an integer of the result
+    # Compute the sum of each file in the collection using Solr to
+    # avoid having to hit Fedora 
+    #
+    # Returns an integer 
     def bytes
-      members.reduce(0) { |sum, gf| sum + gf.content.size.to_i }
+      query = ActiveFedora::SolrQueryBuilder.construct_query_for_rel(has_model: file_model)
+      args = {
+        fq: "{!join from=hasCollectionMember_ssim to=id}id:#{id}",
+        fl: "id, #{file_size_field}",
+        rows: members.count
+      }
+
+      files = ActiveFedora::SolrService.query(query, args)
+      files.reduce(0) { |sum, f| sum += f[file_size_field].to_i }
     end
+
+    protected
+      # Field to look up when locating the size of each file in Solr.
+      # Override for your own installation if using something different
+      def file_size_field
+        Solrizer.solr_name('file_size', stored_integer_descriptor)
+      end
+
+      # Override if you are storing your file size in a different way
+      def stored_integer_descriptor
+        Sufia::GenericFileIndexingService::STORED_INTEGER
+      end
+
+      # Override if not using GenericFiles
+      def file_model
+        ::GenericFile.to_class_uri
+      end
   end
 end
--- a/sufia-models/app/services/sufia/generic_file_indexing_service.rb
+++ b/sufia-models/app/services/sufia/generic_file_indexing_service.rb
@@ -1,11 +1,14 @@
 module Sufia
   class GenericFileIndexingService < ActiveFedora::IndexingService
+    STORED_INTEGER = Solrizer::Descriptor.new(:integer, :stored)
+
     def generate_solr_document
       super.tap do |solr_doc|
         solr_doc[Solrizer.solr_name('label')] = object.label
         solr_doc[Solrizer.solr_name('file_format')] = object.file_format
         solr_doc[Solrizer.solr_name('file_format', :facetable)] = object.file_format
         solr_doc['all_text_timv'] = object.full_text.content
+        solr_doc[Solrizer.solr_name('file_size', STORED_INTEGER)] = object.content.size.to_i
       end
     end
   end
--- a/spec/presenters/sufia/collection_presenter_spec.rb
+++ b/spec/presenters/sufia/collection_presenter_spec.rb
@@ -8,7 +8,7 @@ describe Sufia::CollectionPresenter do
                             :language, :identifier, :based_near, :related_url] }
   end
 
-  let(:collection) { Collection.new(description: 'a nice collection', title: 'A clever title') }
+  let(:collection) { Collection.new(id: 'my-nice-collection', description: 'a nice collection', title: 'A clever title') }
   let(:presenter) { described_class.new(collection) }
 
   describe "#terms_with_values" do


spec/models/collection_spec.rb
```diff
@@ -3,7 +3,7 @@
 describe Collection, :type => :model do
   before do
     @user = FactoryGirl.create(:user)
-    @collection = Collection.new(title: "test collection").tap do |c|
+    @collection = Collection.new(id: 'mock-collection-with-members', title: "test collection").tap do |c|
```
jcoyne: The `.tap` is unnecessary, can you please remove it.

spec/models/collection_spec.rb
```diff
@@ -25,11 +25,31 @@
...
-      before { allow(@collection).to receive(:members).and_return([file, file]) }
-      it { is_expected.to eq 100 }
+      let(:documents) { 
+        [{ 'id' => 'file-1', 'file_size_isi' => 33 }, 
+         { 'id' => 'file-2', 'file_size_isi' => 33 }, 
+         { 'id' => 'file-3', 'file_size_isi' => 33 }
+        ]}
+
+      before { 
+        # This particular line may not be
```
jcoyne: May not be what?

spec/models/collection_spec.rb
```diff
@@ -25,11 +25,31 @@
...
+
+      before { 
+        # This particular line may not be 
+        allow(@collection).to receive(:members).and_return([file, file, file])
+        # Is there a better way to mock this up?
+        allow(ActiveFedora::SolrService).to receive(:query).with(
+            "has_model_ssim:GenericFile",
+            {
+              :fq => "{!join from=hasCollectionMember_ssim to=id}id:#{@collection.id}",
+             :fl => "id, file_size_isi",
```
jcoyne: Check the indent level here. Please use ruby 1.9 hash syntax.

spec/models/collection_spec.rb
```diff
@@ -25,11 +25,31 @@
...
+      let(:documents) { 
+        [{ 'id' => 'file-1', 'file_size_isi' => 33 }, 
+         { 'id' => 'file-2', 'file_size_isi' => 33 }, 
+         { 'id' => 'file-3', 'file_size_isi' => 33 }
+        ]}
+
+      before { 
+        # This particular line may not be 
+        allow(@collection).to receive(:members).and_return([file, file, file])
+        # Is there a better way to mock this up?
```
jcoyne: I don't think there is a better way.

sufia-models/app/models/concerns/sufia/collection_behavior.rb
```diff
@@ -15,10 +15,39 @@ def update_permissions
       self.visibility = "open"
     end
 
-    # Compute the sum of each file in the collection
+    # Compute the sum of each file in the collection using Solr to
+    # avoid having to proces GBs of data for a simple number. Can be
```
jcoyne: typo "proces".  How about just saying "to avoid loading records from Fedora"?

sufia-models/app/models/concerns/sufia/collection_behavior.rb
```diff
@@ -15,10 +15,39 @@ def update_permissions
...
+    #
     # Return an integer of the result
     def bytes
-      members.reduce(0) { |sum, gf| sum + gf.content.size.to_i }
-    end
+      query = "has_model_ssim:#{file_model}"
+      # Cribbed from Hydra::Collections to speed up the resolution
+      # process
+      args = {
+        fq: "{!join from=hasCollectionMember_ssim to=id}id:#{self.id}",
```
jcoyne: the `self.` is unnecessary.

sufia-models/app/models/concerns/sufia/collection_behavior.rb
```diff
@@ -15,10 +15,39 @@ def update_permissions
...
+    # Compute the sum of each file in the collection using Solr to
+    # avoid having to proces GBs of data for a simple number. Can be
+    # cleaned up once it is confirmed to work
+    #
     # Return an integer of the result
     def bytes
-      members.reduce(0) { |sum, gf| sum + gf.content.size.to_i }
-    end
+      query = "has_model_ssim:#{file_model}"
+      # Cribbed from Hydra::Collections to speed up the resolution
```
jcoyne: Please remove this. I don't think it's helpful.

sufia-models/app/models/concerns/sufia/collection_behavior.rb
```diff
@@ -15,10 +15,39 @@ def update_permissions
...
+        rows: members.count
+      }
+
+      files = ActiveFedora::SolrService.query(query, args)
+      files.reduce(0) { |sum, f| sum += f[file_size_field].to_i }
+   end
+
+   # Field to look up when locating the size of each file in Solr.
+   # Override for your own installation if using something different
+   def file_size_field
```
jcoyne: Should this method be private or protected?

sufia-models/app/models/concerns/sufia/collection_behavior.rb
```diff
@@ -15,10 +15,39 @@ def update_permissions
...
+   end
+
+   # Override if you are storing your file size in a different way
+   def stored_integer_descriptor
+     Solrizer::Descriptor.new(:integer, :indexed, :stored)
+   end
+
+   # Override if not using GenericFiles
+   def file_model
+     "GenericFile"
```
jcoyne: How about `::GenericFile.to_class_uri`

sufia-models/app/models/concerns/sufia/collection_behavior.rb
```diff
@@ -15,10 +15,39 @@ def update_permissions
...
-    # Compute the sum of each file in the collection
+    # Compute the sum of each file in the collection using Solr to
+    # avoid having to proces GBs of data for a simple number. Can be
+    # cleaned up once it is confirmed to work
+    #
     # Return an integer of the result
     def bytes
-      members.reduce(0) { |sum, gf| sum + gf.content.size.to_i }
-    end
+      query = "has_model_ssim:#{file_model}"
```
jcoyne: How about `query = ActiveFedora::SolrQueryBuilder.construct_query_for_rel(has_model: file_model)`

sufia-models/app/models/concerns/sufia/collection_behavior.rb
```diff
@@ -15,10 +15,39 @@ def update_permissions
...
+
+   # Field to look up when locating the size of each file in Solr.
+   # Override for your own installation if using something different
+   def file_size_field
+     Solrizer.solr_name('file_size', stored_integer_descriptor)
+   end
+
+   # Override if you are storing your file size in a different way
+   def stored_integer_descriptor
+     Solrizer::Descriptor.new(:integer, :indexed, :stored)
```
jcoyne: Why are we indexing this?  Why not just `(:integer, :stored)`?

sufia-models/app/services/sufia/generic_file_indexing_service.rb
```diff
@@ -6,6 +6,9 @@ def generate_solr_document
         solr_doc[Solrizer.solr_name('file_format')] = object.file_format
         solr_doc[Solrizer.solr_name('file_format', :facetable)] = object.file_format
         solr_doc['all_text_timv'] = object.full_text.content
+
+        int_store = Solrizer::Descriptor.new(:integer, :indexed, :stored);
```
jcoyne: Why are we indexing this?  Why not just `(:integer, :stored)`? Also, how about making this a constant?

sufia-models/app/models/concerns/sufia/collection_behavior.rb
mjgiarlo: Ah, good question. :clap:

sufia-models/app/models/concerns/sufia/collection_behavior.rb
narogers: That's a good question - is there ever any other situation where another class might need access to it? If not then it makes logical sense to convert it to a protected method.

sufia-models/app/models/concerns/sufia/collection_behavior.rb
narogers: Would it ever be useful to sort files by size or otherwise filter on the value? This might be moot with the eventual move to PCDM in which case it makes no sense to index it at all and I can just make it a constant.

spec/models/collection_spec.rb
```diff
@@ -3,7 +3,7 @@
 describe Collection, :type => :model do
   before do
     @user = FactoryGirl.create(:user)
-    @collection = Collection.new(title: "test collection").tap do |c|
+    @collection = Collection.new(id: 'mock-collection-with-members', title: "test collection") do |c|
```
mjgiarlo: Do you need to specify the ID here for some reason?  You don't seem to be using it later in the test, so is it useful?

spec/models/collection_spec.rb
```diff
@@ -25,11 +25,29 @@
...
+        allow(ActiveFedora::SolrService).to receive(:query).with(
+            "_query_:\"{!raw f=has_model_ssim}GenericFile\"",
+            {
+              fq: "{!join from=hasCollectionMember_ssim to=id}id:#{@collection.id}",
+              fl: "id, file_size_is",
+              rows: 3
+            }
+          ).and_return(documents)
+      }
+      it {
```
mjgiarlo: Can you put lines 48-50 on one line?

```ruby
it { is_expected.to eq 99 }
```

spec/models/collection_spec.rb
```diff
@@ -25,11 +25,29 @@
...
-      it { is_expected.to eq 100 }
+      let(:documents) { 
+        [{ 'id' => 'file-1', 'file_size_is' => 33 }, 
+         { 'id' => 'file-2', 'file_size_is' => 33 }, 
+         { 'id' => 'file-3', 'file_size_is' => 33 }
+        ]}
+
+      before { 
+        allow(@collection).to receive(:members).and_return([file, file, file])
+        allow(ActiveFedora::SolrService).to receive(:query).with(
```
mjgiarlo: This `allow` is sprawling a bit. Would you consider something like:

```ruby
let(:query) { "_query_:\"{!raw f=has_model_ssim}GenericFile\"" }
let(:args) do
  {
    fq: "{!join from=hasCollectionMember_ssim to=id}id:#{@collection.id}",
    fl: "id, file_size_is",
    rows: 3
  }
end
before do
  # ...
  allow(ActiveFedora::SolrService).to receive(:query).with(query, args).and_return(documents)
  # ...
end

File: spec/models/collection_spec.rb [hunk 16]
```diff
@@ -25,11 +25,29 @@
...
       let(:file) { mock_model GenericFile, content: bitstream }
-      before { allow(@collection).to receive(:members).and_return([file, file]) }
-      it { is_expected.to eq 100 }
+      let(:documents) { 
+        [{ 'id' => 'file-1', 'file_size_is' => 33 }, 
+         { 'id' => 'file-2', 'file_size_is' => 33 }, 
+         { 'id' => 'file-3', 'file_size_is' => 33 }
+        ]}
+
+      before {
```
mjgiarlo: Use `do` instead of `{` for multi-line blocks.

spec/models/collection_spec.rb
```diff
@@ -25,11 +25,29 @@
...
+        allow(@collection).to receive(:members).and_return([file, file, file])
+        allow(ActiveFedora::SolrService).to receive(:query).with(
+            "_query_:\"{!raw f=has_model_ssim}GenericFile\"",
+            {
+              fq: "{!join from=hasCollectionMember_ssim to=id}id:#{@collection.id}",
+              fl: "id, file_size_is",
+              rows: 3
+            }
+          ).and_return(documents)
+      }
```
mjgiarlo: Use `end` instead of `}` for multi-line blocks.

spec/models/collection_spec.rb
```diff
@@ -25,11 +25,29 @@
...
+      let(:documents) { 
+        [{ 'id' => 'file-1', 'file_size_is' => 33 }, 
+         { 'id' => 'file-2', 'file_size_is' => 33 }, 
+         { 'id' => 'file-3', 'file_size_is' => 33 }
+        ]}
+
+      before { 
+        allow(@collection).to receive(:members).and_return([file, file, file])
+        allow(ActiveFedora::SolrService).to receive(:query).with(
+            "_query_:\"{!raw f=has_model_ssim}GenericFile\"",
```
mjgiarlo: Should you be using `construct_query_for_rel` here for parallelism with the code below?

sufia-models/app/services/sufia/generic_file_indexing_service.rb
```diff
@@ -6,6 +6,7 @@ def generate_solr_document
         solr_doc[Solrizer.solr_name('file_format')] = object.file_format
         solr_doc[Solrizer.solr_name('file_format', :facetable)] = object.file_format
         solr_doc['all_text_timv'] = object.full_text.content
+        solr_doc[Solrizer.solr_name('file_size', Solrizer::Descriptor.new(:integer, :stored);)] = object.content.size.to_i
```
mjgiarlo: I think a stray semi-colon may have crept in between the two closing parens here.

sufia-models/app/models/concerns/sufia/collection_behavior.rb
```diff
@@ -15,10 +15,35 @@ def update_permissions
       self.visibility = "open"
     end
 
-    # Compute the sum of each file in the collection
-    # Return an integer of the result
```
mjgiarlo: Restore this line so folks know what this public method returns.

sufia-models/app/models/concerns/sufia/collection_behavior.rb
```diff
@@ -15,10 +15,35 @@ def update_permissions
...
+        fl: "id, #{file_size_field}",
+        rows: members.count
+      }
+
+      files = ActiveFedora::SolrService.query(query, args)
+      files.reduce(0) { |sum, f| sum += f[file_size_field].to_i }
+    end
+
+    protected
+    # Field to look up when locating the size of each file in Solr.
```
mjgiarlo: Indent all the protected methods two more spaces. (This is a Rails convention, though not a Ruby convention, IIRC.)

sufia-models/app/services/sufia/generic_file_indexing_service.rb
mjgiarlo: I wonder if `Solrizer::Descriptor.new(:integer, :stored)` should be a constant defined in the `GenericFileIndexingService` class, so that `stored_integer_descriptor` in the collection_behavior above can point to that constant (i.e., use `Sufia::GenericFileIndexingService::STORED_INTEGER_DESCRIPTOR`) instead of calling `Descriptor.new` again.

spec/models/collection_spec.rb
narogers: That makes sense since we aren't actually testing ActiveFedora directly here.

spec/models/collection_spec.rb
narogers: The ID is needed because it is used in the bytes method. If you don't supply an ID then the {!join ...} Solr statement fails at https://github.com/narogers/sufia/blob/collection-speed/sufia-models/app/models/concerns/sufia/collection_behavior.rb#L23

mjgiarlo: :ok:

sufia-models/app/services/sufia/generic_file_indexing_service.rb
narogers: The semi colon has been eradicated. Regarding the constant is your concern over performance or just aesthetics? Until STORED_INTEGER_DESCRIPTOR is used in other places the performance difference is neglible as can be seen if you run the following script in a Pry console

@constant = Solrizer::Descriptor.new(:integer, :stored)
def createNewDescriptorWithConstant
	Solrizer.solr_name('file_size', @constant)
end

def createNewDescriptorDynamically
	Solrizer.solr_name('file_size', Solrizer::Descriptor.new(:integer, :stored))
end

Benchmark.bmbm do |x|
	x.report("Constant : ") { 500.times { createNewDescriptorWithConstant }}
	x.report("Dynamic : ") { 500.times { createNewDescriptorDynamically }}
end

mjgiarlo: ----- Original Message -----
> From: "Nathan Rogers" <notifications@github.com>
> To: "projecthydra/sufia" <sufia@noreply.github.com>
> Cc: "Michael J. Giarlo" <mjg36@psu.edu>
> Sent: Thursday, June 18, 2015 8:43:28 AM
> Subject: Re: [sufia] Support collection size calculations using a Solr query instead of lo… (#1204)
> 
> > @@ -6,6 +6,7 @@ def generate_solr_document
> >          solr_doc[Solrizer.solr_name('file_format')] = object.file_format
> >          solr_doc[Solrizer.solr_name('file_format', :facetable)] =
> >          object.file_format
> >          solr_doc['all_text_timv'] = object.full_text.content
> > +        solr_doc[Solrizer.solr_name('file_size',
> > Solrizer::Descriptor.new(:integer, :stored);)] = object.content.size.to_i
> 
> The semi colon has been eradicated. Regarding the constant is your concern
> over performance or just aesthetics? Until STORED_INTEGER_DESCRIPTOR is used
> in other places the performance difference is neglible as can be seen if you
> run the following script in a Pry console

Not performance or aesthetics, but maintainability -- since both are relying on the same descriptor, they should share a definition so that it can be updated in one place when we're ready to change it.

narogers:
I just added a patch for two failing tests in the Presenter class that were also related to a missing collection id. This should mean that all tests pass now

https://github.com/narogers/sufia/blob/collection-speed/spec/presenters/sufia/collection_presenter_spec.rb#L11

sufia-models/app/models/concerns/sufia/collection_behavior.rb
```diff
@@ -15,10 +15,37 @@ def update_permissions
...
+    protected
+      # Field to look up when locating the size of each file in Solr.
+      # Override for your own installation if using something different
+      def file_size_field
+        Solrizer.solr_name('file_size', stored_integer_descriptor)
+      end
+
+      # Override if you are storing your file size in a different way
+      def stored_integer_descriptor
+        Solrizer::Descriptor.new(:integer, :stored)
```
mjgiarlo: This should point at `Sufia::GenericFileIndexingService::STORED_INTEGER` now, right?

narogers: Nice catch. I'll get that updated Friday.

narogers:
Here is a comparison of the original code against the new version running on an i3 laptop.

https://gist.github.com/narogers/2687608229cfc63b7c2d

mjgiarlo:
:clap:

PR #1209, opened 2015-06-17
Stop supporting Rails 4.1

jcoyne:


awead:
Ideally, it would be nice to support at least the version just previous to the current Rails version, but we just don't have the cycles to support this. So I'm :+1:

jcoyne:
Yeah, I don't really want to do this, but I *really* don't want to spend time tracking down the build issue.

mjgiarlo:
Let me ping hydra-tech and see how much folks care about this.

mjgiarlo:
The rationale for this PR is described here:

https://groups.google.com/d/msg/hydra-tech/GBEh6lLNPcY/oH_A3CpZPUYJ

And that thread will help us determine whether or not to merge this PR.

mtribone:
Can we push the button? Can we? Can we? Okay. I'd like to hear from the community about this because it is a blocker.

cam156:
@mtribone Lets give this one a few days.  I'm :+1:  for pushing it.

mjgiarlo:
I've only seen one response to my message so far... giving this a few days seems fair to me.

jcoyne:
I don't think there's any need to wait on this. We've already dropped the ball on 4.1 (the build isn't passing). If someone wants to add 4.1 support in, I will be happy to merge that.

mjgiarlo:
No one chimed in to say they cared, so I'm :+1: on merging this.

awead:
Let the rebasing commence!

PR #1215, opened 2015-06-30
Fixes HTML source order, adds ARIA roles, and skip to content link …

mtribone:
…for better accessibility.

I've swapped the order of the sidebar and main content area, adjusted some of the headings, added some ARIA roles, so that we have better accessibility compliance and remove a few errors. The "skip to content" link gives users the ability to jump over the banner and main navigation, so that they don't have to listen to each element being read by a screen reader if they just want to get to the content. The headings H1-H6 in proper order help users who use assistive devices navigate and understand the hierarchy of the content.


awead:
:+1: Looks good to me.

PR #1218, opened 2015-06-30
Fixes accessibility issues with labels and form elements …

mtribone:
…in add to collection modal.

Fixed the "for" label so that it matches the ID of the radio for users of assistive devices because they need to match for users to understand which labelsd are for which controls. Also added a little more readability for users by adding the fieldset and legend to bring the code and content for the modal into context.


mjgiarlo:
:+1:

PR #1226, opened 2015-07-02
Making the image thumbnail alt tag empty for better accessibility, …

mtribone:
…since there is more information about title and description in the result.

The alt text should indicate the content or the function of the image rather than repeating the title. Rather than having duplicate information, it is preferred to leave the alt attribute empty with one space.


mjgiarlo:
:+1:  looks OK to me.

PR #1227, opened 2015-07-03
Refactoring feature test to make it more efficient

cam156:
I went through 2 steps with the refactoring checking timing at each step.  

The first step was just breaking the tests into different contexts that needed different numbers of  fixtures: zero, one and three.

The second step was to combine the specs that needed the same number of fixtures into one test instead of separate tests.  I realize this step may be controversial, but the speed improved measurably.

## original
  * real	0m42.671s
  * user	0m32.737s
  * sys	0m4.566s

## only getting fixtures needed
  * real	0m35.405s
  * user	0m27.328s
  * sys	0m4.197s

## Combing tests
  * real	0m23.604s
  * user	0m18.438s
  * sys	0m3.479s


jcoyne:
Nice! :+1:

PR #1228, opened 2015-07-03
Don't calculate coverage of code found in spec/ directory

mjgiarlo:


cam156:
:+1:

mjgiarlo:
:ok: This is a real PR now. Our past coverage numbers were kind of a lie.  Ping @cam156 @jcoyne etc.

PR #1229, opened 2015-07-06
Adding recent users query end date so a range can be queried. fixes #1187

### LINKED ISSUE #1187: Admin Stats: New User Range
As an administrator, I would like to have the ability to determine how many new users joined Sufia between any two dates.

I can set date parameters and see the total number of new users and who they are in Sufia for that date range

cam156:
Additionally moved the recent_user query to the User model so it could be tested more efficiently

### DIFF
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -172,4 +172,31 @@ describe User, :type => :model do
       expect(another_user.can_receive_deposits_from.to_a).to eq [@subject]
     end
   end
+  describe "class methods" do
+    describe "recent_users" do
+      let(:new_users) { User.all.order(created_at: :desc) }
+
+      before do
+        (1..3).each {|i| User.create(email: "abc#{i}@blah.frg", password: "blarg1234", created_at:DateTime.now - i.days) }
+      end
+
+      context "when has a start date" do
+        subject { User.recent_users(Date.today - 2.days) }
+        it "returns valid data" do
+          expect(subject.count).to eq 2
+          is_expected.to include(new_users[0],new_users[1])
+          is_expected.not_to include(new_users[2])
+        end
+      end
+
+      context "when has start and end date" do
+        subject { User.recent_users(Date.today - 2.days, Date.today - 1.days) }
+        it "returns valid data" do
+          expect(subject.count).to eq 1
+          is_expected.to include(new_users[1])
+          is_expected.not_to include(new_users[2], new_users[0])
+        end
+      end
+    end
+  end
 end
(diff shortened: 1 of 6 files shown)


spec/models/user_spec.rb
```diff
@@ -172,4 +172,27 @@
...
+  describe "class methods" do
+    describe "recent_users" do
+      let(:new_users) { User.all.order(created_at: :desc) }
+
+      before do
+        (1..3).each {|i| User.create(email: "abc#{i}@blah.frg", password: "blarg1234", created_at:DateTime.now - i.days) }
+      end
+
+      it "returns based on start date" do
+        results = User.recent_users(Date.today - 2.days)
```
awead: make this your subject.

spec/models/user_spec.rb
```diff
@@ -172,4 +172,27 @@
...
+        expect(results.count).to eq 2
+        expect(results).to include(new_users[0],new_users[1])
+        expect(results).not_to include(new_users[2])
+      end
+
+      it "returns based on start date and end date" do
+        results = User.recent_users(Date.today - 2.days, Date.today - 1.days)
```
awead: make this your subject.

spec/models/user_spec.rb
cam156: @awead Really?  I would then need another context to set that as a subject.  Is that more clear?

spec/models/user_spec.rb
cam156: @awead I would need an additional context to set that as the subject.  Is that really more clear?

spec/models/user_spec.rb
jcoyne: cam156: Yeah, the context is "calling with one argument", that is more clear to me. It describe what changes.

spec/models/user_spec.rb
jcoyne: This one is "when called with two arguments"

spec/models/user_spec.rb
awead: I think it can be simplified to:
``` ruby
context "with a start date"
  subject { User.recent_users(Date.today - 2.days) }
  it { is_expected.to contain_exactly(new_users[0],new_users[1]) }
end
```

awead: I'm looking at https://www.relishapp.com/rspec/rspec-expectations/docs/built-in-matchers for the `contains_exactly` matcher which seems to support all three assertions you want to me.

cam156:
@awead I set the context as requested.

awead:
@cam156 Thanks. I also discovered `contain_exactly` matcher which would be good for us to use in the future. I'll merge this as soon as it's green.

PR #1231, opened 2015-07-08
Make my controller retrieve only user's own collections.

ojlyytinen:
My controller used to get all collections from the system rather than just the ones belonging to the user. When adding files to a collection, this would give the user a list of all collections to choose from, not just user's own collections. The user couldn't add the files to someone else's collection anyway so those options were useless and confusing and the list could get potentially get very long.

### DIFF
--- a/spec/controllers/my/files_controller_spec.rb
+++ b/spec/controllers/my/files_controller_spec.rb
@@ -3,9 +3,14 @@ require 'spec_helper'
 describe My::FilesController, :type => :controller do
 
   let(:my_collection) do
-    Collection.new(title: 'test collection').tap do |c|
+    Collection.create(title: 'test collection') do |c|
       c.apply_depositor_metadata(user.user_key)
-      c.save!
+    end
+  end
+
+  let(:other_collection) do
+    Collection.create(title: 'other test collection') do |c|
+      c.apply_depositor_metadata(another_user.user_key)
     end
   end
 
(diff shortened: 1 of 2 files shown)


spec/controllers/my/files_controller_spec.rb
```diff
@@ -9,6 +9,13 @@
     end
   end
 
+  let(:other_collection) do
+    Collection.new(title: 'other test collection').tap do |c|
```
mjgiarlo: Simplify to:

```ruby
Collection.create(title: 'other test collection') do |c|
  c.apply_depositor_metadata(another_user.user_key)
end
```

cam156:
:+1:  We currently have this override in scholar sphere:https://github.com/psu-stewardship/scholarsphere/blob/develop/app/controllers/my_controller.rb#L4

Once the spec is fixed up it is good to go...

mjgiarlo:
@ojlyytinen If you can make the spec fix (replacing the new.tap pattern with create) and then squash your commits, I'd be happy to merge this today.

spec/controllers/my/files_controller_spec.rb
ojlyytinen: This doesn't seem to work. A collection is not valid until the depositor metadata has been applied. So the create alone will not save the collection and it'll still need a save in the tap.

jcoyne: The block passed to create is run before it's saved.  It's equivalent to:

```ruby
c = Collection.new(title: 'other test collection')
c.apply_depositor_metadata(another_user.user_key)
c.save
```

ojlyytinen: Oh right, I misread and had the .tap there still which made it behave the other way. I'll fix this in a minute.

ojlyytinen:
I replaced the tap as suggested and rebased.

mjgiarlo:
:+1:

PR #1232, opened 2015-07-08
Allowing the second input to also auto complete.  fixes #1222

### LINKED ISSUE #1222: Auto Completes on GenericFile edit do not work on second field
When you Edit a File
And you type in the language Field
Then you should get an autocomplete
When you Type in the second Language field
The you should also get an auto complete

First field has auto complete
screen shot 2015-07-01 at 10 00 02 am

second field does not 
screen shot 2015-07-01 at 10 00 15 am

cam156:


mjgiarlo:
:+1: if Travis is :green_heart:

cam156:
@mjgiarlo it passed

cam156:
@mjgiarlo it passed

PR #1233, opened 2015-07-08
Updated to support ActiveFedora 9.2.0

jcoyne:

PR #1234, opened 2015-07-08
Removing any calls to the sort! method

cam156:


mjgiarlo:
Does this supersede #1233?

cam156:
@mjgiarlo it could.  They are both solving the same problem.

mjgiarlo:
It looks like this PR covers everything #1233 covers and then some. I'm inclined to merge this and close the other.

PR #1256, opened 2015-07-16
Adding object type and facet count to get all the files deposited by …

cam156:
…all users, not just all the objects deposited by the top 10 users

### DIFF
--- a/spec/controllers/admin_stats_controller_spec.rb
+++ b/spec/controllers/admin_stats_controller_spec.rb
@@ -113,6 +113,10 @@ describe Admin::StatsController, type: :controller do
           gf.apply_depositor_metadata(user1)
           gf.update_index
         end
+        Collection.new(id: "ccc123") do |c|
+          c.apply_depositor_metadata(user1)
+          c.update_index
+        end
       end
 
       it "gathers user deposits" do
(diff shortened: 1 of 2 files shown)


spec/controllers/admin_stats_controller_spec.rb
```diff
@@ -124,6 +128,26 @@
...
+            GenericFile.new(id: "more#{number}") do |gf|
+              gf.apply_depositor_metadata(luser)
+              gf.update_index
+            end
+          end
+        end
+
+        it "gathers user deposits" do
+          get :index
+          expect(assigns[:depositors]).to include({ key: user1.user_key, deposits: 2, user: user1 }, { key: user2.user_key, deposits: 1, user: user2 })
```
mjgiarlo: Why does user1 have 2 deposits?

cam156: There were two files added here:  https://github.com/psu-stewardship/sufia/blob/depositor_files/spec/controllers/admin_stats_controller_spec.rb#L103-L115

mjgiarlo: @cam156 Ah, I see. And `user1` is not one of the 12 created users. Those are tested on the line below. Gotcha!

PR #1257, opened 2015-07-16
Adding an announcement editable content block to the home page

cam156:
Non Admin User with no Announcement looks the same as it used to:
screen shot 2015-07-16 at 12 40 59 pm

Non Admin user with an announcement:
screen shot 2015-07-16 at 12 40 22 pm

Admin user with an announcement:
screen shot 2015-07-16 at 12 40 34 pm

### DIFF
--- a/spec/controllers/homepage_controller_spec.rb
+++ b/spec/controllers/homepage_controller_spec.rb
@@ -99,5 +99,15 @@ describe HomepageController, :type => :controller do
         expect(assigns(:featured_work_list)).to be_kind_of FeaturedWorkList
       end
     end
+
+    it "sets announcement content block" do
+      get :index
+      expect(response).to be_success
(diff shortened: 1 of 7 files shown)


spec/controllers/homepage_controller_spec.rb
```diff
@@ -99,5 +99,15 @@
         expect(assigns(:featured_work_list)).to be_kind_of FeaturedWorkList
       end
     end
+
+    it "sets announcement content block" do
+      get :index
+      expect(response).to be_success
+      assigns(:announcement_text).tap do |announcement|
```
mjgiarlo: Neat use of `#tap`!

grosscol:
Looks good to me.

mjgiarlo:
:+1:

PR #1258, opened 2015-07-17
Refactor the controller test to assume the ingest job is already tested

carolyncole:


mjgiarlo:
Looks good to me.

Thoughts, @jcoyne ?

PR #1259, opened 2015-07-17
Fixing audit service since audit no longer exists on the GenericFile

cam156:


spec/services/repository_audit_service_spec.rb
```diff
@@ -11,7 +11,11 @@
 
   describe "#audit_everything" do
     it "should audit everything" do
-      expect_any_instance_of(GenericFile).to receive(:audit)
+      # check that the method we are mocking still exists
+      expect(Sufia::GenericFileAuditService.new(nil).respond_to?(:audit)).to be_truthy
```
mjgiarlo: Can this use a more specific matcher?

```ruby
expect(Sufia::GenericFileAuditService.new(nil)).to respond_to(:audit)
```

mjgiarlo:
One small change, and then I'm :+1:

spec/services/repository_audit_service_spec.rb
cam156: Good call!  I'll fix it up...

PR #1260, opened 2015-07-17
Reload the object in the import url job after loading in the file inca…

carolyncole:
…se another job has modified the object in the while the data was being copied


jcoyne:
:+1:

mjgiarlo:
:+1:

PR #1262, opened 2015-07-21
updated link_to_field in sufia_helper_behavior

### LINKED ISSUE #1263: Link to non-string objects in the item page
We would like to add the ability to display attributes such as URI labels in the item page, so we plan to add the to_s in the link_to_field method so that we can print the string part successfully.

luisgreg99:
Updated link_to_field in sufia_helper_behavior so that it can receive other fieldvalue objects such as uri fields. Fixes #1263

### DIFF
--- a/spec/helpers/sufia_helper_spec.rb
+++ b/spec/helpers/sufia_helper_spec.rb
@@ -28,11 +28,21 @@ describe SufiaHelper, :type => :helper do
   end
 
   describe "#link_to_field" do
+    context "when fieldvalue is a string" do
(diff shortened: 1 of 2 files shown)


spec/helpers/sufia_helper_spec.rb
```diff
@@ -27,6 +27,14 @@
     end
   end
 
+  describe "#link_to_field" do
```
terrellt: Could you add a test to make sure that the #to_s interface is used? Maybe pass an RDF::URI as contributor instead of a string?

PR #1264, opened 2015-07-24
Enforce styles automatically so we don't waste each other's time

mjgiarlo:
Rubocop style checking now occurs automatically when the `rake spec` task is run. Travis-CI will also run Rubocop before RSpec and will fail on any style issues (fast) so that we can pinpoint style-related issues in advance of waiting on the whole test suite to finish.

The included Rubocop config file matches our community's styles as best I could make it. Criticism and questions are welcome -- I'd like this to be a style guide we (Sufia developers) can all live with, so the PR process is less time-consuming, less frustrating, and less subjective. 

The style checkers, or "cops," defined in .rubocop.yml that have specific files and directories excluded are ones that raise dozens of warnings and errors, and would require a significant alteration to the codebase (e.g., replacing all instance variables in our test suite with `let` statements). Cops that are disabled altogether are disabled for the same reason. This means that some style-related issues will fall between the cracks until we take the time to bring these files up to par to allow more cops to run, but on the whole I'm confident this PR will reduce chatter about whitespace and parentheses (and so on) in our pull requests.

PR #1265, opened 2015-07-24
User mailbox

hectorcorrea:
After a week of training on "Practical Object-Oriented Design" with Sandi Metz @cam156 @awead and I decided to take a stab at refactoring a small piece of functionality with some of the ideas that we learned.

This PR does not change the functionality of the `MailboxController` but it refactors most of the code out of it to a model class `UserMailbox` This new class encapsulates most of the functionality and the controller is now as skinny as it gets which was one of our goals.

The new `UserMailbox` has a more comprehensive list of tests than the original code and it's easier to test since it does not mix model + controller functionality.

### DIFF
--- a/app/controllers/mailbox_controller.rb
+++ b/app/controllers/mailbox_controller.rb
@@ -2,49 +2,23 @@ class MailboxController < ApplicationController
   before_action :authenticate_user!
 
   def index
-    if user_signed_in?
-      @messages = current_user.mailbox.inbox
-      @messages.each { |m| m.mark_as_read(current_user) }
-    else
-      @messages = []
-    end
+    @messages = user_mailbox.inbox
   end
 
   def delete_all
-    current_user.mailbox.inbox.each do |msg|
-      delete_message(msg)
-    end
-    empty_trash(current_user)
-    redirect_to sufia.notifications_path
+    alert = user_mailbox.delete_all
+    redirect_to sufia.notifications_path, alert: alert
   end
 
   def destroy
-    if current_user
-      msg = Mailboxer::Conversation.find(params[:id])
-      if (msg.participants[0] == current_user) || (msg.participants[1] == current_user)
-        delete_message(msg)
-        empty_trash(msg.participants[0])
-      end
-    else
-      flash[:alert] = "You do not have privileges to delete the notification..."
-    end
-    redirect_to sufia.notifications_path
+    message_id = params[:id]
+    alert = user_mailbox.destroy(message_id)
+    redirect_to sufia.notifications_path, alert: alert
   end
 
   private
 
-    def delete_message(msg)
-      msg.move_to_trash(msg.participants[0])
-      msg.move_to_trash(msg.participants[1])
-    end
-
-    def empty_trash(user)
-      user.mailbox.trash.each do |conv|
-        conv.messages.each do |notify|
-          notify.receipts.each(&:delete)
-          notify.delete
-        end
-        conv.delete
-      end
+    def user_mailbox
+      UserMailbox.new(current_user)
     end
 end
--- a/app/models/user_mailbox.rb
+++ b/app/models/user_mailbox.rb
@@ -1,0 +1,44 @@
+class UserMailbox
+  attr_reader :user
+
+  def initialize(user)
+    @user = user
+  end
+
+  def inbox
+    messages = user.mailbox.inbox
+    messages.each { |m| m.mark_as_read(user) }
+  end
+
+  def delete_all
+    user.mailbox.inbox.each do |msg|
+      delete_message(msg)
+    end
+    empty_trash(user)
+  end
+
+  def destroy(message_id)
+    msg = Mailboxer::Conversation.find(message_id)
+    return "You do not have privileges to delete the notification..." unless msg.participants.include? user
+    delete_message(msg)
+    empty_trash(msg.participants[0])
+    nil
+  end
+
+  private
+
+    def delete_message(msg)
+      msg.move_to_trash(msg.participants[0])
+      msg.move_to_trash(msg.participants[1])
+    end
+
+    def empty_trash(user)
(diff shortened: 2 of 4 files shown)


mjgiarlo:
Thanks, @skmetz!

app/controllers/mailbox_controller.rb
```diff
@@ -2,43 +2,17 @@ class MailboxController < ApplicationController
   before_filter :authenticate_user!
 
   def index
-    if user_signed_in?
-      @messages = current_user.mailbox.inbox
-      @messages.each{|m| m.mark_as_read(current_user)}
-    else
-      @messages =[]
-    end 
+    @messages = UserMailbox.new(current_user).inbox
```
jcoyne: how about creating  ?
```ruby
private
def user_mailbox
  UserMailbox.new(current_user)
end
```

app/models/user_mailbox.rb
```diff
@@ -0,0 +1,44 @@
+class UserMailbox
+  attr_reader :user
+
+  def initialize(user)
+    raise ArgumentError.new("Must provide a user") if user == nil
```
jcoyne: This feels overly paranoid to me. Especially since we never catch this error. If I kept it I'd write it as: `unless user`

jcoyne:
This is a big improvement. Just 2 smallish things to consider.

app/controllers/mailbox_controller.rb
mjgiarlo: :+1: to isolating this in a private method, so it only has to change in one place

app/models/user_mailbox.rb
mjgiarlo: Not catching this in the codebase raises my hackles too, because that means if it triggers, users will see an error message (and one that means nothing to them).

Under what conditions will an instance of this class get created with a nil user? What situation are we trying to prevent?

mjgiarlo:
Can you squash your commits?

app/models/user_mailbox.rb
awead: Yeah, this caught my eye too. You should always have a user, so rely on Ruby's failure message if the argument isn't supplied.

app/controllers/mailbox_controller.rb
hectorcorrea: Done.

app/models/user_mailbox.rb
hectorcorrea: Since the three of you seem to be against the guard I went ahead and removed it. 

@mjgiarlo the `user` parameter will come from the controller which is why I was trying to make sure it wasn't `nil`. As a developer I would rather see an error that indicates that "no user was passed" than "nil didn't have property xyz" (and then realize oh... the user must have been nil)

app/models/user_mailbox.rb
```diff
@@ -0,0 +1,43 @@
+class UserMailbox
+  attr_reader :user
+
+  def initialize(user)
+    @user = user
+  end
+
+  def inbox
+    messages = user.mailbox.inbox
+    messages.each{|m| m.mark_as_read(user)}
```
mjgiarlo: Rubocop (#1264) would say do this instead:

```ruby
messages.each { |m| m.mark_as_read(user) }
```

app/models/user_mailbox.rb
```diff
@@ -0,0 +1,43 @@
...
+  end
+
+  def delete_all
+    user.mailbox.inbox.each do |msg|
+      delete_message(msg)
+    end
+    empty_trash(user)
+  end
+
+  def destroy message_id
```
mjgiarlo: Always surround params with parentheses in method definitions. (Parroting Rubocop.)

app/models/user_mailbox.rb
```diff
@@ -0,0 +1,43 @@
...
+    error_msg = nil
+    msg = Mailboxer::Conversation.find(message_id)
+    if (msg.participants[0] == user) || (msg.participants[1] == user)
```
mjgiarlo: Rubocop says these parentheses are unnecessary.  Could you refactor this as:

```ruby
return 'You do not have privileges to delete the notification' unless msg.participants.include? user
delete_message(msg)
empty_trash(msg.participants[0])
nil
```

app/models/user_mailbox.rb
```diff
@@ -0,0 +1,43 @@
...
+      empty_trash(msg.participants[0])
+    else
+      error_msg = "You do not have privileges to delete the notification..."
+    end
+    error_msg
+  end
+
+  private
+
+  def delete_message (msg)
```
mjgiarlo: Rubocop says: no spaces between a method name and its parenthesized params in a method definition.

app/models/user_mailbox.rb
```diff
@@ -0,0 +1,43 @@
...
+    msg.move_to_trash(msg.participants[0])
+    msg.move_to_trash(msg.participants[1])
+  end
+
+  def empty_trash (user)
```
mjgiarlo: Rubocop says: no spaces between a method name and its parenthesized params in a method definition.

app/models/user_mailbox.rb
```diff
@@ -0,0 +1,43 @@
...
+    user.mailbox.trash.each { |conv| conv.messages.each {|notify| notify.receipts.each { |receipt| receipt.delete}; notify.delete}; conv.delete}
```
mjgiarlo: Can you break this into multiple lines?

```ruby
user.mailbox.trash.each do |conv|
  conv.messages.each do |notify|
    notify.receipts.each { |receipt| receipt.delete}
    notify.delete
  end
  conv.delete
end
```

hectorcorrea:
I've implemented most of the round two of suggestions and rebased it with the new "Rubocop approved" master.

mjgiarlo:
:+1: :shipit:

PR #1278, opened 2015-08-11
Fix problems with single use download links.

ojlyytinen:
Fixes two problems that caused single use download links not to work.

First problem was caused by Turbolinks attempting to load the file asynchronously before the browser did a proper request. The first request by Turbolinks invalidated the single use link and then the browser failed to download the file. This is fixed by adding data-no-turbolink in the link attributes, which disabled Turbolinks for this link.

The second problem was also caused by multiple requests for the same single use link. At least in Chrome, trying to download a PDF file causes the browser to first issue a request, then it opens a PDF viewer (presumably after seeing the content-type in the response headers) which issues another request for the file. The second request will fail because the link is invalid at this point. This is fixed by instructing the browser to save the file with a Content-Disposition: attachment header.

### DIFF
--- a/app/controllers/concerns/sufia/single_use_links_viewer_controller_behavior.rb
+++ b/app/controllers/concerns/sufia/single_use_links_viewer_controller_behavior.rb
@@ -31,6 +31,12 @@ module Sufia
 
     protected
 
+      def content_options
+        super.tap do |options|
+          options[:disposition] = 'attachment' if action_name == 'download'
+        end
+      end
+
       def presenter
         presenter_class.new(@asset)
       end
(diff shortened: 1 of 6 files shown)


jcoyne:
Are you sure about turbolinks? It's not a preloader. So it shouldn't trigger any links you don't click on.

ojlyytinen:
No it doesn't preload anything, but when you do click the link, Turbolinks makes its own request and then the browser itself makes another. I'm not sure why this happens but I'm guessing that Turbolinks sees that it's a binary file instead of html so then it just gives it to the browser.

jcoyne:
That does make sense.

mjgiarlo:
:+1:

PR #1279, opened 2015-08-13
Fix sorting issues with collections.

### LINKED ISSUE #1272: Ordering of collection lists
I was trying to figure out why the ordering of collections in various places is different and discovered a few issues with how collections are ordered. 

The My Collections page seems to order collections in Solr using date_uploaded in descending order. However, date_uploaded is never set for collections, thus the ordering here is unspecified but seems to be based on creation time in ascending order. Hydra::Collections::Actions has the before filters to set date_uploaded, but this file is deprecated and not used anywhere. I wonder what the history here is, has this file been used before and why has it been deprecated?

The My Collections page has a column for Date Uploaded, but this is always blank because it's never set for collections. Either the column should be removed or the date uploaded field used in collections.

The header of that column is clickable and will toggle between ascending and descending ordering, but the link for this takes you to My Files, even when clicked on My Collections page. Furthermore, the form above the table that you can use to select ordering or page size always takes you to the My Files page as well. So you can't change any of that for My Collections page. This should be fixed regardless anything else.

In My Files page if you add files to a collection you get a list where you can select which collection to add the files to. This list seems to do the sorting in the view template (views/collections/_form_for_select_collection.html.erb) and uses ascending order of date_modified. Date modified also is never set for collections. And even if it were, it would be confusing as this is a different sorting criterion compared to the My Collections page.

I was going to fix this in our version but if this all sounds good, I might as well make it a pull request upstream. As I see it, there are two options, enable the before filters that set date uploaded and modified for collections and then make everything use date uploaded consistently. Or don't set them and use something else for ordering (Solr has at least timestamp), but do it consistently. In this case the date uploaded column could be removed from My Collections page.

ojlyytinen:
Fixes #1272. There are a some potential issues with this pull request which should be considered before merging.

As discussed in the issue page, collections don't have date_uploaded or date_modified set so something else needs to be used for sorting. I changed sorting to use create_date and modified_date, which are indexed in Solr as system_create_dtsi and system_modified_dtsi. This change applies to both GenericFiles and Collections. One file that needed changing was the [catalog_controller generator template](https://github.com/projecthydra/sufia/blob/master/lib/generators/sufia/templates/catalog_controller.rb#L22-L28) where it's difficult to differentiate the behaviour between GenericFiles and Collections.

At first I thought that create_date and date_uploaded would be identical in GenericFiles so this wouldn't be a problem, but I then came across [this comment](https://github.com/projecthydra/sufia/blob/master/sufia-models/app/models/concerns/sufia/generic_file/metadata.rb#L51-L55) in metadata.rb

```ruby
# We reserve date_uploaded for the original creation date of the record.
# For example, when migrating data from a fedora3 repo to fedora4,
# fedora's system created date will reflect the date when the record
# was created in fedora4, but the date_uploaded will preserve the
# original creation date from the old repository.
```

So using the create_date instead of date_uploaded is less than ideal in cases where migration has been used at some point. (As a side note, not having the original creation date of collections preserved at all in migrations might be a problem in itself.)

Note that this PR also adds a small improvement that was not part of the original issue but is related. When clicking the Add Files link in a collection page, I set it to pass the id of the collection as a parameter which goes through to the collection list template so that the correct collection can be preselected there. The list used to have the checked attribute set for all options which resulted in the last collection in the list to be selected (at least on the browser I use). The user however might expect the collection where they started the operation at to be selected and if they don't pay attention, they could easily add the files to the wrong collection.

### DIFF
--- a/app/controllers/concerns/sufia/my_controller_behavior.rb
+++ b/app/controllers/concerns/sufia/my_controller_behavior.rb
@@ -52,6 +52,8 @@ module Sufia
       @batch_size_on_other_page = batch_size - count_on_page
       @batch_part_on_other_page = (@batch_size_on_other_page) > 0
 
+      @add_files_to_collection = params.fetch(:add_files_to_collection, '')
+
       respond_to do |format|
         format.html {}
         format.rss  { render layout: false }
--- a/app/models/concerns/sufia/solr_document_behavior.rb
+++ b/app/models/concerns/sufia/solr_document_behavior.rb
@@ -45,7 +45,17 @@ module Sufia
       return unless field.present?
       begin
         Date.parse(field).to_formatted_s(:standard)
-      rescue
+      rescue ArgumentError
+        ActiveFedora::Base.logger.info "Unable to parse date: #{field.first.inspect} for #{self['id']}"
+      end
+    end
+
+    def create_date
+      field = self[CatalogController.uploaded_field]
+      return unless field.present?
+      begin
+        Date.parse(field).to_formatted_s(:standard)
+      rescue ArgumentError
         ActiveFedora::Base.logger.info "Unable to parse date: #{field.first.inspect} for #{self['id']}"
       end
     end
--- a/app/views/collections/_form_for_select_collection.html.erb
+++ b/app/views/collections/_form_for_select_collection.html.erb
@@ -13,9 +13,10 @@
               <fieldset>
                 <legend><%= t("sufia.collection.select_form.select_heading") %></legend>
               <ul>
-                <% user_collections.sort { |c1,c2| c1['date_modified_dtsi'] <=> c2['date_modified_dtsi'] }.each do |collection|  %>
+                <% user_collections.sort { |c1,c2| c2[CatalogController.uploaded_field] <=> c1[CatalogController.uploaded_field] }.each_with_index do |collection,index|  %>
                   <li>
-                    <%= radio_button_tag(:id, collection.id, true, class: "collection-selector") %>
+                    <% selected = (collection.id == @add_files_to_collection) || (@add_files_to_collection.blank? && index == 0) %>
+                    <%= radio_button_tag(:id, collection.id, selected, class: "collection-selector") %>
                     <%= label_tag(:collection, collection.title, for: "id_#{collection.id}") %>
                   </li>
                 <% end %>
@@ -28,10 +29,10 @@
           <button type="button" class="btn btn-default" data-dismiss="modal"><%= t("sufia.collection.select_form.close")%></button>
           <% if user_collections.blank? %>
             <%= button_for_create_collection t("sufia.collection.select_form.create") %>
-          <% else %> 
+          <% else %>
             <%= button_for_update_collection t("sufia.collection.select_form.update") %>
             <%= button_for_create_collection t("sufia.collection.select_form.create_new") %>
-          <% end %> 
+          <% end %>
         </div>
       </div><!-- /.modal-content -->
   </div><!-- /.modal-dialog -->
--- a/spec/features/collection_spec.rb
+++ b/spec/features/collection_spec.rb
@@ -25,6 +25,15 @@ describe 'collection', type: :feature do
   let(:title2) { "Test Collection 2" }
   let(:description2) { "Description for collection 2 we are testing." }
 
+  let(:collection1) do
+    Collection.create(title: title1, description: description1,
+                      members: []) { |c| c.apply_depositor_metadata(user.user_key) }
+  end
+  let(:collection2) do
+    Collection.create(title: title2, description: description2,
+                      members: []) { |c| c.apply_depositor_metadata(user.user_key) }
+  end
+
   let(:user) { FactoryGirl.create(:user) }
 
   let(:gfs) do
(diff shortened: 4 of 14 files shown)


mjgiarlo:
I'll review this very soon!  Sorry for the wait.

hectorcorrea:
:+1: Nice job @ojlyytinen

ojlyytinen:
Does this look ok as it is then, despite the possible issues I mentioned? I'm mainly concerned about using create_date instead of date_uploaded in files where it seems like date_uploaded was specifically made to be preserved through migrations whereas create_date would be reset. So after migration, it would be impossible to sort files by their original creation date.

I can add a couple more tests just to have the coverage check pass.

mjgiarlo:
Since you offered, @ojlyytinen, sure, go ahead and toss a couple extra tests in. Because this is a thorough and well documented PR, I haven't had a chance to fully review it yet -- apologies that you've had to wait for a merge.  You have raised important issues in your comments on this PR that we'll want to consider and discuss here.  Hoping some other folks will chime in. @cam156 @jcoyne @awead ?

spec/features/collection_spec.rb
```diff
@@ -140,6 +140,82 @@ def create_collection(title, description)
...
+      expect(page).to have_css("#document_#{collection1.id}")
+      expect(page).to have_css("#document_#{collection2.id}")
+      expect(page.body.index("id=\"document_#{collection1.id}")).to be < page.body.index("id=\"document_#{collection2.id}")
+    end
+  end
+
+  describe 'add files to collection' do
+    let!(:collection1) do
+      Collection.create(title: title1, description: description1,
+                        members: []) { |c| c.apply_depositor_metadata(user.user_key) }
```
awead: This is the same as collection1 above. Can you just reuse the existing one?

spec/features/collection_spec.rb
```diff
@@ -140,6 +140,82 @@ def create_collection(title, description)
...
+    end
+    let!(:collection2) do
+      Collection.create(title: title2, description: description2,
+                        members: []) { |c| c.apply_depositor_metadata(user.user_key) }
```
awead: dito here. Reuse existing collection2?

spec/views/collections/_form_for_select_collection.html.erb_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  let(:doc) { Nokogiri::HTML(rendered) }
+
+  before do
+    allow(view).to receive(:user_collections).and_return(solr_collections)
+  end
+
+  it "sorts the collections" do
+    render
```
awead: try assigning render to a `let!` variable so it might only render the page once.

app/models/concerns/sufia/solr_document_behavior.rb
```diff
@@ -50,6 +50,16 @@ def date_uploaded
       end
     end
 
+    def create_date
+      field = self[Solrizer.solr_name("system_create", :stored_sortable, type: :date)]
+      return unless field.present?
+      begin
+        Date.parse(field).to_formatted_s(:standard)
+      rescue
```
awead: It might be better to rescue from the specific data parsing error, instead of just any error.

app/controllers/concerns/sufia/my_controller_behavior.rb
```diff
@@ -52,6 +52,8 @@ def index
       @batch_size_on_other_page = batch_size - count_on_page
       @batch_part_on_other_page = (@batch_size_on_other_page) > 0
 
+      @add_files_to_collection = params[:add_files_to_collection] || ''
```
awead: try using fetch here:
```
@add_files_to_collection = fetch(params[:add_files_to_collection], '')
```

app/views/collections/_form_for_select_collection.html.erb
```diff
@@ -13,9 +13,10 @@
               <fieldset>
                 <legend><%= t("sufia.collection.select_form.select_heading") %></legend>
               <ul>
-                <% user_collections.sort { |c1,c2| c1['date_modified_dtsi'] <=> c2['date_modified_dtsi'] }.each do |collection|  %>
+                <% user_collections.sort { |c1,c2| c2['system_create_dtsi'] <=> c1['system_create_dtsi'] }.each_with_index do |collection,index|  %>
```
awead: using solr_name or `CatalogController.uploaded_field` here would avoid having to know _dtsi.

app/models/concerns/sufia/solr_document_behavior.rb
```diff
@@ -50,6 +50,16 @@ def date_uploaded
       end
     end
 
+    def create_date
+      field = self[Solrizer.solr_name("system_create", :stored_sortable, type: :date)]
```
awead: you could call `CatalogController.uploaded_field` instead.

spec/models/solr_document_spec.rb
```diff
@@ -10,6 +10,15 @@
...
 
+  describe "create_date" do
+    before do
+      subject['system_create_dtsi'] = '2013-03-14T00:00:00Z'
+    end
+    it "is a date" do
+      expect(subject.create_date).to eq '03/14/2013'
+    end
+  end
+
```
awead: might be good to test your rescue clause here too.

awead:
Looks good overall, @ojlyytinen. I added a few comments to shore things up a bit.

app/controllers/concerns/sufia/my_controller_behavior.rb
mjgiarlo: Typo? I think you meant:

```ruby
@add_files_to_collection = params.fetch(:add_files_to_collection, '')
```

awead: sorry, pre :coffee: comment.

mjgiarlo: happens!

mjgiarlo:
@awead @cam156 @hectorcorrea Are we concerned that this changes the date for sorting? (Using the F4 date rather than date_uploaded, which we used to carry forward F3 dates.)  If not, I'm :+1:.

awead:
@mjgiarlo if we always refer to the sorting field as `CatalogController.uploaded_field` then it ought be easy to change around.

mjgiarlo:
@awead so we'd deal with this in ScholarSphere (which this PR will break because we need the dates that reflect F3 values) by redefining `CatalogController.uploaded_field`?

awead:
@mjgiarlo well, since we've defined CatalogController.uploaded_field, Sufia would use whatever that field is.

mjgiarlo:
@awead I'm not following.  Once @ojlyytinen makes the requested tweaks (including using `CatalogController.uploaded_field`, if we merge this, you're saying this PR won't break ScholarSphere because we already have this in place for our F3 date needs? https://github.com/psu-stewardship/scholarsphere/blob/develop/app/controllers/catalog_controller.rb#L21-L27

If so, :+1:

awead:
@mjgiarlo it shouldn't, because Sufia will look to our controller's class variable to determine what the field is. At least that's how I'm thinking it will.

ojlyytinen:
I've made the changes, with one exception. I'm not quite sure what @awead meant with putting the render in a let! variable. I thought that let! variables get evaluated once per example anyway so it still wouldn't render the page only once. And besides, the middle example assigns one extra variable and needs its own rendering anyway. Or am I completely misunderstanding what you meant?

For the duplicate collection definitions, I moved them up a level so they don't need to be repeated. I also changed them to lazy evaluation so they won't be created for all examples. Their creation is then forced where needed by just referencing them.

All system_create_dtsi's should be changed to CatalogController.uploaded_field now. There were a few more in views/my/_index_partials/_default_group.html.erb

awead:
:+1:

ojlyytinen:
Pushed in one more update with a few extra examples in solr_document_spec to satisfy the coverage check.

mjgiarlo:
Thanks, @ojlyytinen, for the contribution and for your patience!

PR #1281, opened 2015-08-17
Added citation configuration to sufia. Allows for rendering citation link on show page

straleyb:
Citations are enabled in sufia. There is an ability that you can give to people which allows for viewing of these citations. The issue is that there is no way to activate a link to route to the citations page. Therefore I propose we add this snippet of code to sufia which allows for the configuration of being able to show these citations.

### DIFF
--- a/sufia-models/lib/generators/sufia/models/citation_config_generator.rb
+++ b/sufia-models/lib/generators/sufia/models/citation_config_generator.rb
@@ -1,0 +1,22 @@
+require 'rails/generators'
+
+class Sufia::Models::CitationConfigGenerator < Rails::Generators::Base
+  source_root File.expand_path('../templates', __FILE__)
+
+  desc """
+This Generator makes the following changes to your application:
+  1. Updates existing sufia.rb initializer to include a citation configuration
+       """
+
+  def banner
+    say_status("info", "ADDING CITATIONS OPTION TO SUFIA CONFIG", :blue)
+  end
+
+  def inject_config_initializer
+    inject_into_file 'config/initializers/sufia.rb', before: "# Where to store tempfiles, leave blank for the system temp directory (e.g. /tmp)" do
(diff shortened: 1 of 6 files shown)


mjgiarlo:
PR looks good, @straleyb.  I have a few requests.

1. Can you add commented-out lines to the sufia.rb template that is generated into new installs showing the default value (`false`), so folks know this is a value they can tweak? (Hereish, I'm thinking: https://github.com/projecthydra/sufia/blob/master/sufia-models/lib/generators/sufia/models/templates/config/sufia.rb#L94)
1. Can you also add a generator that patches the lines from the above bullet into existing sufia.rb files for folks who are upgrading from 6.3.0 to the version that includes your patch?
1. Could you squash your commits?

Thanks for the contribution!

straleyb:
@mjgiarlo Will do! Thanks for the quick response.

straleyb:
Im not too terribly familiar with generators. Any feedback will be generously accepted.

sufia-models/lib/generators/sufia/models/citation_config_generator.rb
```diff
@@ -0,0 +1,22 @@
...
+class Sufia::Models::CitationConfigGenerator < Rails::Generators::Base
+  source_root File.expand_path('../templates', __FILE__)
+
+  desc """
+This Generator makes the following changes to your application:
+  1. Updates existing sufia.rb initializer to include a citation configuration
+       """
+
+  def banner
+    say_status("info", "Patching sufia.rb", :blue)
```
mjgiarlo: Can you change the message string to "ADDING CITATIONS OPTION TO SUFIA CONFIG"?

mjgiarlo:
You nailed it, @straleyb!  I have one more small change to make the message string in the generator look like the other ones (CAPS LOCK IS CRUISE CONTROL FOR COOL), and then I'm happy to merge this.  Well done.

PR #1282, opened 2015-08-18
Fixes issue with FireFox displaying search results outside containing div

lfarrell:
In FireFox (40.0.2) search results appear well outside their containing div, switching overflow-x to auto fixes this behavior.

sufia_css

PR #1285, opened 2015-08-19
Removed Bootstrap media object DIVs for accessibility, redundant …

mtribone:
…links, and responsive layout.

The WAVE accessibility tool bar flagged the thumbnail image and resource title as redundant links, since they go to the same place. In order to get the thumbnail and title into the same hyperlink, I had to break apart the Bootstrap media object approach to organizing images and other elements. It makes the code a little cleaner, too.


mtribone:
Might want to wait for https://github.com/projecthydra/sufia/pull/1282 and rebase?

Issue #1284, opened 2015-08-18
Rake tasks are getting called twice


awead:
When running the included rake tasks from sufia-models_tasks.rake, the tasks are getting called twice. This probably as result of them being loaded more than once.

mjgiarlo:
I don't believe this is relevant now that sufia-core has been merged back. Closing.

PR #1286, opened 2015-08-19
Fixes rendering of files list outside containing div in FireFox v40 a…

kerchner:
…nd sufia 6.3.0, but introduces new issue where Select An Action dropdown renders beneath containing div.  Credit goes to @lfarrell.

### DIFF
--- a/app/assets/stylesheets/sufia/_file-listing.scss
+++ b/app/assets/stylesheets/sufia/_file-listing.scss
@@ -13,7 +13,7 @@ h4 .small {
 }
 
 .table-responsive {
-  overflow-x: visible;
+  overflow-x: auto;
 }
 
 .table thead th {


kerchner:
Note that credit for this fix goes to @lfarrell !

hectorcorrea:
:+1:

kerchner:
@awead The Ruby 2.1 Travis-CI build for this PR failed because many of the spec tests encountered "Address already in use - bind(2) for 127.0.0.1:51465".  Based on what you observed that led you to create #128, do you think the issues with the Travis build might be related to #1284?

awead:
Interesting, @kerchner. I restarted the build, just in case it was some kind of fluke, but honestly I don't know if the Travis failures are related to the rake issue or not. I opened the issue just to document the problem. I had looked into what might be the cause.

kerchner:
@awead looks like restarting the build did it - must have been one of those travis-ci timing things, which I've seen before.  Thanks for restarting the build!

PR #1288, opened 2015-08-21
Fix bug in total upload file size calculation.

hackmastera:
Prior to this fix, uploadAdded() temporarily computes but does not store
a new total file size. If this excedes the max, the 'cancel' button is
clicked for the file that's been added. This triggers uploadFail() which
subtracts the size of the file in question from the stored total,
incorrectly deflating the total value.

As a result, more files are permitted in the list than configuration
would suggest are allowed. When files are all around the same size, you
see that after a cutoff point every other file is permitted, while the
ones between are removed.

This commit writes the new total file size during uploadAdded so that
when the subtraction in uploadFail is performed it results in the correct
total.


mjgiarlo:
How does this look, @cam156 @jcoyne ?

jcoyne:
Looks good to me, but I defer to @cam156

mjgiarlo:
:+1: pending @cam156's review.

PR #1292, opened 2015-08-24
extracting a presenter from the admin stats controller behavior

cam156:
We noticed that we were setting 8 instance variables for the view.  We moved those instance variables into a class and modified the code to use that presenter.


jcoyne:
:heart_eyes:

jcoyne:
Looks good. waiting on the travis build.

mjgiarlo:
:+1:

PR #1293, opened 2015-08-25
Index Fedora digest to link NRSes on disk to findable objects.

mjgiarlo:
Also, provide a query method that knows how to convert a digest string into a URN.

This PR indexes the Fedora-generated SHA1 digest to create a linkage between files on disk (in fcrepo.binary-store-path) and objects in the repository. Penn State has a use case for this: our security folks run a personally identifiable information scan on the filesystem and see a file with SSNs. All we know at that point is the SHA1. This PR lets us look up the repository object containing the offending binary.

### DIFF
--- a/sufia-models/app/services/sufia/generic_file_indexing_service.rb
+++ b/sufia-models/app/services/sufia/generic_file_indexing_service.rb
@@ -9,7 +9,18 @@ module Sufia
         solr_doc[Solrizer.solr_name('file_format', :facetable)] = object.file_format
         solr_doc['all_text_timv'] = object.full_text.content
         solr_doc[Solrizer.solr_name('file_size', STORED_INTEGER)] = object.content.size.to_i
+        # Index the Fedora-generated SHA1 digest to create a linkage
+        # between files on disk (in fcrepo.binary-store-path) and objects
+        # in the repository.
+        solr_doc[Solrizer.solr_name('digest', :symbol)] = digest_from_content
       end
     end
+
+    private
+
+      def digest_from_content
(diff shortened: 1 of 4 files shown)


awead:
:shipit:

sufia-models/app/services/sufia/generic_file_indexing_service.rb
```diff
@@ -9,7 +9,18 @@ def generate_solr_document
...
+        # in the repository.
+        solr_doc[Solrizer.solr_name('digest', :symbol)] = digest_from_content
       end
     end
+
+    private
+
+      def digest_from_content
+        return unless object.content.has_content?
+        object.content.metadata.ldp_source.graph.query(predicate: RDF::Vocab::Fcrepo4.digest).first.object.to_s
```
jcoyne: Move `metadata.ldp_source.graph.query(predicate: RDF::Vocab::Fcrepo4.digest).first.object.to_s` to a property in ActiveFedora https://github.com/projecthydra/active_fedora/blob/master/lib/active_fedora/with_metadata/metadata_node.rb

jcoyne: Actually, isn't it already there? https://github.com/projecthydra/active_fedora/blob/d6784e23a7ad324e2d03a23c7c063fc888b89398/lib/active_fedora/file/attributes.rb#L17

awead: @jcoyne yeah, I forgot I put it there.

PR #1294, opened 2015-08-25
Fix Generic File metadata update button

phebal:
`f.submit` doesn't take a block in the current version of Rails.  This fixes it and makes the button match the one in `app/views/batch/edit.html.erb`. Sorry for an extremely small PR but I was fixing it on my end, and thought I'll contribute upstream instead.

### DIFF
--- a/app/views/generic_files/_descriptions.html.erb
+++ b/app/views/generic_files/_descriptions.html.erb
@@ -10,7 +10,7 @@
 
       </div><!-- /well -->
     <div>
-      <%= f.submit class: 'btn btn-primary', onclick: "confirmation_needed = false;", id: "upload_submit", name: "update_descriptions" do %>
+      <%= button_tag class: 'btn btn-primary btn-lg', id: "upload_submit", name: "update_descriptions" do %>
         <i class="glyphicon glyphicon-floppy-disk"></i> Save Descriptions
       <% end %>
     </div>
--- a/spec/views/generic_file/edit.html.erb_spec.rb
+++ b/spec/views/generic_file/edit.html.erb_spec.rb
@@ -1,6 +1,35 @@
 require 'spec_helper'
 
 describe 'generic_files/edit.html.erb', :no_clean do
+  let(:content) { double('content', mimeType: 'application/pdf') }
+
+  let(:generic_file) do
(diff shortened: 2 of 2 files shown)


mjgiarlo:
:+1: if Travis is green. Thoughts, others?

mjgiarlo:
^ @mtribone @cam156 @jcoyne @awead

app/views/generic_files/_descriptions.html.erb
```diff
@@ -10,7 +10,7 @@
 
       </div><!-- /well -->
     <div>
-      <%= f.submit class: 'btn btn-primary', onclick: "confirmation_needed = false;", id: "upload_submit", name: "update_descriptions" do %>
+      <%= button_tag type: 'submit', class: 'btn btn-primary btn-lg', onclick: "confirmation_needed = false;", id: "upload_submit", name: "update_descriptions" do %>
```
awead: You should be able to omit the `type: 'submit'` as that should be the default. Also, is confirmation_needed the same as the `confirm:` option? Otherwise this is fine as-is. Just my picky comments.

spec/views/generic_file/edit.html.erb_spec.rb
```diff
@@ -40,4 +40,34 @@
...
+      render
+      Capybara::Node::Simple.new(rendered)
+    end
+
+    it "onlies draw one resource_type multiselect" do
+      within('#descriptions_display') do
+        expect(page).to have_button('Save Descriptions')
+      end
+    end
+  end
```
awead: you shouldn't have to replicate the entire setup for the test. The above section does it for you. Can you just move lines 67-71 to below 41?

jcoyne:
I did some research. I don't think `FormHelper#submit` (https://github.com/rails/rails/blob/master/actionview/lib/action_view/helpers/form_helper.rb#L1849) ever took a block. Nice find @phebal!

mtribone:
Good to go if this addresses @awead 's feedback.

phebal:
@awead thanks for the feedback, if fixed the button and specs I hope.

mjgiarlo:
Thx, @phebal

Issue #1289, opened 2015-08-21
Misleading error message when max_total_file_size is exceeded


hackmastera:
The message reads:
"All files selected from " + first_file_after_max + " and after will not be uploaded because your total upload is too big. You may not upload more than " + max_total_file_size_str + " in one upload."

However, the code does continue to evaluate each subsequent file, and if there is a file that is small enough that it doesn't bring the total over that max, it will be allowed / uploaded.

I think this is probably not desired behavior, since users are likely to want to go back to first_file_after_max and do a big block multiselect to continue their batch upload.

I propose we change the behavior of the code to match the error message, as opposed to vice-versa.

mjgiarlo:
:+1: to that, @hackmastera 

Good find!

PR #1296, opened 2015-08-26
Fix projecthydra/sufia#1289: Stop adding new files

hackmastera:
once max_total_file_size is reached.


hackmastera:
(It's a bit of a hack but it was the simplest fix.)

mjgiarlo:
OK with this change, @cam156 ?

PR #1297, opened 2015-08-26
Remove unused class RoleMapper

jcoyne:


mjgiarlo:
This isn't going to affect ScholarSphere because we define (and test) the same class in our codebase, I believe.  :+1: if @cam156 is similarly sanguine.

cam156:
@jcoyne & @mjgiarlo seems ok to me.

PR #1299, opened 2015-08-27
Fix styling of the 'Sort and Per-page Bar'  for files and collections liststings in the dashboard

phebal:
I think this was broken when fixing the responsive-table for FF40. It now looks like this:
sufia_broken_my
This makes it look like it did before and moves the css for `sort-toggler` to where it was getting overwritten from. `padding-bottom` is set to 0 because `fieldset` added in 04c20aa63a0b81c6b7480c6221927e7f54f39602 comes with it's own margin at the bottom. Here's how it looks after the fix:
sufia-fixed

FWIW I'm not much of a CSS expert.

### DIFF
--- a/app/assets/stylesheets/sufia/_dashboard.scss
+++ b/app/assets/stylesheets/sufia/_dashboard.scss
@@ -106,8 +106,13 @@ div.heading-tile:hover .glyphicon {
   margin-bottom: 0;
 }
 
-.sort-toggle, .batch-toggle {
+.batch-toggle {
   padding: 1.5em 0 1.5em 1em;
+}
+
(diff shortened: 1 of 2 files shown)


awead:
I'm not seeing the problem. Was this is a responsive issue?

phebal:
@awead do you mean that the issues is not obvious from my screenshot or that you're not seeing it in your browser? I tried circling in the screenshot where there's a break in the border on the left and right and although hard to see color switches from white to some sort of beige in that break.
sufia_broken_2

mjgiarlo:
Looks good to me.  @mtribone ?

PR #1304, opened 2015-08-31
Allows the developer to configure the frequency of the messages...

hectorcorrea:
pinging the server to check for user notifications.

The main goal of this is to allow the developer to remove the clutter of (mostly) useless messages in order to see the rest of the information that Rails logs by default (e.g. the controller hit on each request, the parameters sent, et cetera.)

The default behavior is to ping the server every 30 seconds. By passing "?notification_seconds=120" the developer can request the ping to happen every 120 seconds instead.

### DIFF
--- a/app/assets/javascripts/notifications_check.js.erb
+++ b/app/assets/javascripts/notifications_check.js.erb
@@ -1,0 +1,46 @@
+Blacklight.onLoad(function() {
+
+  <% if Rails.env.development? %>
+
+    // During development allow the frequency of the notifications check to
+    // be overwritten via query string parameter notification_seconds to
+    // prevent cluttering the terminal with (mostly) meaninless messages.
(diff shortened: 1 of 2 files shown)


mjgiarlo:
Jasmine is sad. :rice:

hectorcorrea:
@mjgiarlo I fixed the issue with Jasmine but I am still getting a error that seems to be unrelated to my change. I also get the same tests to fail on my local master branch. Any ideas?

mjgiarlo:
Maybe you need to rebase on master to pick this up: https://github.com/projecthydra/sufia/commit/f7500e7e1bee1cb8d20c2c41635ae6aa9984a71b

hectorcorrea:
Travis, Jasmine, and Santa Claus are now happy with this PR.

PR #1307, opened 2015-09-01
Add ability to upload to collection during Batch Upload

elrayle:
This adds the ability to select a collection during batch upload for:
* My Computer
* Cloud Providers
* Network/Server Location

### DIFF
--- a/History.md
+++ b/History.md
@@ -1,5 +1,11 @@
 # History of Sufia releases
 
+## 6.3.1
+
+* Add configuration allowing select menu on batch upload to upload files to a collection. [E. Lynette Rayle]
+* Bump active-fedora version to 9.4 [E. Lynette Rayle]
+* Bump hydra-collections to 5.0.3 [E. Lynette Rayle]
+
 ## 6.3.0
 
 * Fix problems with single use download links. [Olli Lyytinen]
--- a/app/controllers/concerns/sufia/files_controller_behavior.rb
+++ b/app/controllers/concerns/sufia/files_controller_behavior.rb
@@ -9,6 +9,8 @@ module Sufia
     extend ActiveSupport::Concern
     extend Sufia::FilesController::UploadCompleteBehavior
     include Sufia::Breadcrumbs
+    include Hydra::Catalog
+    include Hydra::Collections::SelectsCollections
 
     included do
       include Hydra::Controller::ControllerBehavior
@@ -38,6 +40,9 @@ module Sufia
       before_action :authenticate_user!, except: [:show, :citation, :stats]
       before_action :has_access?, except: [:show]
       before_action :build_breadcrumbs, only: [:show, :edit, :stats]
+      before_action only: [:new] do
+        find_collections_with_edit_access(true, -1, t("sufia.upload.collection.select_default_option"))
+      end
       load_resource only: [:audit]
       load_and_authorize_resource except: [:index, :audit, :show]
 
@@ -227,7 +232,7 @@ module Sufia
 
         update_metadata_from_upload_screen
         actor.create_metadata(params[:batch_id])
-        if actor.create_content(file, file.original_filename, file_path, file.content_type)
+        if actor.create_content(file, file.original_filename, file_path, file.content_type, params[:collection])
           respond_to do |format|
             format.html do
               render 'jq_upload', formats: 'json', content_type: 'text/html'
--- a/app/views/generic_files/_browse_everything.html.erb
+++ b/app/views/generic_files/_browse_everything.html.erb
@@ -4,6 +4,9 @@
 </div>
 <div class="well">
   <%= form_tag(sufia.generic_files_path, id: 'browse_everything_form', method: 'post') do %>
+      <% if Sufia.config.upload_to_collection %>
+          <%= render partial: 'generic_files/upload/to_collection' %>
+      <% end %>
       <%= render partial: 'generic_files/upload/tos_checkbox' %>
       <%= hidden_field_tag(:batch_id, @batch_id) %>
       <%= button_tag(t('sufia.upload.browse_everything.browse_files_button'), type: 'button', class: 'btn btn-lg btn-success', id: "browse-btn",
--- a/app/views/generic_files/_generic_file.html.erb
+++ b/app/views/generic_files/_generic_file.html.erb
@@ -9,7 +9,7 @@
     </div>
     <div class="col-sm-9">
       <%= render 'shared/attributes', work: generic_file %>
-    </table>
+    </div>
   </div>
 </li>
 
--- a/app/views/generic_files/upload/_to_collection.html.erb
+++ b/app/views/generic_files/upload/_to_collection.html.erb
@@ -1,0 +1,5 @@
+<p>
+  <label><%= t("sufia.upload.collection.select_label") %></label>
+  <%= select_tag 'collection', options_from_collection_for_select(@user_collections, "id", "title") %>
+</p>
+<p>&nbsp;</p>
--- a/spec/controllers/generic_files_controller_spec.rb
+++ b/spec/controllers/generic_files_controller_spec.rb
@@ -15,6 +15,12 @@ describe GenericFilesController do
       let(:mock) { GenericFile.new(id: 'test123') }
       let(:batch) { Batch.create }
       let(:batch_id) { batch.id }
+      let(:collection) {
+        Collection.create(title: 'test collection') do |c|
+          c.apply_depositor_metadata(user.user_key)
+        end
+      }
+      let(:collection_id) { collection.id }
       let(:file) { fixture_file_upload('/world.png', 'image/png') }
 
       before do
@@ -42,7 +48,7 @@ describe GenericFilesController do
       context "when everything is perfect" do
         render_views
         it "spawns a content deposit event job" do
-          expect_any_instance_of(Sufia::GenericFile::Actor).to receive(:create_content).with(file, 'world.png', 'content', 'image/png').and_return(true)
+          expect_any_instance_of(Sufia::GenericFile::Actor).to receive(:create_content).with(file, 'world.png', 'content', 'image/png', nil).and_return(true)
           xhr :post, :create, files: [file], 'Filename' => 'The world', batch_id: batch_id, permission: { group: { public: 'read' } }, terms_of_service: '1'
           expect(response.body).to eq '[{"name":null,"size":null,"url":"/files/test123","thumbnail_url":"test123","delete_url":"deleteme","delete_type":"DELETE"}]'
           expect(flash[:error]).to be_nil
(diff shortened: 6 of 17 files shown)


elrayle:
One quirky thing in [files_controller_behavior.rb](https://github.com/projecthydra/sufia/blob/upload_to_collection/app/controllers/concerns/sufia/files_controller_behavior.rb) around line 65, I create a SolrDoc to hold the default value and stuff it into the SolrDocs holding the list of collections.  I am open to doing this differently if someone has a suggestion.

elrayle:
HOLD MERGE - looking into test failure

app/controllers/concerns/sufia/files_controller_behavior.rb
```diff
@@ -59,6 +61,9 @@ def load_resource_from_solr
 
     # routed to /files/new
     def new
+      find_collections_with_edit_access
```
mjgiarlo: What does this line do?

app/controllers/concerns/sufia/files_controller_behavior.rb
```diff
@@ -59,6 +61,9 @@ def load_resource_from_solr
...
+      default_option = SolrDocument.new(id: -1, title_tesim: t("sufia.upload.collection.select_default_option"))
```
mjgiarlo: I might like to see this code (lines 65-66) live in SelectsCollections in hydra-collections.

app/controllers/concerns/sufia/files_controller_behavior.rb
```diff
@@ -228,6 +233,7 @@ def process_file(file)
         update_metadata_from_upload_screen
         actor.create_metadata(params[:batch_id])
         if actor.create_content(file, file.original_filename, file_path, file.content_type)
+          add_file_to_collection
```
mjgiarlo: I think the actor should perform this step as part of the `#create_content` method.

app/views/generic_files/_browse_everything.html.erb
```diff
@@ -4,6 +4,7 @@
 </div>
 <div class="well">
   <%= form_tag(sufia.generic_files_path, id: 'browse_everything_form', method: 'post') do %>
+      <%= render partial: 'generic_files/upload/to_collection' %>
```
mjgiarlo: Can you make the UI changes dependent on a config value (with a default value of `false` until we're sure this is functionality that the majority of folks want turned on)?

app/views/generic_files/_generic_file.html.erb
```diff
@@ -9,7 +9,7 @@
     </div>
     <div class="col-sm-9">
       <%= render 'shared/attributes', work: generic_file %>
-    </table>
+    </div>
```
mjgiarlo: Good catch!

app/views/generic_files/upload/_to_collection.html.erb
```diff
@@ -0,0 +1,6 @@
+<p>
+  <%= @user_collections = [] if @user_collections.nil? %>
```
mjgiarlo: I'd like to see the controller in hydra-collections handle this.

spec/controllers/generic_files_controller_spec.rb
```diff
@@ -15,6 +15,13 @@
       let(:mock) { GenericFile.new(id: 'test123') }
       let(:batch) { Batch.create }
       let(:batch_id) { batch.id }
+      let(:collection) {
+        Collection.new(title: "test collection").tap do |c|
```
mjgiarlo: You don't need the `#tap` here because `#new` takes a block. You can also use `#create` here for fewer LOC:

```ruby
Collection.create(title: 'test collection') do |c|
  c.apply_depositor_metadata(user.user_key)
end
```

app/controllers/concerns/sufia/files_controller_behavior.rb
elrayle: Uses Hydra::Collections to get the list of collections the current user is allowed to update.

mjgiarlo: OK, so this hooks into SelectsCollections. Can you pop this into a `before_action` hook?

app/controllers/concerns/sufia/files_controller_behavior.rb
elrayle: find_collections_with_edit_access is from Hydra:Collections.  65-66 add a default option with id=-1 and title=Select collection...  Are you sure you want the default added in Hydra:Collections?  If so, you could pass a parameter to find_collections_with_edit_access asking for a default to be added and possibly passing in a default value and title.  If one isn't provided, Hydra:Collections could default to something.  There are other times find_collections_with_edit_access is called when adding 'Select collection...' would not be appropriate.

app/controllers/concerns/sufia/files_controller_behavior.rb
elrayle: I can look at that.

app/views/generic_files/_browse_everything.html.erb
elrayle: I can do that.

app/controllers/concerns/sufia/files_controller_behavior.rb
```diff
@@ -268,5 +274,13 @@ def update_metadata_from_upload_screen
         @generic_file.relative_path = params[:relative_path] if params[:relative_path]
         @generic_file.on_behalf_of = params[:on_behalf_of] if params[:on_behalf_of]
       end
+
+      def add_file_to_collection
+        collection_id = params[:collection]
+        return if collection_id.nil? || collection_id == -1
+        collection = Collection.find(collection_id)
+        collection.member_ids = [@generic_file.id].concat(collection.member_ids)
+        ActiveFedora::SolrService.instance.conn.commit if collection.save
```
cam156: @elrayle I'm not used to seeing a Solr commit.  What does this do?  I thought save would solrize correctly.

app/views/generic_files/upload/_to_collection.html.erb
elrayle: Line 2 was added as a fix for failing test [_browse_everything.html.erb_spec.rb](https://github.com/projecthydra/sufia/blob/master/spec/views/generic_file/_browse_everything.html.erb_spec.rb).  The controller always adds the default 'Select collection...', such that @user_collections is never nil.  But the test renders the page without going through the controller's #new and thus @user_collections is nil.  Maybe there is a better fix for the failing test.

app/controllers/concerns/sufia/files_controller_behavior.rb
elrayle: Needs input.

app/views/generic_files/upload/_to_collection.html.erb
elrayle: Needs input.

app/controllers/concerns/sufia/files_controller_behavior.rb
mjgiarlo: Perhaps `#find_collections_with_edit_access` in hydra-collections should take an optional param that sets a default, and the code here could invoke that?

elrayle:
Thanks for the review Mike.  I'll look at the issues you bring up.  Two issues need further comment.  I marked them with 'Needs input' to make them easier to find.

app/views/generic_files/upload/_to_collection.html.erb
mjgiarlo: Use `assign` in a `before` block in the view spec?

app/controllers/concerns/sufia/files_controller_behavior.rb
awead: .save should commit too... in fact, I'm pretty sure we've got autocommit enabled.

app/views/generic_files/_browse_everything.html.erb
elrayle: Can you confirm that proper place to put the config for hide/show upload to collection is in [sufia/models/engine.rb](https://github.com/projecthydra/sufia/blob/master/sufia-models/lib/sufia/models/engine.rb)?  If not, where should the configuration be defined?

app/controllers/concerns/sufia/files_controller_behavior.rb
elrayle: This came from the code used by Dashboard -> My Files -> select file -> Add to Collection -> Update Collection, which runs code in Hydra:Collection - [Hydra::CollectionControllerBehavior](https://github.com/projecthydra/hydra-collections/blob/master/app/controllers/concerns/hydra/collections_controller_behavior.rb) #create (line 68) -> #add_members_to_collection (line 173) -> #after_create (line 53).  I wasn't able to use this code directly.  It was reproduced in Sufia.  Ideally, Hydra:Collections would have this code generalized out of the controller so it can be called in other situations where including the controller is not an option.  I wasn't looking at updating Hydra:Collections, but I can look at that if you like.

elrayle:
DO NOT MERGE:  Waiting for permission to release gem hydra-collections to version 5.0.3.  Tests will fail until this gem is released.

elrayle:
Ready to merge.

History.md
```diff
@@ -1,5 +1,11 @@
 # History of Sufia releases
 
+## 6.3.1
```
mjgiarlo: Thanks for adding to the changelog!  Can you `s/6.3.1/6.4.0/`?

jcoyne:
Shouldn't this go in `Actor#create_metadata` rather than `#create_content`?  Is there some reason this isn't just a simple attribute that is set here? https://github.com/projecthydra/sufia/blob/1e26950153d82645d05637668f7bcd0b9f25d351/app/controllers/concerns/sufia/files_controller_behavior.rb#L271

app/views/generic_files/upload/_to_collection.html.erb
```diff
@@ -0,0 +1,5 @@
+<p>
+  <label><%= t("sufia.upload.collection.select_label") %></label>
+  <%= select_tag 'collection', options_from_collection_for_select(@user_collections, "id", "title") %>
```
jcoyne: This should probably be `collection_id` as it is the Rails convention to append `_id` when you're talking about the identifier that underlies the relationship.

app/controllers/concerns/sufia/files_controller_behavior.rb
```diff
@@ -227,7 +232,7 @@ def process_file(file)
 
         update_metadata_from_upload_screen
         actor.create_metadata(params[:batch_id])
-        if actor.create_content(file, file.original_filename, file_path, file.content_type)
+        if actor.create_content(file, file.original_filename, file_path, file.content_type, params[:collection])
```
jcoyne: Content is just for the bitstream. The collection is metadata.

PR #1308, opened 2015-09-01
Changed rights input type for /edit to select_with_modal_help to match field on batch …

kerchner:
…edit page - and also render rights_modal


mjgiarlo:
:+1: if Travis is :green_heart: 

@straleyb mentioned putting in a PR related to this too, but this view change can be made independently. Thanks, @kerchner.

PR #1312, opened 2015-09-09
upload files into a collection starting from its show page

elrayle:
Steps to use this feature:
* On collection show page, a new action was added (under thumbnail) to 'Upload files' into the current collection being viewed

upload_files_from_collection

* Clicking 'Upload file' goes to the batch upload page and pre-selects the calling collection in the 'Upload to' selection menu.

batch_upload_to_collection

* Both 'Upload files' action on collection show page and 'Upload to' selection menu on batch upload page only appear when the upload_to_collection=true
* Text for actions on collection page and upload selection menu are configurable in sufia.en.yml

Changes also include...
* includes 2 fixes for rubocop unrelated to the changes for this feature

### DIFF
--- a/app/controllers/concerns/sufia/files_controller_behavior.rb
+++ b/app/controllers/concerns/sufia/files_controller_behavior.rb
@@ -64,6 +64,7 @@ module Sufia
 
     # routed to /files/new
     def new
+      set_variables_for_new_form
       @batch_id = ActiveFedora::Noid::Service.new.mint
     end
 
@@ -156,6 +157,10 @@ module Sufia
     end
 
     protected
+
+      def set_variables_for_new_form
+        @collection_id = params[:collection] if @user_collections.collect(&:id).include? params[:collection]
+      end
 
       def set_variables_for_edit_form
         @form = edit_form
--- a/spec/controllers/generic_files_controller_spec.rb
+++ b/spec/controllers/generic_files_controller_spec.rb
@@ -21,6 +21,8 @@ describe GenericFilesController do
         end
       }
       let(:collection_id) { collection.id }
+      let(:collection_noedit) { Collection.create(title: 'test collection - NO EDIT') }
+      let(:collection_noedit_id) { collection_noedit.id }
       let(:file) { fixture_file_upload('/world.png', 'image/png') }
 
       before do
@@ -42,6 +44,18 @@ describe GenericFilesController do
           xhr :post, :create, files: [file], Filename: "The World", batch_id: 'sample_batch_id', permission: { "group" => { "public" => "read" } }, terms_of_service: '1'
           expect(response.status).to eq 422
           expect(JSON.parse(response.body).first['error']).to match(/no file for upload/i)
+        end
+      end
+
+      context "when user tries to upload to a collection they can't edit" do
+        # This shouldn't happen via the UI which will only present to the user collections they can edit.
+        it "adds new file but does not put the file in the collection" do
+          xhr :post, :create, files: [file], Filename: "The world", batch_id: batch_id, permission: { "group" => { "public" => "read" } }, terms_of_service: "1", collection: collection_noedit_id
+          expect(response).to be_success
+
+          updated_collection = Collection.find(collection_id)
+          # This is confirming that the file was NOT added to the collection the user cannot edit
+          expect(updated_collection.member_ids).to eq []
         end
       end
 
@@ -87,6 +101,11 @@ describe GenericFilesController do
           # This is confirming that apply_depositor_metadata recorded the depositor
           expect(saved_file.depositor).to eq 'jilluser@example.com'
           expect(saved_file.to_solr['depositor_tesim']).to eq ['jilluser@example.com']
+        end
+
+        it "adds new file when collection is instructions id" do
+          xhr :post, :create, files: [file], Filename: "The world", batch_id: batch_id, permission: { "group" => { "public" => "read" } }, terms_of_service: "1", collection: '-1'
+          expect(response).to be_success
         end
 
         it "adds new file to collection" do
--- a/spec/views/generic_file/new.html.erb_spec.rb
+++ b/spec/views/generic_file/new.html.erb_spec.rb
@@ -2,9 +2,9 @@ require 'spec_helper'
 
 describe 'generic_files/new.html.erb', type: :view do
   let(:user_collections) do
-    default_option = SolrDocument.new(id: -1, title_tesim: "Select collection...")
-    col1 = SolrDocument.new(id: "c1", title_tesim: "col1")
-    col2 = SolrDocument.new(id: "c2", title_tesim: "col2")
+    default_option = SolrDocument.new(id: -1, title_tesim: 'Select collection...')
+    col1 = SolrDocument.new(id: 'c1', title_tesim: 'col1')
+    col2 = SolrDocument.new(id: 'c2', title_tesim: 'col2')
     [default_option, col1, col2]
   end
   let(:generic_file) { stub_model(GenericFile, id: '123') }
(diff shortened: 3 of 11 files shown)


config/routes.rb
```diff
@@ -25,6 +25,7 @@
       post 'audit'
     end
   end
+  get 'files/new/in_collection/:collection_id' => 'generic_files#new', as: 'generic_files_upload_in_collection'
```
mjgiarlo: I'm probably overlooking something obvious, but why do we need a dedicated route for this rather than using the existing `/files/new` route with a param passed to it (e.g., `/files/new?collection=abc123def`)?

app/controllers/concerns/sufia/files_controller_behavior.rb
```diff
@@ -64,6 +64,7 @@ def load_resource_from_solr
 
     # routed to /files/new
     def new
+      set_variables_for_new_form
```
mjgiarlo: Should we also be checking to make sure that current_user .can? edit collection? That'd be a good test to add.

spec/views/generic_file/new.html.erb_spec.rb
```diff
@@ -16,7 +16,6 @@
       assign(:batch_id, batch_id)
       assign(:user_collections, user_collections)
       allow(controller).to receive(:current_user).and_return(stub_model(User))
-      allow(GenericFilesController).to receive(:upload_complete_path).with(batch_id).and_return("foo")
```
mjgiarlo: Why was this line removed?

mjgiarlo:
I have just a few questions.  Thanks for the PR, @elrayle

config/routes.rb
elrayle: I'll look at making that adjustment.

app/controllers/concerns/sufia/files_controller_behavior.rb
elrayle: The @user_collections which comes from hydra-collections returns only the list of collections the user has permissions controlled by line before_action on [line 43](https://github.com/projecthydra/sufia/blob/upload_with_collection/app/controllers/concerns/sufia/files_controller_behavior.rb) (find_collections_with_edit_access).  Then in set_variables_for_new_form, it checks that the passed in collection_id is in the @user_collections list.  Seems sufficient to me.  Does that work for you?

spec/views/generic_file/new.html.erb_spec.rb
elrayle: It didn't seem to be doing anything in the test since the test passes without it.  The only other test that uses something similar is controllers/generic_files_controller_spec.rb.  So honestly, I don't know why I put it in there in the first place.  It seems like kruft to me, so I removed it.

app/controllers/concerns/sufia/files_controller_behavior.rb
elrayle: I'm adding two more tests that cover the scenario where a user tries to add a file to a collection they cannot edit.

mjgiarlo: :+1: Thanks much for indulging my paranoia. :)

config/routes.rb
mjgiarlo: Thank you!

spec/views/generic_file/new.html.erb_spec.rb
mjgiarlo: :+1:

app/controllers/concerns/sufia/files_controller_behavior.rb
elrayle: I never mind adding more tests.  I threw in a couple of others just to be more complete.

elrayle:
Updated as recommended.

spec/controllers/generic_files_controller_spec.rb
```diff
@@ -89,6 +103,16 @@
           expect(saved_file.to_solr['depositor_tesim']).to eq ['jilluser@example.com']
         end
 
+        it "adds new file when collection is is instructions id passed as numeric" do
+          xhr :post, :create, files: [file], Filename: "The world", batch_id: batch_id, permission: { "group" => { "public" => "read" } }, terms_of_service: "1", collection: -1
+          expect(response).to be_success
+        end
+
+        it "adds new file when collection is is instructions id passed as string" do
```
mjgiarlo: Typo on lines 111 and 106: "is is".

sufia-models/app/actors/sufia/generic_file/actor.rb
```diff
@@ -41,7 +41,7 @@ def create_content(file, file_name, path, mime_type, collection_id = nil)
     end
 
     def add_file_to_collection(collection_id)
-      return if collection_id.nil? || collection_id == -1
+      return if collection_id.nil? || ([-1, "-1"].include? collection_id)
```
mjgiarlo: Do you find that this ID is sometimes passed as a string and other times as a fixnum?

elrayle:
@mjgiarlo, your paranoia is contagious.  I decided to write a test at the lowest level  to see if a user could add a file to a collection they do not own, and guess what, they can.  There are two places that the fix could go in, 1) collection.add_members, 2) a check before calling collection.add_members

I'm not sure hydra-collections knows anything about users, which would lean toward option #2.

I don't see a way that the user can pass a collection they don't own to [actor](https://github.com/projecthydra/sufia/blob/master/sufia-models/app/actors/sufia/generic_file/actor.rb).create_content (line 30) which calls add_file_to_collection (line 43) because there are several places in the UI that prevent that and users don't have a means to call that method directly.  But being real paranoid, if they some how did call actor.create_content with a collection they can't edit, the code will happily put the file in that collection.

Anyone want to weigh in on this?

elrayle:
Ready for review.

mjgiarlo:
Any thoughts on @elrayle's comment https://github.com/projecthydra/sufia/pull/1312#issuecomment-139546211 above, @cam156 @awead @jcoyne @hectorcorrea ?

elrayle:
BTW, I took approach #2.

mjgiarlo:
I'm inclined to :+1: but let's give the others a couple hours to take a look. @cam156 @jcoyne @hectorcorrea @awead

awead:
I like it! One question: If the user can't edit the collection, maybe they shouldn't even be able to even attempt to add the files. It seems odd to me if they can go ahead and add the file, but then it doesn't get it added to the collection. Are there any messages informing them of this outcome?

mjgiarlo:
@awead should a flash alert be populated there? would that suffice?

awead:
@mjgiarlo sure, but I'm wondering, why not hide the "Add files" button entirely if they don't have edit access to the collection?

elrayle:
@awead  ++  There is a check to see if the user can edit the collection in [_show_actions.html.erb](https://github.com/projecthydra/sufia/blob/upload_with_collection/app/views/collections/_show_actions.html.erb).  The actions (i.e., Edit, Add files, Upload files) only appear on the collection show if the user can edit.

elrayle:
The test to be sure the user can edit before adding members is to stop malicious attacks.  If they go through the UI, they will never be offered the opportunity to add file to a collection they can't edit.

awead:
ah, good then! thanks for the explanation, @elrayle

cam156:
@elrayle have you done anything with the timing issue that occurs when more than one thread tries to update the collection?  We have has this issue in ArchiveSPhere, and @jcoyne just hit this issue in CurationConcerns with multi file upload.

elrayle:
I was unaware of this issue.  Do you have an issue number for reference?

elrayle:
@cam156 and others, would you be comfortable accepting this PR and opening an issue for the concurrent update problem?  This PR is not the cause of the concurrent update problem.  Additionally, this PR works in conjunction with PR https://github.com/projecthydra/sufia/pull/1307 which added the basic functionality on which this depends.  

Note that this functionality is turned off by default.

mjgiarlo:
Sounds reasonable to me. Thanks, @elrayle!

mjgiarlo:
Thanks too to @cam156 and @awead for comments.

PR #1314, opened 2015-09-15
Moving admin statistics logic out of the controller.

carolyncole:
Specifically leaving the admin stats test alone to prove I did not break the original code

### DIFF
--- a/app/controllers/concerns/sufia/admin/stats_behavior.rb
+++ b/app/controllers/concerns/sufia/admin/stats_behavior.rb
@@ -8,89 +8,15 @@ module Sufia
       end
 
       def index
-        @presenter = AdminStatsPresenter.new
+        # initialize the presenter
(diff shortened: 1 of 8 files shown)


cam156:
@jcoyne can you clarify what you mean by this all?   The SystemStats class? The calls to the SystemStats?

jcoyne:
@cam156 I was thinking the call to system stats class could be encapsulated in the presenter, since it only sets variables on the presenter.

cam156:
@jcoyne that make sense to me. I will work on that.

cam156:
@jcoyne moved the stats admin calls to the presenter. Better?

mjgiarlo:
Travis is :unamused: 

```
     Failure/Error: Sufia::AdminStatsPresenter.new.tap do |p|
     ArgumentError:
       wrong number of arguments (0 for 2)
```

cam156:
Not sure why this is showing a failing build.   When you click trough to travis all builds passed.

jcoyne:
Such an improvement! :clap:

PR #1316, opened 2015-09-15
Adding csv generation service to create csv formatted GenericFiles

carolyncole:


sufia-models/app/models/concerns/sufia/generic_file/csv.rb
```diff
@@ -0,0 +1,23 @@
+module Sufia
+  module GenericFile
+    module CSV
+      def to_csv
+        ::CSV.generate do |csv|
```
mjgiarlo: Should this method also spit out a header row so that folks know what the various values are for?

sufia-models/app/models/concerns/sufia/generic_file/csv.rb
```diff
@@ -0,0 +1,23 @@
...
+          csv << csv_terms.map do |term|
+            values = send(term)
+            values = [values].flatten # make sure we have an array
+            values.join(multi_value_separator)
```
mjgiarlo: @jcoyne has worked on a CSV importer that uses multiple columns with the same name to handle repeatability which is a little safer (because metadata values can contain the pipe character or any other weird character(s)). Would this be acceptable for the use case you're working on here or would that, e.g., break the way Excel imports the CSV?

sufia-models/app/models/concerns/sufia/generic_file/csv.rb
```diff
@@ -0,0 +1,23 @@
+module Sufia
+  module GenericFile
+    module CSV
+      def to_csv
+        ::CSV.generate do |csv|
+          csv << csv_terms.map do |term|
+            values = send(term)
+            values = [values].flatten # make sure we have an array
```
mjgiarlo: You can also use `Array(values)` (or `Array.wrap(values)` in Rails environments) to do this in a way that may be more idiomatic.

spec/models/generic_file_spec.rb
```diff
@@ -744,4 +744,18 @@ def paranoid_edit_permissions
...
+  describe "to_csv" do
+    before do
+      @file.title = ["My Title"]
+      @file.creator = ["Von, Creator"]
+      @file.resource_type = ["Book", "Other"]
+      @file.rights = ["Mine"]
+      @file.mime_type = "application/pdf"
+    end
+    subject { @file.to_csv }
+    it do
```
mjgiarlo: I get what's going on here but having an `it do` block looks weird to me. Is this a rubocop auto-fix? I'd expect to see:

```ruby
it { is_expected.to eq 'My Title, ...' }
it 'parses into an array' do
  expect(CSV.parse(subject)).to eq [['My Title', ... ]]
end
```
(And do we expect an array there or an array of arrays?)

cam156: @mjgiarlo This was my effort to make the test run a bit faster.  If you break it the way you suggest then the subject gets loaded twice, and is slower.

mjgiarlo: :ok: that makes sense!

mjgiarlo: Maybe a minor reformat would suffice (there are currently 0 `it do`s in the codebase):

```ruby
it 'parses into an array' do
  expect(subject).to eq 'My Title,...'
  expect(CSV.parse(subject)).to eq [['My Title', ... ]]
end
```

And do we want that to return an array of arrays, or just an array?

sufia-models/app/models/concerns/sufia/generic_file/csv.rb
cam156: @mjgiarlo I would think it would be hard to make a consistent table with multiple files if one row has 3 values and the next only has two.  @mjgiarlo How is @jcoyne's importer handling that?  Is is he determining the maximum and having all rows the same?  for example I have one file with 2 contributors and one file with 1.  How does that show up in the file?

sufia-models/app/models/concerns/sufia/generic_file/csv.rb
cam156: @mjgiarlo good idea,  I will add an optional header row.

sufia-models/app/models/concerns/sufia/generic_file/csv.rb
mjgiarlo: I'd have to defer to @jcoyne there, but I should say that I wouldn't hold up the PR because of this.

cam156:
@mjgiarlo I still have done nothing good with the multi valued fields.  I ran timing tests and having the two test separate was such a slim difference there was little reason to do it.

mjgiarlo:
@cam156 :ok: I'm curious why you made the header row optional.  Other than that, :+1:

cam156:
@mjgiarlo I am envisioning a loop for a list of files.  Therefor I would only want the header to occur once at the top of the output for the list of files in CVS format
title heading,....
title1,...
title2,...

mjgiarlo:
@cam156 Ah, duh. OK, then I think it's safe to remove the header from this PR (and add the header back when you write the code to do the looping). Sorry for the hassle!

cam156:
@mjgiarlo The header row seems like a good addition to me since another user could be just outputting a single file.  Did you want me to remove it?

mjgiarlo:
I'm :ok: with it staying there.

jcoyne:
So, this indicates there is only a single representation in CSV for the file.  I can imagine a case where there are more types than that.  How about making this a service instead and not adding it on the model itself?

cam156:
@jcoyne Ok.  I can look into it tomorrow. Do you have a service in mind you would like me to emulate?

cam156:
@jcoyne :bomb: And now it is a service...

PR #1317, opened 2015-09-18
Reformatting the statistics page to only have one date filter form

cam156:
currently there are 3 forms which are entirely independent and the data on the page is only partially filtered.

**Current Form with many date filters**

screen shot 2015-09-18 at 6 48 34 am

**New form with one date filter**

screen shot 2015-09-18 at 11 15 33 am


mtribone:
Looks good from a source order perspective.

PR #1320, opened 2015-10-01
Use sufia.root_path instead of root_path

randycoulman:
In the navigation bar, the `Home` link and logo go to `root_path`.  When Sufia is mounted as an engine at a sub-path, this will take the user to the `root_path` of the host Rails application rather than back to
the main Sufia page.

This change makes it so that the Home link and logo navigate to Sufia’s `root_path` instead, so that the Sufia interface is more self-contained.

If this is not the desired behavior, that's fine.  But in that case, _logo.html.erb should be modified to use `root_path` instead of the hard-coded `/` that is there now.


mjgiarlo:
:+1: 

Thanks, @randycoulman

Issue #1325, opened 2015-10-02
Configure mini_magick to use posix-spawn instead of open3


hackmastera:
I and many others have observed failed resque jobs as reported in https://groups.google.com/forum/#!searchin/hydra-tech/derivatives$20deadlock/hydra-tech/_L2ysL1A0M8/2agSkvWVFAAJ

(see that thread for stack trace).

Since many of us have made this switch locally, I propose we include a generator to make posix-spawn the default configuration. PR to follow.

There may or may not be a fix needed at the hydra-derivatives level, see https://github.com/projecthydra/hydra-derivatives/issues/81 for more info.

PR #1326, opened 2015-10-02
Configure mini_magick to use posix_spawn instead of open3.

hackmastera:
Fixes projecthydra/sufia#1325

### DIFF
--- a/History.md
+++ b/History.md
@@ -2,6 +2,7 @@
 
 ## 6.3.1
 
+* Configure minimagick to use posix-spawn instead of popen3 (rake task must be run). [Anna Headley]
 * Add configuration allowing select menu on batch upload to upload files to a collection. [E. Lynette Rayle]
 * Bump active-fedora version to 9.4 [E. Lynette Rayle]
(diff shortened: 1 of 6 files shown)


mjgiarlo:
Looks good to me. @jcoyne @cam156 @terrellt @awead

awead:
:+1:

hackmastera:
I think this will only fix the problem for newly-generated projects. Not sure how this would get to be part of a release's upgrade path.

mjgiarlo:
@hackmastera You can use an approach like the one you see here:

https://github.com/hackmastera/sufia/commit/9fd48874f034299f07fafab893f44172a1f52121

Basically, put your generator change into its own named generator and invoke that from the install generator. That way we can hook into the named generator in the next upgrade. Does that make sense?

Feel free to update the release notes (History.md) in this PR too so that when we're cutting the next release (which will probably be quite big since we haven't cut one in a while) we include this in the notes.

hackmastera:
@mjgiarlo does that look right? if so i will squash.

PR #1330, opened 2015-10-16
DO NOT MERGE - Test using Fedora 4.4

awead:
This will be be broken until we can release a new version of ActiveFedora with support for Fedora's new fixity predicates.


cam156:
Looks like you need both comits to make Travis happy.

mjgiarlo:
@cam156 Yep.

PR #1333, opened 2015-10-21
Disentangle shared state and fix issue that was revealed by different AF::Base#where behavior in latest AF.

mjgiarlo:


mjgiarlo:
@awead FWIW, I suspect we still need this change in AF < 9.5 and Fedora 4.2 (since, as you pointed out, we were merely *hiding* the issues before).

awead:
@mjgiarlo yes, see https://github.com/projecthydra/sufia/tree/fedora-4.2

mjgiarlo:
:+1: right on!

PR #1337, opened 2015-10-30
Fixes empty white space being clickable when the intrinsic value of the thumb image is...

mtribone:
…smaller than the width of the containing block. Overrides the Bootstrap .img-responsive display: block but inherits max-width to keep the image from pixellating.

The issue:

screen shot 2015-10-30 at 1 35 04 pm

The solution:

screen shot 2015-10-30 at 1 35 41 pm

### DIFF
--- a/app/assets/stylesheets/sufia/_file-show.scss
+++ b/app/assets/stylesheets/sufia/_file-show.scss
@@ -22,3 +22,7 @@
   -webkit-column-width: 20em;
   column-width: 20em;
 }
+
+.img-responsive, figure {
+  display: inline-block;
+}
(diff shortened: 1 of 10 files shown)


hackmastera:
@mtribone this is mostly working but you need to also remove the '+ title' portion of this line: https://github.com/projecthydra/sufia/blob/master/app/helpers/generic_file_helper.rb#L16

mtribone:
@hackmastera Will do. And I'll see what is up with the Travis failure, too.

hackmastera:
refs projecthydra/sufia/issues/1335

PR #1338, opened 2015-11-02
Make geonames username configurable. Closes #1336

### LINKED ISSUE #1336: geonames username should be configurable
The username used with geonames api is hard coded into the sufia-model. Someone using the sufia gem cannot change it easily.
The api username should be configurable for other applications who are using the sufia gem.

tonmoy18:
Removed change from History.md

### DIFF
--- a/sufia-models/lib/generators/sufia/models/geonames_username_config_generator.rb
+++ b/sufia-models/lib/generators/sufia/models/geonames_username_config_generator.rb
@@ -1,0 +1,20 @@
+require 'rails/generators'
+
+class Sufia::Models::GeonamesUsernameConfigGenerator < Rails::Generators::Base
+  source_root File.expand_path('../templates', __FILE__)
+
+  desc """
+This Generator makes the following changes to your application:
+  1. Updates existing sufia.rb initializer to include a geonames_username configuration
+       """
+
+  def banner
+    say_status("info", "ADDING GEONAMES_USERNAME OPTION TO SUFIA CONFIG", :blue)
+  end
+
+  def inject_config_initializer
+    inject_into_file 'config/initializers/sufia.rb', before: "# Where to store tempfiles, leave blank for the system temp directory (e.g. /tmp)" do
+      "config.geonames_username = ''\n"
+    end
+  end
+end
--- a/sufia-models/lib/generators/sufia/models/templates/config/sufia.rb
+++ b/sufia-models/lib/generators/sufia/models/templates/config/sufia.rb
@@ -152,6 +152,8 @@ Sufia.config do |config|
   # NOTE: if you have always sent analytics to GA for downloads and page views leave this commented out
   # config.analytic_start_date = DateTime.new(2014,9,10)
 
+  config.geonames_username = ''
+
   # If browse-everything has been configured, load the configs.  Otherwise, set to nil.
   begin
     if defined? BrowseEverything
--- a/sufia-models/lib/sufia/models/engine.rb
+++ b/sufia-models/lib/sufia/models/engine.rb
@@ -34,6 +34,7 @@ module Sufia
       config.activity_to_show_default_seconds_since_now = 24 * 60 * 60
       config.arkivo_api = false
       config.retry_unless_sleep = 0.0
+      config.geonames_username = ""
 
       # Noid identifiers
       config.enable_noids = true
--- a/spec/models/geo_names_resource_spec.rb
+++ b/spec/models/geo_names_resource_spec.rb
@@ -1,6 +1,16 @@
 require 'spec_helper'
 
 describe GeoNamesResource, type: :model do
+  before do
+    state = {
+      "totalResultsCount" => 1, "geonames" => [{
+        "countryId" => "1327865", "adminCode1" => "11", "countryName" => "Myanmar [Burma]", "fclName" => "country, state, region,...", "countryCode" => "MM", "lng" => "98", "fcodeName" => "first-order administrative division", "toponymName" => "Shan State", "fcl" => "A", "name" => "Shan State", "fcode" => "ADM1", "geonameId" => 1_297_099, "lat" => "22", "adminName1" => "Shan", "population" => 5_815_384
+      }]
+    }
+    ActiveResource::HttpMock.respond_to do |mock|
+      mock.get "/searchJSON?maxRows=10&q=State&username=", {}, state.to_json
+    end
+  end
   it "finds locations" do
     hits = described_class.find_location("State")
     expect(hits).not_to be_nil
--- a/sufia-models/app/models/geo_names_resource.rb
+++ b/sufia-models/app/models/geo_names_resource.rb
@@ -13,6 +13,6 @@ class GeoNamesResource < ActiveResource::Base
   end
 
   def self.find_location(location)
-    GeoNamesResource.find(:all, params: { q: location, username: "cam156", maxRows: 10 })
+    GeoNamesResource.find(:all, params: { q: location, username: Sufia.config.geonames_username, maxRows: 10 })
   end
 end


mjgiarlo:
Hi @tonmoy18 

This is a great start. I have a few additional requests.

Can you also:

1. add the following lines to `sufia-models/lib/generators/sufia/models/templates/config/sufia.rb` so that applications will have this new config value generated into them?

```ruby
  # Specifies a user account for the Geonames API (please override this value)
  # Default is empty string
  config.geonames_username = ''
```

2. add a generator that injects the above config into existing Sufia applications. see here for an example: https://github.com/projecthydra/sufia/pull/1307/files#diff-d427f194c17a07478bef7b9cefa34759

3. have the install generator call your new generator. see here for an example: https://github.com/projecthydra/sufia/pull/1307/files#diff-e0c165d9fe01a2b2b88f7d151436ff4e

:question: 

These changes help make sure that all Sufia users, whether they're new installs or upgrades, have access to all the same config.

Thanks for the PR!

sufia-models/lib/sufia/models/engine.rb
```diff
@@ -31,6 +31,7 @@ class Engine < ::Rails::Engine
       config.activity_to_show_default_seconds_since_now = 24 * 60 * 60
       config.arkivo_api = false
       config.retry_unless_sleep = 0.0
+      config.geonames_username = "cam156"
```
mjgiarlo: I think we should default this value to an empty string. This will be a breaking change, but I don't think all sufia instances should be using the same config value. Thanks for finding this and working on it, @tonmoy18

tonmoy18:
I think I modified the things you mentioned. Please let me know if something else needs changing. And thanks again

mjgiarlo:
I'll have a look early (Pacific time) tomorrow, @tonmoy18. Thanks!

tonmoy18:
Awesome, thanks

sufia-models/lib/generators/sufia/models/geonames_username_config_generator.rb
```diff
@@ -0,0 +1,21 @@
...
+This Generator makes the following changes to your application:
+  1. Updates existing sufia.rb initializer to include a geonames_username configuration
+       """
+
+  def banner
+    say_status("info", "ADDING GEONAMES_USERNAME OPTION TO SUFIA CONFIG", :blue)
+  end
+
+  def inject_config_initializer
+    inject_into_file 'config/initializers/sufia.rb', before: "# Specifies a user account for the Geonames API (please override this value)" do
```
mjgiarlo: I think you want the "# Specifies a user account..." string on the next line, and you want an existing string in the Sufia config here. Like:

```ruby
inject_into_file 'config/initializers/sufia.rb', before: "# Where to store tempfiles, leave blank for the system temp  directory (e.g. /tmp)" do
  "# Specifies a user account for the Geonames API (please override this value).\n" +
  "# Default is empty string\n" +
  "config.geonames_username = ''\n"
end
```

sufia-models/lib/generators/sufia/models/install_generator.rb
```diff
@@ -127,4 +128,9 @@ def upload_to_collection_config
   def minimagic_config
     generate 'sufia:models:minimagick_config'
   end
+
+  # Adds geonames_username initialization
+  def upload_to_collection_config
```
mjgiarlo: The method on line 123 is already named this. Rename to something like `geonames_config`.

mjgiarlo:
@tonmoy18 Just a couple more changes and then I think we're ready to go.

tonmoy18:
@mjgiarlo Fixed the things you mentioned. It seems that there is still one rubocop offense.. It says that the class in install_generator has too many lines (104 of 100). Any suggestion?

mjgiarlo:
@tonmoy18 Yeah, that was bound to happen. Feel free to add an exception for this file around here:

https://github.com/projecthydra/sufia/blob/master/.rubocop.yml#L38-L42

And once you do that, could you also squash your commits?

tonmoy18:
I wanted to squash once all tests were passing. I will fix these things soon.

mjgiarlo:
OK, good call, @tonmoy18!

tonmoy18:
Fixed those things, not sure why the tests are still failing @mjgiarlo

mjgiarlo:
Hi @tonmoy18, this may have been related to the new version of engine_cart that came out. Would you mind trying another rebase?

tonmoy18:
@mjgiarlo Thanks, waiting for the latest offenses to be fixed, I guess there was a newer release of rubocop.

tonmoy18:
@mjgiarlo If you want to, I can fix these offenses on a different PR

mjgiarlo:
I just merged another PR about rubocop. Rebase?

tonmoy18:
It worked. The geonames username is blank by default and I am not quite sure how to mock the active resource http call to geonames, so I couldn't get the geonames test to pass.

tonmoy18:
@mjgiarlo I had to change the spec example a little bit to pass the test, I hope it is all right now

awead:
@tonmoy18 this will need another rebase due to some config changes with #1346 but after that, if @mjgiarlo is good, we should merge and look at releasing 6.4,

mjgiarlo:
@tonmoy18 Re: testing, another strategy that wouldn't require mocking could be setting `Sufia.config.geonames_username` to `'cam156'` in the `before` block of that spec.

After you rebase, would you make one change to the PR? I gave you some bad advice at the outset. Namely, could you revert all changes to the `install_generator`? We don't need to run your new generator for new installs because you've already made the necessary changes in `sufia.rb`

Thanks for your patience! I'm looking forward to including this work in the next release of Sufia.

tonmoy18:
@mjgiarlo <https://github.com/mjgiarlo> No problem at all. But I already
added HttpMock in test, should I remove that too?

mjgiarlo:
@tonmoy18 Let's see what some others think about the change in this test to use a mock. @cam156 @jcoyne  ?

sufia-models/lib/generators/sufia/models/geonames_username_config_generator.rb
```diff
@@ -0,0 +1,22 @@
...
+       """
+
+  def banner
+    say_status("info", "ADDING GEONAMES_USERNAME OPTION TO SUFIA CONFIG", :blue)
+  end
+
+  def inject_config_initializer
+    inject_into_file 'config/initializers/sufia.rb', before: "# Where to store tempfiles, leave blank for the system temp directory (e.g. /tmp)" do
+      "# Specifies a user account for the Geonames API (please override this value)\n" +
+        "# Default is empty string\n" +
```
jcoyne: I don't think we need to document the default and supply the default. It is obvious that it is an empty string.

sufia-models/lib/generators/sufia/models/templates/config/sufia.rb
```diff
@@ -152,6 +152,10 @@
   # NOTE: if you have always sent analytics to GA for downloads and page views leave this commented out
   # config.analytic_start_date = DateTime.new(2014,9,10)
 
+  # Specifies a user account for the Geonames API (please override this value)
+  # Default is empty string
```
jcoyne: This is the defaults are set and it's plain to see it's an empty string. I don't think this documentation is necessary.

jcoyne:
I'm happy with the mocking strategy. I just left a few concerns about unnecessary documentation.

.rubocop.yml
```diff
@@ -40,7 +40,6 @@ Metrics/ClassLength:
     - 'lib/generators/sufia/templates/catalog_controller.rb'
     - 'sufia-models/app/models/datastreams/fits_datastream.rb'
     - 'sufia-models/app/actors/sufia/generic_file/actor.rb'
-    - 'sufia-models/lib/generators/sufia/models/install_generator.rb'
```
jcoyne: rubocop is now complaining about this. You may need to revert this.

sufia-models/lib/generators/sufia/models/geonames_username_config_generator.rb
tonmoy18: should this method be completely removed?

sufia-models/lib/generators/sufia/models/geonames_username_config_generator.rb
```diff
@@ -0,0 +1,14 @@
+require 'rails/generators'
+
+class Sufia::Models::GeonamesUsernameConfigGenerator < Rails::Generators::Base
+  source_root File.expand_path('../templates', __FILE__)
+
+  desc """
+This Generator makes the following changes to your application:
+  1. Updates existing sufia.rb initializer to include a geonames_username configuration
+       """
+
```
mjgiarlo: Where'd the method to inject the config go? We need that for the upgrade path.

tonmoy18: sorry, I got confused, adding it now.

PR #1339, opened 2015-11-04
Refactor BatchUpdateJob

awead:
Allows for overriding BatchUpdateJob to inject additional jobs or additional metadata on each of the files in the batch.

### DIFF
--- a/sufia-models/app/jobs/batch_update_job.rb
+++ b/sufia-models/app/jobs/batch_update_job.rb
@@ -1,3 +1,4 @@
+# Resque job that updates files uploaded via the BatchController
 class BatchUpdateJob
   include Hydra::PermissionsQuery
   include Sufia::Messages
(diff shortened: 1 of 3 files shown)


sufia-models/app/jobs/batch_update_job.rb
```diff
@@ -6,44 +7,45 @@ def queue_name
...
     if denied.empty?
-      send_user_success_message(user, batch) unless saved.empty?
+      send_user_success_message unless saved.empty?
     else
-      send_user_failure_message(user, batch)
+      send_user_failure_message
     end
   end
 
-  def update_file(gf, user)
```
jcoyne: Why are you changing signatures? That's not backwards compatible.

mjgiarlo: If these aren't private methods, perhaps they oughta be.

awead:
I made it private, for @mjgiarlo, and added an optional user to the method params for @jcoyne.

sufia-models/app/jobs/batch_update_job.rb
```diff
@@ -6,67 +7,103 @@ def queue_name
...
     message = denied.count > 1 ? multiple_failure(batch.id, denied) : single_failure(batch.id, denied.first)
     User.batchuser.send_message(user, message, failure_subject, false)
   end
+
+  private
+
+    def update_file(gf, you = nil)
+      you ||= user
+      unless you.can? :edit, gf
+        ActiveFedora::Base.logger.error "User #{you.user_key} DENIED access to #{gf.id}!"
```
grosscol: Will `#{you.user_key}` throw an error with when `you` is `nil`?  If so, why the default value of `nil` for `you`?

awead: If you is nil, it get's set to user on line 62. Right?

grosscol: misinterpreted that when I read it.

PR #1350, opened 2015-11-12
Remove deprecation warnings

awead:
ping @jcoyne @elrayle


app/controllers/concerns/sufia/homepage_controller.rb
```diff
@@ -4,7 +4,7 @@ module Sufia::HomepageController
   included do
     # Adds Hydra behaviors into the application controller
     include Hydra::Controller::ControllerBehavior
-    include Blacklight::Catalog::SearchContext
```
jcoyne: In order to make this change, you'll need to pin to blacklight ~> 5.16

jcoyne:
Just need to update the gemspec for blacklight and :+1:

PR #1352, opened 2015-11-16
log upload errors separate to help with debugging

elrayle:
Add a log file for recording upload errors.  This helps in debugging upload problems.
* adds `config.upload_logger` to create an error log specific to upload errors in default location `/logs/upload_error.log`


sufia-models/app/actors/sufia/generic_file/actor.rb
```diff
@@ -46,7 +46,12 @@ def add_file_to_collection(collection_id)
       return unless user.can? :edit, collection
       acquire_lock_for(collection_id) do
         collection.add_members [generic_file.id]
-        collection.save
+        begin
+          collection.save
+        rescue StandardError => error
+          Sufia.config.upload_logger.warn "Sufia::GenericFile::Actor::add_file_to_collection Error adding #{generic_file.title} to collection #{collection.title}. Caught Exception error #{error.inspect}"
```
mjgiarlo: Is there some Rails magic here? I didn't think you could treat a string as a logging object.

sufia-models/app/actors/sufia/generic_file/actor.rb
```diff
@@ -102,11 +107,22 @@ def save_and_record_committer
...
+        retry
+      rescue Ldp::Gone => error
+        Sufia.config.upload_logger.warn "Sufia::GenericFile::Actor::save_and_record_committer Error saving #{generic_file.title}. Caught LDP error #{error.inspect}"
         save_tries += 1
         # fail for good if the tries is greater than 3
         raise error if save_tries >= 3
         sleep 0.01
         retry
+      rescue StandardError => error
+        Sufia.config.upload_logger.warn "Sufia::GenericFile::Actor::save_and_record_committer Error saving #{generic_file.title}. Caught Exception error #{error.inspect}"
```
mjgiarlo: The changes in this file bring up a couple related questions:

1. Why doesn't the `StandardError` rescue block also deal with `save_tries` and the other code in the above blocks?
2. Seeing all the repetition in the blocks... maybe we should just rescue all exceptions and treat them all the same way.

sufia-models/lib/generators/sufia/models/templates/config/sufia.rb
```diff
@@ -85,6 +85,9 @@
   # Requires a Google Analytics id and OAuth2 keyfile.  See README for more info
   config.analytics = false
 
+  # Specify file for logging upload file errors.
+  config.upload_logger = 'log/upload_error.log'
```
mjgiarlo: I'd like to see this feature off by default, for folks who already use Sufia and expect all their logging to go into one place.

jcoyne:
I don't want a separate error log to have to look through. I think that will make debugging and log scanning even more difficult.

elrayle:
I find looking through development.log onerous to wade through.  You basically have to know the error in advance so you can search for the error by name and then explore what happened.  There is a lot of stuff written to the log that is of no use in debugging.  It makes finding errors of interest extremely difficult.  The only thing logged to upload_error.log is actual errors that occur during an attempt to upload.

sufia-models/app/actors/sufia/generic_file/actor.rb
elrayle: I have no problem consolidating.  I'm not sure why Solr error was singled out for retry.  Now that I understand the meaning of the Ldp::Gone error, it doesn't make sense to do a retry for that.  So the other option is to retry for Solr errors and simply log for all others.

sufia-models/app/actors/sufia/generic_file/actor.rb
elrayle: There is no Rails magic.  Sufic.config.upload_logger is an instance of Logger that is created in initializers/sufia.rb.

elrayle:
The error in the tests is, I believe, related to the fact that log/upload_error.log does not exist prior to testing.  In my rails app that uses this, upload_error.log is created automatically if it doesn't exist.  Perhaps something else is causing the error.

sufia-models/lib/generators/sufia/models/templates/config/sufia.rb
```diff
@@ -85,6 +85,9 @@
   # Requires a Google Analytics id and OAuth2 keyfile.  See README for more info
   config.analytics = false
 
+  # Specify file for logging upload file errors.
+  config.upload_logger = Logger.new('log/upload_error.log')
```
elrayle: This is where config.upload_logger is created as an instance of Logger.

mjgiarlo: I must have misread this line the first time around. :ok:

sufia-models/app/actors/sufia/generic_file/actor.rb
mjgiarlo: :ok_hand: I missed that the first time around.

sufia-models/app/actors/sufia/generic_file/actor.rb
mjgiarlo: Sounds good, @elrayle 

I understand that there are different perspectives on logging represented in this PR, and I think this is a valid point of differing opinions. Therefore, I recommend that we merge this feature but make it optional and have the default be off; folks should opt in to using a secondary logger. Can you tweak the PR so that the default value of `upload_logger` is `nil`, and have the calls to `upload_logger.warn` check first to make sure the value of `upload_logger` isn't `nil`? How does that sound, @jcoyne @elrayle ?

mjgiarlo:
I understand that there are different perspectives on logging represented in this PR, and I think this is a valid point of differing opinions. Therefore, I recommend that we merge this feature but make it optional and have the default be off; folks should opt in to using a secondary logger. Can you tweak the PR so that the default value of `upload_logger` is `nil`, and have the calls to `upload_logger.warn` check first to make sure the value of `upload_logger` isn't `nil`? How does that sound, @jcoyne @elrayle ?

mjgiarlo:
Hi, @elrayle. We're getting very close to merging sufia-core back into sufia. Is it :ok:  if we wait on this PR until that has happened, and you've rebased this atop sufia-core?

elrayle:
I may just drop this if others aren't interested.  I don't have time to spend debugging the tests and it is easy enough for me to add a logger on the fly when I need one for my own debugging.  If you want this too, speak up or it will be closed.

hackmastera:
It strikes me that the mailboxer notifications are the real problem w/r/t logs. At the risk of hijacking this thread, what is the point of those? Maybe it would be more useful to turn them off? Is such a thing possible?

hectorcorrea:
@hackmastera I have always found the mailboxer messages cumbersome. I added this to Sufia so that at development time we can suppress them or at least reduce the number of entries that it logs: https://github.com/projecthydra/sufia/commit/d8d2b25ca422508991668ee5a4155b292ace66af

hackmastera:
@hectorcorrea I saw that and tried to use it but I don't think it worked for me for some reason. Definitely intend to try again because it is super helpful.

Is there a reason they need to be in any of the logs, though?

I'd rather have a way to turn them off without changing the actual functionality as is required for d8d2b25.

hectorcorrea:
@hackmastera One issue with the current query string parameter approach that I used is that you must submit the query string parameter on each page that you visit. If you change to a different page you need to manually add the query string parameter again.

I suspect the current logging that we are seeing is the default Rails log level. What makes this a problem is that we call the mailbox controller every 5 seconds and all of a sudden the default level does not make sense. I suspect we could change the logging level for this one controller (to error or warning for example) and that would help.

mjgiarlo:
@elrayle Are you still inclined to close this PR for now?

mjgiarlo:
Closing since @elrayle was leaning in that direction. Feel free to reopen to take this forward. Thanks for all your contributions, @elrayle!

Issue #1354, opened 2015-11-20
Remove ugly trailing underscore on 'pop-out' icon in rights field


hackmastera:
Here's a screenshot:

![ugly-underscore-screen-shot](https://cloud.githubusercontent.com/assets/845363/11309245/b7e4bc0e-8f90-11e5-89f3-2f1110d12eed.png)

PR to follow.

mjgiarlo:
:+1: :sparkles:

PR #1355, opened 2015-11-23
Remove space after icon if no text is present, fix projecthydra/sufia…

hackmastera:
…#1354

### DIFF
--- a/app/helpers/sufia/sufia_helper_behavior.rb
+++ b/app/helpers/sufia/sufia_helper_behavior.rb
@@ -114,8 +114,10 @@ module Sufia
     end
 
     def iconify_auto_link(text, showLink = true)
+      # this block is only executed when a link is inserted;
+      # if we pass text containing no links, it just returns text.
       auto_link(text) do |value|
-        "<i class='glyphicon glyphicon-new-window'></i>&nbsp;#{value if showLink}<br />"
+        "<i class='glyphicon glyphicon-new-window'></i>#{('&nbsp;' + value) if showLink}<br />"
       end
     end
 


app/helpers/sufia/sufia_helper_behavior.rb
```diff
@@ -114,8 +114,10 @@ def link_to_field(fieldname, fieldvalue, displayvalue = nil)
     end
 
     def iconify_auto_link(text, showLink = true)
+      # this block is only executed when a link is inserted;
+      # if we pass text containing no links, it just returns text.
       auto_link(text) do |value|
-        "<i class='glyphicon glyphicon-new-window'></i>&nbsp;#{value if showLink}<br />"
+        "<i class='glyphicon glyphicon-new-window'></i>#{"&nbsp;#{value}" if showLink}<br />"
```
mbklein: Nested `#{}`s almost never work. I think what you need here is something like `#{'(&nbsp;' + value) if showLink}`

jcoyne:
How about a test for this method?

app/helpers/sufia/sufia_helper_behavior.rb
hackmastera: Hm, works fine when I look at it in the internal-app. but maybe it's confusing a parser or compiler; will try your suggestion.

hackmastera:
@mbklein thanks for the suggestion; unfortunately no effect. It's nicer to look at your way, though.

@awead says this will be resolved by https://github.com/projecthydra/sufia/pull/1357

@jcoyne I would maybe write a test if i could get tests passing for sufia but it's a losing battle for me right now and I'm demoralized. If that is a blocker y'all can close the PR as will not merge.

mjgiarlo:
@hackmastera #1357 is merged. Could you rebase your PR? I'd love to get this merged.

hackmastera:
@mjgiarlo it's ready

PR #1357, opened 2015-11-23
Ignore visibility of files[] input, it's hidden from Capybara

awead:
The files input is hidden from Capybara due to something in the css `fileinput-button` class. This gets around the problem by ignoring the visibility issue. The button is still visible with a screen-reader, so I'm not sure why it's invisible to Capybara.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -62,6 +62,7 @@ db/*.sqlite3
 db/*.sqlite3-journal
 log/*.log
 tmp
+.byebug_history
 
 *.sqlite3
 *.log
--- a/spec/features/ingest_upload_files_spec.rb
+++ b/spec/features/ingest_upload_files_spec.rb
@@ -15,8 +15,8 @@ describe "Uploading files via web form", type: :feature do
 
   context "the terms of service", :js do
     it "is required to be checked" do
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -98,3 +98,4 @@ public/system/
 
 ## Sitemap
 public/sitemap.xml
+.byebug_history
```
cam156: @awead this seems extraneous to the PR.

awead: @cam156 I can put it in a separate PR if you like.

cam156: @awead I'm curios why it is needed now.  Should it be part of a global git ignore instead of and individual project?

awead: makes sense.

jcoyne: Since the project requires byebug, I think it's okay to put .byebug_history in the .gitignore.

awead:
@cam156 ping

.gitignore
awead: @jcoyne I already amended the PR to remove that change because making a global gitignore made more sense, but I can push a change to .gitignore after this PR is merged.

awead:
@jcoyne ok. how about now?

PR #1360, opened 2015-11-24
Option to show collection facet in Dashboard and search results

elrayle:
Option: Sufia.config.collection_facet
Values:
* nil - do not show collection facets
* :user - show for logged in users
* :public - show for everyone (e.g. logged in and non-logged in users)

### DIFF
--- a/app/controllers/concerns/sufia/collections_controller_behavior.rb
+++ b/app/controllers/concerns/sufia/collections_controller_behavior.rb
@@ -30,6 +30,12 @@ module Sufia
       presenter
     end
 
+    def update
+      super
+      new_members = ActiveFedora::Base.find(params[:batch_document_ids])
+      new_members.each(&:update_index)
+    end
+
     protected
 
       def presenter
--- a/app/helpers/facets_helper.rb
+++ b/app/helpers/facets_helper.rb
@@ -1,0 +1,13 @@
+module FacetsHelper
+  include Blacklight::FacetsHelperBehavior
+
+  def display_collection_facet
+    if Sufia.config.collection_facet == :public
+      return true
+    elsif Sufia.config.collection_facet == :user && current_user
+      return true
+    else
+      return false
+    end
+  end
+end
--- a/lib/generators/sufia/templates/catalog_controller.rb
+++ b/lib/generators/sufia/templates/catalog_controller.rb
@@ -57,6 +57,7 @@ class CatalogController < ApplicationController
     config.add_facet_field solr_name("based_near", :facetable), label: "Location", limit: 5
     config.add_facet_field solr_name("publisher", :facetable), label: "Publisher", limit: 5
     config.add_facet_field solr_name("file_format", :facetable), label: "File Format", limit: 5
+    config.add_facet_field solr_name("collection", :facetable), label: "Collection", helper_method: :collection_name, limit: 5, show: :display_collection_facet
 
     # Have BL send all facet field names to Solr, which has been the default
     # previously. Simply remove these lines if you'd rather use Solr request
--- a/spec/features/collection_facet_spec.rb
+++ b/spec/features/collection_facet_spec.rb
@@ -1,0 +1,118 @@
+require 'spec_helper'
+
+describe "Collection Facet", type: :feature do
+  def visit_dashboard_my_files
+    visit '/dashboard/files'
+    expect(page.title).to eq 'Files listing'
+    expect(page).to have_link('My Files')
+    expect(page).to have_link('My Collections')
+    expect(page).to have_link('Fake PDF Title')
+    expect(page).to have_css('h3', text: 'Filter your files')
+  end
+
+  def search_for_files
+    visit '/'
+    fill_in('search-field-header', with: 'pdf')
+    click_button('search-submit-header')
+    expect(page.title).to eq 'Keyword: pdf - Blacklight Search Results'
+    expect(page).to have_link('Fake PDF Title')
+    expect(page).to have_css('h4', text: 'Limit your search')
+  end
+
+  def facet?(page)
+    f = page.all('a', text: /\ACollection\z/)
+    return true if f.size == 1
+    false
+  end
+
+  let(:auser) { FactoryGirl.create(:curator) }
+
+  let!(:fixtures) do
+    create_file_fixtures(auser.user_key, [:public_pdf]) do |f|
+      f.tag = ['pdf']
+      f.apply_depositor_metadata(auser.user_key)
+    end
+  end
+
+  let(:collection1) do
+    Collection.create(title: 'Test Collection 1', description: 'Description for collection 1',
+                      members: []) { |c| c.apply_depositor_metadata(auser.user_key) }
+  end
+
+  context "with collection having a member" do
+    before do
+      collection1.add_members([fixtures[0].id])
+      collection1.save
+      fixtures[0].update_index
+    end
+
+    context "and collection_facet config == nil" do
+      before do
+        Sufia.config.collection_facet = nil
+      end
+
+      it "does NOT show collection facet in Dashboard -> My Files" do
+        login_as auser
+        visit_dashboard_my_files
+        expect(facet?(page)).to eq false
+      end
+
+      it "does NOT show collection facet in pubic search with user logged in" do
+        login_as auser
+        search_for_files
+        expect(facet?(page)).to eq false
+      end
+
+      it "does NOT show collection facet in pubic search with user NOT logged in" do
+        search_for_files
+        expect(facet?(page)).to eq false
+      end
+    end
+
+    context "and collection_facet config == :user" do
+      before do
+        Sufia.config.collection_facet = :user
+      end
+
+      it "shows collection facet in Dashboard -> My Files" do
+        login_as auser
+        visit_dashboard_my_files
+        expect(facet?(page)).to eq true
+      end
+
+      it "does show collection facet in pubic search with user logged in" do
+        login_as auser
+        search_for_files
+        expect(facet?(page)).to eq true
+      end
+
+      it "does NOT show collection facet in pubic search with user NOT logged in" do
+        search_for_files
+        expect(facet?(page)).to eq false
+      end
+    end
+
+    context "and collection_facet config == :public" do
+      before do
+        Sufia.config.collection_facet = :public
+      end
+
+      it "shows collection facet in Dashboard -> My Files" do
+        login_as auser
+        visit_dashboard_my_files
+        expect(facet?(page)).to eq true
+      end
+
+      it "shows collection facet in pubic search with user logged in" do
+        login_as auser
+        search_for_files
+        expect(facet?(page)).to eq true
+      end
+
+      it "shows collection facet in pubic search with user NOT logged in" do
+        search_for_files
+        expect(facet?(page)).to eq true
+      end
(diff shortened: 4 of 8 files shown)


elrayle:
Following instructions in hydra-collections, adding a collection facet involves adding...
```
index_collection_ids(solr_doc)
```
to add the collection information to the solr doc that is used by the collection facet.  This line produces a deprecation warning.  Is there replacement code that does the same thing?  Or is there a different approach to take for adding facets in the next release?

Reference:
* [hydra-collections README#make-your-models-collectible](https://github.com/projecthydra/hydra-collections/blob/5.0-stable/README.md#make-your-models-collectible)
* [hydra-collections README#make-the-collection-show-as-a-facet-in-your-catalogcontroller](https://github.com/projecthydra/hydra-collections/blob/5.0-stable/README.md#make-the-collection-show-as-a-facet-in-your-catalogcontroller)

mjgiarlo:
@elrayle We're getting very close to merging sufia-core back into sufia. Is it :ok: if we wait on this PR until that has happened, or is this urgent?

Thanks for the PR!

jcoyne:
Is this going to show the collections that I create on your search results?  It seems like that could get very confusing. Especially if you and I both have collections with the same names.

elrayle:
@mjgiarlo What is the timing for the sufia-core merge?

elrayle:
@jcoyne It behaves as follows:
* Dashboard -> My Files -- only shows your own files and therefore, only your collections in the facets
* Home -> Search -> Search Results -- shows all public files which may come from multiple users who may have named their collections the same.  If there are two collections with the same name, both will appear in the Collection facet list and will show the same name.  This is the same whether the user is logged in or not.  To find a collection by a specific user, you can limit with the Creator facet first and then limit to a collection with the collection facet.  Since Sufia does not enforce unique named collections across users, this is probably the best we can do, but I am open to suggestions.

elrayle:
@jcoyne I imagine that some sights will limit this facet to logged in users only (:user).  And some sites may control collection names more closely and want to allow public users (:public) to have this facet.

I am hoping to add in the concept of public/private collections as well, but that is for another day.

mjgiarlo:
@elrayle :speech_balloon: 
> @mjgiarlo What is the timing for the sufia-core merge?

It's already started and I'm guessing given the short week will be ready next week? Key word: guessing. @jcoyne is taking point on this, and I'm not sure what his schedule and priorities are for next week.

elrayle:
@mjgiarlo Is the merge of sufia-core going to become Sufia 6.5?  Will it be substantially different?

mjgiarlo:
@elrayle The merger of sufia-core into Sufia master will become Sufia 7.0.0 and will include CurationConcerns, Hydra::Works, Hydra::PCDM, and all that jazz. It will be substantially different in terms of the code and modeling, yes.

elrayle:
@mjgiarlo Then I would hope to get this in before that push and make a release 6.4.1 since there are very few changes, or we could call it 6.5 if you like.  We are about to complete our Sufia project that is currently based on 6.4 and hoped to get this into Sufia as opposed to maintaining it separately in our code.  I am happy to work on bringing forward the collection changes I've been making in Sufia 6.x into the Sufia 7.

elrayle:
@mjgiarlo I am somewhat out of the loop at this point.  What is the state of Sufia 7?  Will it be a fully functioning Sufia including the concept of works?

mjgiarlo:
@elrayle Sufia 7 will be all of Sufia 6's stuff + CurationConcerns + PCDM/Works-based models, yep. There will be a gap between when sufia-core merges into Sufia's master branch and when 7.0.0 is released, though, because we have yet to do much testing at the UI level, and we haven't yet seen the output of the Sufia UI WG's work on updating the wireframes.

jcoyne:
`index_collection_ids` is deprecated because it doesn't scale well. If I recall correctly when you get beyond a few dozen collections it gets unbearably slow. It's original purpose was to allow you to display the collection members on a show page, we now do that with join queries.. It was not intended for adding collection facets.

mjgiarlo:
@elrayle It looks like this PR has some bugs in it related to how you're calling `AF::Base#find`, judging by Travis.

Regarding the intent behind this PR, does Cornell currently have a need for both the `:user` and `:public` config values?

elrayle:
The errors are caused by two code changes...

1) generic_file_indexing_service.rb - addition of `object.index_collection_ids(solr_doc)` causes two of the failure in collection_controller_spec.rb where the collection id was expected, but nil was returned.  This is required to add the information to the solr doc that hydra-collections uses for collection facets.

2) collection_controller_behavior.rb - override of update method to call `#update_index` for each member that is changed.  The errors from this change get `ArgumentError: Couldn't find ActiveFedora::Base without an ID`.  This is required to update the solr doc when a file is added to a collection via Dashboard -> My Files -> bookmark -> Add to Collection.

I am exploring the errors more to determine whether the code is causing a real problem or the tests should change.

elrayle:
@jcoyne What is the recommended way to get the information into the solr doc to allow for collection faceting?  I based my code on the hydra-collections README (which is now the [classic README](https://github.com/projecthydra/hydra-collections/blob/5.0-stable/README.md)).

mjgiarlo:
@elrayle this has generated a lot of discussion here and on IRC. How about we discuss this on next week's Hydra Tech call? (Dec. 2nd)

elrayle:
@mjgiarlo At Cornell, initially we will use :user because there is no concept of private collections and we have collections that are used for internal organization.  Eventually, we will want to use :public and have public/private collections such that the public users (i.e. non-logged in users) will only see public collections and logged-in users will see all collections (or collections for which they have permissions).

jcoyne:
@elrayle I'm pretty strongly against faceting on collections. As a user I would be confused by seeing collections that you created in the facet.  I would potentially be offended by seeing names that some people picked for their collections (these aren't moderated).  I would be further confused if multiple collections had the same name, especially if one of them was created by me.  I think these things all need to be ironed out before pushing this out to everyone.  It seems to me like you might want to have AdminSets instead. These would be managed by a select group.  Since they would be curated, it would eliminate the confusion.

mjgiarlo:
@elrayle In that case, I wonder if we should consider allowing access controls on collections rather than having this exceptional method of handling visibility of a certain facet.

For historical context, the team at PSU didn't build access controls into collections not for architectural reasons but because we needed to be expedient during that sprint. Copying them here for their perspective now: @cam156 @awead @hectorcorrea @mtribone @DanCoughlin

elrayle:
I can make these changes in our local app, but it seems from conversations I've had, that other institutions are also interested in beefing up the functionality of collections so that they fit more of a mediated deposit IR instead of a self-deposit IR.  Our app has curators who manage the collections and items in the collections.  Collections can have thousands to hunderds of thousands of items.

In our local app, I have added a simple workflow.  Long term I can add this to Sufia as well if desired, but wanted to shake it up in our local app first.  We may be tweaking the design based on feedback from our metadata team.

Making these changes controllable via configuration makes sense so that users of Sufia can choose the type of IR that makes works for them.

mjgiarlo:
@elrayle Yes, indeed, mediated deposit has been identified as the Sufia community's top need after support for multi-file works. I'm not sure what the needs are around collections in a pure mediated deposit context. The terminology here is way overloaded (not your fault!). Could you take a look at the following document and comment on which notion seems to fit Cornell's notion of collections best? https://wiki.duraspace.org/display/hydra/Collections%2C+Admin+Sets%2C+Display+Sets

This is related to some discussions happening in the Sufia UI WG, so I'm pinging @mtribone so he can follow this too.

awead:
I'm still a little fuzzy on the user case. It sounds like you just want a facet for collection name. That's a one-line addition to CatalogController. We did some fancy things with collections in Archivesphere, such as searching within a collection, and have done similar things in the dashboard to display different sets of information. All of this can be accomplished by adding methods to the search builder in Blacklight which will pass different filter queries to Solr. I'm not convinced we need configuration options in order to do it. However, if there are a lot of folks in the community that need this sort of thing, then I'm all for it.

Another option would be to add the relevant Solr parts to `Hydra::Collections::SearchBehaviors` or another new module in that gem and mix them into Sufia from there.

elrayle:
@awead From the [hydra-collections README#make-the-collection-show-as-a-facet-in-your-catalogcontroller](https://github.com/projecthydra/hydra-collections/blob/5.0-stable/README.md#make-the-collection-show-as-a-facet-in-your-catalogcontroller), it sounds like all you need to do is add one line to the catalog controller, but in actuality, that is not sufficient.  

Additional  changes required:
* modify the indexer code to index the collections in the solr doc according to instructions at [hydra-collections README#make-your-models-collectible](https://github.com/projecthydra/hydra-collections/blob/5.0-stable/README.md#make-your-models-collectible)
* modify sufia code to regenerate the solr doc when changes are made.  More specifically, adding a file to a collection through Dashboard -> My Files -> bookmark -> Add to Collection fails to regenerate the solr doc.
* The only other change was to add the flexibility of a site to choose whether they wanted collection facets not at all, only when logged in, or for everyone including the non-logged in users.  This was put in place in recognition that not all sites will have the exact same use case that we have.

If there is an alternate recommended way to accomplish this, then perhaps someone who knows that approach can update the documentation.

elrayle:
@mjgiarlo I read the document [User Collections, Admin Sets, Display Sets](https://wiki.duraspace.org/x/IIbVAw) and better understand the terminology being used in this discussion and on IRC.  I will update this document with our use case.

This PR is a first step toward having that kind of control over "Collections" as a User Collection, Admin Set, or Display Set.  It is an imperfect implementation, but it was do-able without making major changes to the architecture to do a full implementation of the various collection types.

As far as I can see, collections currently have the following behaviors.
* all collections are public and private
 * private because only the person who creates a collection will see that collection in My Collections and only the person who creates the collection will see that collection in the list of collections when adding files to a collection
 * public because if you do a general search and the search terms match a collection the user doesn't own, that collection will show up in the search results (current behavior) and if you add a collection facet and do a search, collections not owned by the current user can be part of the faceted collections.  The search results inclusion of collections is the same whether the user is logged in or not.

It would be an excellent addition to add permissions to the collections that are consistent with the current permissions for files... private, public, limited to users/groups/institution. In our use case, we  only use collection facet for :user because we have a mix of collections that operate as Admin Sets and Display Sets with no way to distinguish between them in the current implementation.  

With permissions, we would define...
* User Sets - any private collection
* Admin Sets - any shared collection (i.e. limited to users/groups/institutions)
* Display Sets - any public collection

In an ideal world, there would be roles that control what types of collections a user can create.  If the user has enough privileges, the user can promote a set from User Set to Admin or Display and from Admin to Display.  The use case would be that the curator(s) work on a set as an Admin Set and when it is complete, they promote it to a Display Set which the public can see.

With permissions, we would enable facet collections for all users and expect that logged-in users would see their User Sets and Admin Sets that they have permission to see and all users would see Display Sets.

elrayle:
@jcoyne The problem you describe already exists.  To reproduce:
* login as user A and create a collection with name Test and add a few files with keyword=TEST
* login as user B and create a collection with name Test and add a few files with keyword=TEST
* logout and search for test

You will see collection Test in the results twice for user A's Test collection and user B's Test collection along with the files you added with keyword=TEST.

elrayle:
Moving this PR to be against 6.0-stable.

PR #1363, opened 2015-11-25
Option to show collection facet in Dashboard and search results

elrayle:
Option: Sufia.config.collection_facet
Values:
* nil - do not show collection facets
* :user - show for logged in users
* :public - show for everyone (e.g. logged in and non-logged in users)

### DIFF
--- a/app/controllers/concerns/sufia/collections_controller_behavior.rb
+++ b/app/controllers/concerns/sufia/collections_controller_behavior.rb
@@ -30,6 +30,16 @@ module Sufia
       presenter
     end
 
+    def create
+      super
+      update_members_indices
+    end
+
+    def update
+      super
+      update_members_indices
+    end
+
     protected
 
       def presenter
@@ -76,6 +86,13 @@ module Sufia
 
       def _prefixes
         @_prefixes ||= super + ['catalog']
+      end
+
+      def update_members_indices
+        return if Sufia.config.collection_facet.nil?
+        Array(params[:batch_document_ids]).each do |gf|
+          Sufia.queue.push(ResolrizeGenericFileJob.new(gf))
+        end
       end
   end
 end
--- a/app/helpers/facets_helper.rb
+++ b/app/helpers/facets_helper.rb
@@ -1,0 +1,13 @@
+module FacetsHelper
+  include Blacklight::FacetsHelperBehavior
+
+  def display_collection_facet
+    if Sufia.config.collection_facet == :public
+      true
+    elsif Sufia.config.collection_facet == :user && current_user
+      true
+    else
+      false
+    end
+  end
+end
--- a/sufia-models/app/services/sufia/generic_file_indexing_service.rb
+++ b/sufia-models/app/services/sufia/generic_file_indexing_service.rb
@@ -13,6 +13,7 @@ module Sufia
         # between files on disk (in fcrepo.binary-store-path) and objects
         # in the repository.
         solr_doc[Solrizer.solr_name('digest', :symbol)] = digest_from_content
+        object.index_collection_ids(solr_doc) unless Sufia.config.collection_facet.nil?
       end
     end
 
--- a/sufia-models/lib/generators/sufia/models/collection_facet_config_generator.rb
+++ b/sufia-models/lib/generators/sufia/models/collection_facet_config_generator.rb
@@ -1,0 +1,24 @@
+require 'rails/generators'
+
+class Sufia::Models::CollectionFacetConfigGenerator < Rails::Generators::Base
+  source_root File.expand_path('../templates', __FILE__)
+
+  desc """
+This Generator makes the following changes to your application:
+  1. Updates existing sufia.rb initializer to include a collection_facet configuration
+       """
+
+  def banner
+    say_status("info", "ADDING COLLECTION_FACET OPTION TO SUFIA CONFIG", :blue)
+  end
+
+  def inject_config_initializer
+    inject_into_file 'config/initializers/sufia.rb', before: "# Where to store tempfiles, leave blank for the system temp directory (e.g. /tmp)" do
+      "  # Add a collection facet to search results.  Possible values are...\n" +
+        "  #   nil (default) - do not include collection facet\n" +
+        "  #   :user - show for logged in users\n" +
+        "  #   :public - show for everyone (e.g. logged in and non-logged in users)\n" +
+        "  # config.collection_facet = nil\n"
+    end
+  end
+end
--- a/sufia-models/lib/generators/sufia/models/templates/config/sufia.rb
+++ b/sufia-models/lib/generators/sufia/models/templates/config/sufia.rb
@@ -95,6 +95,12 @@ Sufia.config do |config|
   # Default is false
   # config.citations = false
 
+  # Add a collection facet to search results.  Possible values are...
+  #   nil (default) - do not include collection facet
+  #   :user - show for logged in users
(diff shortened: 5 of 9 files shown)


app/controllers/concerns/sufia/collections_controller_behavior.rb
```diff
@@ -30,6 +30,16 @@ def show
       presenter
     end
 
+    def create
+      super
+      ActiveFedora::Base.find(params[:batch_document_ids]).each(&:update_index) unless params[:batch_document_ids].nil?
```
mjgiarlo: I'm concerned about performance implications here. Loading a bunch of objects from Fedora, one at a time, and then calling `#update_index` on each seems like it will hang up creates and updates. Perhaps this should be done in a background job. @jcoyne what do you think?

app/controllers/concerns/sufia/collections_controller_behavior.rb
```diff
@@ -30,6 +30,16 @@ def show
...
+    end
+
+    def update
+      super
+      ActiveFedora::Base.find(params[:batch_document_ids]).each(&:update_index) unless params[:batch_document_ids].nil?
```
mjgiarlo: Since this method is identical in both `#create` and `#update`, could you extract it to a private method and call it from both places?

app/helpers/facets_helper.rb
```diff
@@ -0,0 +1,13 @@
+module FacetsHelper
+  include Blacklight::FacetsHelperBehavior
+
+  def display_collection_facet
+    if Sufia.config.collection_facet == :public
+      return true
```
mjgiarlo: Can you drop the explicit `return`s in this method?

sufia-models/app/services/sufia/generic_file_indexing_service.rb
```diff
@@ -13,6 +13,7 @@ def generate_solr_document
         # between files on disk (in fcrepo.binary-store-path) and objects
         # in the repository.
         solr_doc[Solrizer.solr_name('digest', :symbol)] = digest_from_content
+        object.index_collection_ids(solr_doc)
```
mjgiarlo: Should we bother doing this if `config.collection_facet` is set to `nil`?

sufia-models/lib/generators/sufia/models/collection_facet_config_generator.rb
```diff
@@ -0,0 +1,26 @@
...
+  end
+
+  def inject_config_initializer
+    inject_into_file 'config/initializers/sufia.rb', before: "# Where to store tempfiles, leave blank for the system temp directory (e.g. /tmp)" do
+      "  # Add a collection facet to search results.  Possible values are...\n" +
+        "  #   nil (default) - do not include collection facet\n" +
+        "  #   :user - show for logged in users\n" +
+        "  #   :public - show for everyone (e.g. logged in and non-logged in users)\n" +
+        "  config.collection_facet = nil\n" +
+        "  # config.collection_facet = :user\n" +
```
mjgiarlo: Can you drop this line and the next? I think the documentation provided above is good enough as is, myself.

sufia-models/lib/generators/sufia/models/templates/config/sufia.rb
```diff
@@ -95,6 +95,14 @@
   # Default is false
   # config.citations = false
 
+  # Add a collection facet to search results.  Possible values are...
+  #   nil (default) - do not include collection facet
+  #   :user - show for logged in users
+  #   :public - show for everyone (e.g. logged in and non-logged in users)
+  # config.collection_facet = nil
+  # config.collection_facet = :user
```
mjgiarlo: Can you drop this line and the next? I think the documentation provided above is good enough as is, myself.

sufia-models/app/services/sufia/generic_file_indexing_service.rb
elrayle: I originally did not do this unless config.collection_facet had a value other than nil.  I can return to that.  What it buys you if it is always indexed is you can change from nil to :public or :user and the information for the facets is already there.  But since @jcoyne pointed out that this method is expensive, maybe it should only be there when facets are being used.

elrayle:
More discussion at PR https://github.com/projecthydra/sufia/pull/1360.

All requested changes have been made except converting indexing code into a background job.

elrayle:
DO NOT MERGE YET... testing indexing as background job.

elrayle:
Made re-indexing of generic files effected by adding them to a collection run in background.

Tests pass.  Need final review and merge.

app/controllers/concerns/sufia/collections_controller_behavior.rb
```diff
@@ -77,5 +87,9 @@ def form_class
       def _prefixes
         @_prefixes ||= super + ['catalog']
       end
+
+      def update_members_indices
+        params[:batch_document_ids].each { |gf| Sufia.queue.push(ResolrizeGenericFileJob.new(gf)) } unless params[:batch_document_ids].nil? || Sufia.config.collection_facet.nil?
```
mjgiarlo: This is low priority, but would you consider reformatting this a bit using a guard clause and a coercion to an array?

```ruby
return if Sufia.config.collection_facet.nil?
Array(params[:batch_document_ids]).each do |gf|
  Sufia.queue.push(ResolrizeGenericFileJob.new(gf))
end
```

sufia-models/app/services/sufia/generic_file_indexing_service.rb
```diff
@@ -13,6 +13,7 @@ def generate_solr_document
         # between files on disk (in fcrepo.binary-store-path) and objects
         # in the repository.
         solr_doc[Solrizer.solr_name('digest', :symbol)] = digest_from_content
+        object.index_collection_ids(solr_doc) unless Sufia.config.collection_facet.nil?
```
mjgiarlo: Should this also be in the job, or is there a good reason to keep it out?

app/controllers/concerns/sufia/collections_controller_behavior.rb
elrayle: Done.

sufia-models/app/services/sufia/generic_file_indexing_service.rb
elrayle: Not sure how this would become part of the background job.  The background job calls update_index which ends up calling the indexing service which calls object.index_collection_ids.  The background job itself does not have access to the solr_doc.  Do you have a suggestion for how to do this differently?

mjgiarlo: Can you do this inside the job:

```ruby
def run
  generic_file.index_collection_ids(generic_file.to_solr)
  generic_file.update_index
end
```

:question:

elrayle: This code does not work.  The fields added by index_collection_ids are not added to the solr doc when called in this way.

Also, there are at least two paths for adding a file to a collection and they don't both go through the CollectionControllerBehavior code.  When adding from Dashboard, this code is used such that the collection is causing the reindex.  When adding through Batch Upload, each generic file is indexed and the collection update code is never called.  So even if the proposed code worked, taking index_collection_ids out of the solr indexer code would prevent generic files from having the proper fields any time they are indexed from anywhere besides a collection update.

mjgiarlo: Ah, good catch. Thanks, @elrayle

PR #1391, opened 2015-12-07
Create a default work for an uploaded file

awead:
This is up for discussion. This fixes the issue with uploading files, but leaves a couple of glaring problems. The larger question is what do we do when files are uploaded without a work.

### DIFF
--- a/app/actors/sufia/file_set_actor.rb
+++ b/app/actors/sufia/file_set_actor.rb
@@ -1,0 +1,18 @@
+module Sufia
+  class FileSetActor < CurationConcerns::FileSetActor
+    # Overrides the method in CurationConcerns::FileSetActor so that if a nil work
+    # is passed, a default work is create instead.
+    def create_metadata(upload_set_id, work, file_set_params = {})
+      work ||= default_work
+      super
+    end
+
+    private
+
+      def default_work
+        GenericWork.create(title: ['Default title']) do |w|
+          w.apply_depositor_metadata(user)
+        end
+      end
+  end
+end
--- a/app/controllers/concerns/sufia/files_controller_behavior.rb
+++ b/app/controllers/concerns/sufia/files_controller_behavior.rb
@@ -96,5 +96,14 @@ module Sufia
         return {} if params[:file_set].blank?
         super
       end
+
+      def find_parent_by_id
+        return if parent_id.empty?
+        super
+      end
+
+      def actor
+        @actor ||= ::Sufia::FileSetActor.new(@file_set, current_user)
+      end
   end
 end
--- a/app/views/curation_concerns/file_sets/upload/_form_fields.html.erb
+++ b/app/views/curation_concerns/file_sets/upload/_form_fields.html.erb
@@ -12,14 +12,14 @@
                 <span class="btn btn-primary fileinput-button">
                     <i class="glyphicon glyphicon-plus" aria-hidden="true"></i>
                     <span aria-hidden="true">Select files...</span>
-                    <input type="file" name="files[]" multiple />
+                    <input type="file" name="file_set[files][]" multiple />
                 </span>
                 <% ua = request.env['HTTP_USER_AGENT'] %>
                 <% if !!(ua =~ /Chrome/) %>
                 <span class="btn btn-primary fileinput-button">
                     <i class="glyphicon glyphicon-plus"  aria-hidden="true"></i>
                     <span aria-hidden="true">Select folder...</span>
-                    <input type="file" name="files[]" directory webkitdirectory mozdirectory />
+                    <input type="file" name="file_set[files][]" directory webkitdirectory mozdirectory />
                 </span>
                 <% end %>
                 <div id="main_upload_start_span" class="activate-container visible-all-inline-block" data-toggle="tooltip" data-title="<%= t('sufia.upload_tooltip') %>">
--- a/spec/actors/file_set_actor_spec.rb
+++ b/spec/actors/file_set_actor_spec.rb
@@ -1,0 +1,24 @@
+require 'spec_helper'
+
+describe Sufia::FileSetActor do
+  include ActionDispatch::TestProcess
+
+  let(:user) { create(:user) }
+  let(:file_set) { create(:file_set) }
+  let(:actor) { described_class.new(file_set, user) }
+
+  describe 'creating metadata and content' do
+    let(:date_today)    { DateTime.now }
+    let(:upload_set_id) { "upload_id" }
+    subject { file_set.reload }
+    context "when no work is provided" do
+      before do
+        allow(DateTime).to receive(:now).and_return(date_today)
+        actor.create_metadata(upload_set_id, nil)
+      end
+      it "assigns a default one" do
+        expect(subject.generic_works.first.title).to contain_exactly("Default title")
+      end
+    end
+  end
+end
--- a/app/views/curation_concerns/file_sets/upload/_form.html.erb
+++ b/app/views/curation_concerns/file_sets/upload/_form.html.erb
@@ -6,6 +6,6 @@
     </div>
   <% end %>
   <div class="well">
-    <%= render 'curation_concerns/file_sets/upload/form_fields' %>
+    <%= render partial: 'curation_concerns/file_sets/upload/form_fields' %>
   </div>
 <% end %>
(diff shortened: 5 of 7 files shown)


app/actors/sufia/file_set_actor.rb
```diff
@@ -0,0 +1,19 @@
...
+
+    def create_metadata(upload_set_id, work, file_set_params = {})
+      work ||= default_work
+      super
+    end
+
+    private
+
+      def default_work
+        GenericWork.create.tap do |w|
```
jcoyne: The `.tap` is unnecessary here.

app/actors/sufia/file_set_actor.rb
```diff
@@ -0,0 +1,19 @@
...
+          w.apply_depositor_metadata(user)
+          w.title = ["Default title"]
+          w.save
```
jcoyne: `.save` is unnecessary in create

app/actors/sufia/file_set_actor.rb
```diff
@@ -0,0 +1,19 @@
...
+      work ||= default_work
+      super
+    end
+
+    private
+
+      def default_work
+        GenericWork.create.tap do |w|
+          w.apply_depositor_metadata(user)
+          w.title = ["Default title"]
```
jcoyne: pass `title: ['Default title']` as an argument to `create()`

mjgiarlo: You can also apply the title in the call to `#create` and save a line here:

```ruby
GenericWork.create(title: ['Default title']) do |w|
  # ...
end
```

app/actors/sufia/file_set_actor.rb
mjgiarlo: No need for the `#save` since you're in a `#create` block.

app/views/curation_concerns/file_sets/upload/_form_fields.html.erb
```diff
@@ -12,14 +12,14 @@
                 <span class="btn btn-primary fileinput-button">
                     <i class="glyphicon glyphicon-plus" aria-hidden="true"></i>
                     <span aria-hidden="true">Select files...</span>
-                    <input type="file" name="files[]" multiple />
```
jcoyne: I think we can remove this view entirely and we'll use https://github.com/projecthydra-labs/curation_concerns/blob/master/app/views/curation_concerns/file_sets/upload/_form_fields.html.erb instead.

app/controllers/concerns/sufia/files_controller_behavior.rb
```diff
@@ -96,5 +96,17 @@ def file_set_params
         return {} if params[:file_set].blank?
         super
       end
+
+      def find_parent_by_id
```
mjgiarlo: I might prefer this as a guard clause to make explicit there's some errant behavior you're protecting against, after which you do processing, like so:

```ruby
return if parent_id.empty?
ActiveFedora::Base.find(parent_id)
```

But I wouldn't hold the PR over that alone.

app/controllers/concerns/sufia/files_controller_behavior.rb
```diff
@@ -96,5 +96,17 @@ def file_set_params
...
+        if parent_id.empty?
+          nil
+        else
+          ActiveFedora::Base.find(parent_id)
```
jcoyne: call `super`?

app/controllers/concerns/sufia/files_controller_behavior.rb
```diff
@@ -96,5 +96,17 @@ def file_set_params
         return {} if params[:file_set].blank?
         super
       end
+
+      def find_parent_by_id
+        if parent_id.empty?
```
jcoyne: rephrase to `return if parent_id.empty?`

app/actors/sufia/file_set_actor.rb
```diff
@@ -0,0 +1,19 @@
+module Sufia
+  class FileSetActor < CurationConcerns::FileSetActor
+
+    def create_metadata(upload_set_id, work, file_set_params = {})
```
jcoyne: Document the difference between this method and the superclass.

spec/actors/file_set_actor_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+    let(:date_today) { DateTime.now }
+
+    before do
+      allow(DateTime).to receive(:now).and_return(date_today)
+    end
+
+    before do
+      expect(CharacterizeJob).to receive(:perform_later)
+      expect(IngestFileJob).to receive(:perform_later).with(file_set.id, /world\.png$/, 'image/png', user.user_key)
+      allow(actor).to receive(:acquire_lock_for).and_yield
```
jcoyne: Is there a reason you stub this out?

spec/actors/file_set_actor_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+      actor.create_metadata(upload_set_id, work)
+      actor.create_content(uploaded_file)
```
jcoyne: Is it necessary to test `create_content`?  We haven't made any changes there.

jcoyne:
@awead there are a few details to fix, but this looks like the correct approach to me.

app/views/curation_concerns/file_sets/upload/_form_fields.html.erb
awead: @jcoyne hmmm. it's blowing up here:
https://github.com/projecthydra-labs/curation_concerns/blob/master/app/views/curation_concerns/file_sets/upload/_form_fields.html.erb#L4
with `undefined local variable or method `upload_set_id'`

app/actors/sufia/file_set_actor.rb
```diff
@@ -0,0 +1,19 @@
...
+      work ||= default_work
+      super
+    end
+
+    private
+
+      def default_work
+        GenericWork.create.tap do |w|
+          w.apply_depositor_metadata(user)
```
jcoyne: what about setting upload_set_id.  I think that will be required if we want to batch apply metadata, right?

app/views/curation_concerns/file_sets/upload/_form_fields.html.erb
jcoyne: we need to pass the upload_set_id to it as we do here: https://github.com/projecthydra-labs/curation_concerns/blob/64eab4d87fac8b88367ebb5e34b4d34ad7ce6716/app/views/curation_concerns/file_sets/upload/_form.html.erb

app/actors/sufia/file_set_actor.rb
awead: @jcoyne GenericWork doesn't have an upload_set_id defined on it, so we'd need to add one.

app/views/curation_concerns/file_sets/upload/_form_fields.html.erb
awead: @jcoyne the curation_concerns form_fields lacks the terms of service checkbox.

jcoyne: @awead ah! good catch.

app/actors/sufia/file_set_actor.rb
```diff
@@ -0,0 +1,18 @@
...
+    def create_metadata(upload_set_id, work, file_set_params = {})
+      work ||= default_work
+      super
+    end
+
+    private
+
+      def default_work
+        GenericWork.create(title: ['Default title']) do |w|
+          w.apply_depositor_metadata(user)
```
jcoyne: we will need to add upload_set_id here. Because the batch controller updates metadata on the works.

jcoyne: I suppose this could be a new ticket if you don't want to do this here.

awead:
@jcoyne I think this is ready to go. I've added an additional commit to redirect to the My Files view on the dashboard, for now. I've documented this in other tickets. See #1397 #1398 and projecthydra-labs/curation_concerns#541

jcoyne:
@awead I think we should merge https://github.com/projecthydra-labs/curation_concerns/pull/544 first and then set `upload_set_id` in the `FileSetActor`

jcoyne:
Okay, that work will be done as https://github.com/projecthydra/sufia/issues/1397

PR #1366, opened 2015-12-04
Makes upload_set_id available in the WorkShowPresenter. 

hectorcorrea:
Fixes https://github.com/projecthydra-labs/sufia-core/issues/389

PR #1395, opened 2015-12-08
Makes upload_set_id available in the WorkShowPresenter. 

hectorcorrea:
This PR replaces https://github.com/projecthydra/sufia/pull/1366 (the old one was from sufia-core but that project is now readonly and it was easier for me just to create a new PR with the same commits)

### DIFF
--- a/app/presenters/sufia/work_show_presenter.rb
+++ b/app/presenters/sufia/work_show_presenter.rb
@@ -2,6 +2,7 @@ module Sufia
   class WorkShowPresenter < ::CurationConcerns::WorkShowPresenter
     # delegate fields from Sufia::Works::Metadata to solr_document
     delegate :based_near, :depositor, :identifier, :resource_type, :tag, to: :solr_document
+    attr_reader :upload_set_id
 
     def tweeter
       user = ::User.find_by_user_key(depositor)
--- a/app/views/curation_concerns/file_sets/upload/_script_templates.html.erb
+++ b/app/views/curation_concerns/file_sets/upload/_script_templates.html.erb
@@ -1,8 +1,7 @@
-<!-- TODO: make the upload_set_id available on the presenter -->
-<% if false && presenter.upload_set_id %>
+<% if @presenter.upload_set_id %>
     <div class="alert alert-info hide" id="redirect-loc" type="text/x-jquery-tmpl">
     <%# Script redirects here once the uploads are complete %>
-    <%= CurationConcerns::FileSetsController.upload_complete_path(presenter.upload_set_id) %>
+    <%= CurationConcerns::FileSetsController.upload_complete_path(@presenter.upload_set_id) %>
     </div>
 <% end %>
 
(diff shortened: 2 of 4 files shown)


app/presenters/sufia/work_show_presenter.rb
```diff
@@ -2,6 +2,7 @@ module Sufia
   class WorkShowPresenter < ::CurationConcerns::WorkShowPresenter
     # delegate fields from Sufia::Works::Metadata to solr_document
     delegate :based_near, :depositor, :identifier, :resource_type, :tag, to: :solr_document
+    attr_reader :upload_set_id
```
jcoyne: Where is this value set?

app/views/curation_concerns/file_sets/upload/_script_templates.html.erb
```diff
@@ -1,8 +1,7 @@
-<!-- TODO: make the upload_set_id available on the presenter -->
-<% if false && presenter.upload_set_id %>
+<% if @presenter && @presenter.upload_set_id %>
```
jcoyne: You should never check for @presenter, if it isn't set there's a bug we should address.

app/presenters/sufia/work_show_presenter.rb
hectorcorrea: I don't know. I only added here because view `_script_templates.html.erb` (see below) references it.

app/views/curation_concerns/file_sets/upload/_script_templates.html.erb
hectorcorrea: This is the partial that calls it without a presenter: https://github.com/projecthydra/sufia/blob/master/app/views/curation_concerns/file_sets/_multiple_upload.html.erb#L3

jcoyne: @hectorcorrea yes, but that controller has set `@presenter`, right?  If it doesn't we want it to raise an error.  Every time this page is rendered we want to require a `@presenter`

hectorcorrea: I know this view is rendered when the user clicks Upload a file without a work (e.g. click upload form the homepage) and it does not seem that `@presenter` is set there. 

But there is a presenter set when a the user is viewing a work (i.e. GenericWork show).

jcoyne: Then the fix should be to create a presenter when uploading files without a work.

jcoyne: In sufia 6, the files_controller_behavior.rb the `#new` action set:
```
@batch_id = ActiveFedora::Noid::Service.new.mint
```
you should just make the new action create a presenter that has a upload_set_id

hectorcorrea: OK, I've set the presenter on the `app/controllers/concerns/sufia/files_controller_behavior.rb`

hectorcorrea:
This PR is not needed. The work that @awead did on PR #1391 takes care of this. In particular the changes to `app/views/curation_concerns/file_sets/upload/_script_templates.html.erb`

mjgiarlo:
See @hectorcorrea's comment.

Issue #1386, opened 2015-12-04
On create page set `@form.depositor`


mjgiarlo:
_From @jcoyne on December 3, 2015 18:11_

Presently it's `nil` at this page http://localhost:3000/concern/generic_works/new

_Copied from original issue: projecthydra-labs/sufia-core#391_

PR #1407, opened 2015-12-08
add depositor metadata to curation_concern in controller#new

grosscol:
This solution is a bit peculiar in calling `apply_depositor_metadata` from the `generic_works_controller#new` method instead of an actor.  I'm going to see if I can track down where the call to create a new generic_work or curation_concern instance is and see if that's a more appropriate place for the apply metadata call.

### DIFF
--- a/app/controllers/concerns/sufia/generic_works_controller_behavior.rb
+++ b/app/controllers/concerns/sufia/generic_works_controller_behavior.rb
@@ -15,6 +15,7 @@ module Sufia
     def new
       # TODO: move this to curation_concerns
       @form = form_class.new(curation_concern, current_ability)
+      curation_concern.depositor = (current_user.user_key)
       super
     end
 
--- a/spec/controllers/generic_works_controller_spec.rb
+++ b/spec/controllers/generic_works_controller_spec.rb
@@ -7,11 +7,16 @@ describe CurationConcerns::GenericWorksController do
   routes { Rails.application.routes }
 
   describe "#new" do
+    before { get :new }
     it "is successful" do
-      get :new
       expect(response).to be_successful
       expect(response).to render_template("layouts/sufia-one-column")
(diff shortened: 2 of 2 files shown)


jcoyne:
Please rebase this and then we'll see if the tests are passing.

jcoyne:
Can you explain the reason you are making this change in the commit message?

grosscol:
I would if I was aware of the reason depositor needed to be known to the form when `new` is called on the works controller.  My assumption had been that the depositor metadata would get applied by some actor before persisting the new work.  My guess that the ticket existed because someone needed to the depositor metadata to be displayed prior to the work be saved.

I'll look into it some more.

jcoyne:
@grosscol which ticket? This PR isn't linked to a ticket.

grosscol:
Updated the comment to reflect that this closes #1386.
The GenericWork is created by the call to `set_curation_concern_type GenericWork`. So it doesn't appear that there is a better place to apply the depositor metadata if it is needed by the form for the /new view.

app/controllers/concerns/sufia/generic_works_controller_behavior.rb
```diff
@@ -15,6 +15,7 @@ module GenericWorksControllerBehavior
     def new
       # TODO: move this to curation_concerns
       @form = form_class.new(curation_concern, current_ability)
+      curation_concern.apply_depositor_metadata(current_user)
```
jcoyne: Set `curation_concern.depositor=` instead of `apply_depositor_metadata` the latter creates a bunch of ACLs which aren't needed here.

grosscol: Will do.  I originally had it set like that, but only found `depositor=` used inside of `apply_depositor_metadata` so I thought I might have been missing other important attributes.

PR #1409, opened 2015-12-09
Don't hardcode emails into factories

jcoyne:


mjgiarlo:
Travis is angry and needs rebase.

Issue #1424, opened 2015-12-09
Sufia::FilesController versus CurationConcerns::FileSetsController


awead:
With the change from `Files` to `FileSets` in the curation_concerns gem, what is the different between the Sufia::FilesController and the CurationConcerns::FileSetsController? Is the latter supposed to replace the former? If not, what's the distinction between the two?

mjgiarlo:
Sufia's FilesControllerBehavior is Sufia-specific stuff that we inject into https://github.com/projecthydra/sufia/blob/master/app/controllers/curation_concerns/file_sets_controller.rb to make selective overrides and additions to the FileSetsController provided by CC. But that's not obvious by looking at the behavior in Sufia because there's no mention of CC in that file.

awead:
@mjgiarlo ok, that makes sense now, but isn't immediately obvious. Given the questions raised in #1427, I'm wondering if there's a way to make this clearer in the code. Renaming any instance of FilesController to FileSetsController might help, but also maybe some re-kajiggering of modules includes too?

mjgiarlo:
Renaming would be a good start. Re-kajiggering is OK by me too if we can make it the override a bit clearer.

awead:
My only thought on the re-kajiggering was moving the FileSetsController to app/controllers/file_sets_controller.rb so that:
``` ruby
class FileSetsController
  include CurationConcerns::FileSetsController
  include Sufia::Controller
  include Sufia::FileSetsControllerBehavior
end
```
As it stands now, it's burried under app/controllers/curation_concerns

mjgiarlo:
I thought it was in that directory intentionally, taking advantage of Ruby namespacing to effectively inject our concerns into CC's FileSetsController. There are other ways to do that of course.

awead:
Oh, ok, that makes sense then. I think I'm down with the further `Files` to `FileSets` renaming. I'll take that up. I think ideally, with also the Files versus Works distinction, there should be little occurrence of the word "file" or "files" in the final released gem.

mjgiarlo:
Agree that "file" in Sufia codebase should only be used in the pcdm:File context. OTOH, I think we should still label FileSets as "files" in the UI because our users have no idea what a FileSet is and (IMO) shouldn't need to. You probably meant this but I wanted to tease that out.

awead:
Yeah, good clarification!

grosscol:
[...]

PR #1427, opened 2015-12-09
Rename Sufia::FilesControllerBehavior to Sufia::FileSetsControllerBehavior

### LINKED ISSUE #1379: Should Sufia::FilesControllerBehavior be Sufia::FileSetsControllerBehavior
_From @grosscol on November 12, 2015 17:18_

It looks like the functionality there is being mixed into the FileSetsController.  Is this naming a hold over from the GenericFile history?

_Copied from original issue: projecthydra-labs/sufia-core#187_

awead:
fixes #1379

### DIFF
--- a/app/controllers/concerns/sufia/file_sets_controller_behavior.rb
+++ b/app/controllers/concerns/sufia/file_sets_controller_behavior.rb
@@ -5,7 +5,7 @@ module Sufia
     autoload :LocalIngestBehavior
     autoload :UploadCompleteBehavior
   end
-  module FilesControllerBehavior
+  module FileSetsControllerBehavior
     extend ActiveSupport::Concern
     include Sufia::Breadcrumbs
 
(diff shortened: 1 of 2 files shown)


app/controllers/concerns/sufia/file_sets_controller_behavior.rb
```diff
@@ -5,7 +5,7 @@ module FilesController
     autoload :LocalIngestBehavior
```
mjgiarlo: @awead I see that on line 2 we're still referencing a `FilesController` and we have some directory structure around this to pull in the other modules. Do you think this PR should expand to nuke any and all references to "FilesController" (or "files_controller") in the code, the filenames, and directory names? Or shall we merge this PR and ticket that as a follow-on?

mjgiarlo:
:+1: (though see my :question:)

app/controllers/concerns/sufia/file_sets_controller_behavior.rb
awead: @mjgiarlo yeah, I wondered about that too... this also prompted my question in #1424. I would go with a follow up ticket and perhaps more discussion on the issue above.

mjgiarlo: :ok_hand:

PR #1428, opened 2015-12-09
Removed files that moved to CurationConcerns

jcoyne:


mjgiarlo:
Sweet, sweet deletion!

Issue #1415, opened 2015-12-09
Uploading from work does not add file to work


cam156:

grosscol:
@awead noticed that when uploading a file, there was not a work to attach it to.  However, when `generic_work#new` is called, the `curation_concern` is a work, but it has not been persisted. The work's id is nil.  I think this is what was causing the observation that the work had not been created.

So what is happening now, is that a work exists when `generic_work#new` is called and another one is created after the file is uploaded.  I think the fix for this is to pass the work and not work.id.  That should avoid the logic that creates a work if one does not exist.

awead:
@grosscol from what page are you uploading?

grosscol:
generic_works/new

PR #1429, opened 2015-12-09
Sets work_id so that FileSets added are linked to the work

hectorcorrea:
Fixes https://github.com/projecthydra/sufia/issues/1415

### DIFF
--- a/app/controllers/concerns/sufia/generic_works_controller_behavior.rb
+++ b/app/controllers/concerns/sufia/generic_works_controller_behavior.rb
@@ -12,6 +12,11 @@ module Sufia
       layout "sufia-one-column"
     end
 
+    def show
+      @work_id = params[:id]
+      super
+    end
+
     def new
       # TODO: move this to curation_concerns
       @form = form_class.new(curation_concern, current_ability)


jcoyne:
Can you write a feature spec for this path that would have failed had you not made this change?

app/views/curation_concerns/file_sets/upload/_form_fields.html.erb
```diff
@@ -2,7 +2,7 @@
       <%= hidden_field_tag(:total_upload_size, 0) %>
       <%= hidden_field_tag(:relative_path) %>
       <%= hidden_field_tag(:upload_set_id, @upload_set_id) %>
-      <%= hidden_field_tag(:parent_id, @work_id) %>
+      <%= hidden_field_tag(:parent_id, @presenter.id) %>
```
jcoyne: I don't understand why having the hidden field is necessary. It should be passed in the route:
```
curation_concerns_file_sets POST   /concern/container/:parent_id/file_sets
```

app/controllers/concerns/sufia/file_sets_controller_behavior.rb
```diff
@@ -29,6 +29,7 @@ module FileSetsControllerBehavior
 
     def new
       @upload_set_id = ActiveFedora::Noid::Service.new.mint
+      @presenter = Sufia::WorkShowPresenter.new(::SolrDocument.new, nil)
```
grosscol: This presenter is used so that work_id isn't loaded from fedora, but is using solr instead?  Having the work presenter in the fileset controller seem incongruous.

app/views/curation_concerns/file_sets/upload/_form_fields.html.erb
hectorcorrea: The reason `@work_id` is required is because sometimes this view is rendered as part of Generic Work Show page and sometimes standalone (e.g. when adding files without works.) 

I've set the code to use `@work_id` again (rather than the presenter) but made sure the variable is set.

app/controllers/concerns/sufia/file_sets_controller_behavior.rb
hectorcorrea: Good point. I've updated the controller that calls this view to set the `@work_id` variable instead.

hectorcorrea:
I still need to write the darn test ;) but I think the updated PR is a better solution to the problem.

jcoyne:
I've solved this with #1443 and it has an integration test. Can we close this?

jcoyne:
Ack, maybe not. Lemme look again.

jcoyne:
Okay, I've added changes into #1443 such that the file gets attached to the work.

hectorcorrea:
I am not sure #1443 will fix the problem. 

Without passing a `parent_id` parameter on the post, the controller is not realizing what work the FileSets needs to be linked to. This is the line that depends on the `parent_id` 

https://github.com/projecthydra/sufia/blob/master/app/controllers/concerns/sufia/file_sets_controller_behavior.rb#L101

jcoyne:
@hectorcorrea that's taken care of here. https://github.com/projecthydra/sufia/pull/1443/files#diff-b945b4174189941b06bbeaa71a155615R4  parent_id is in the url

PR #1433, opened 2015-12-09
Redirect to upload_set#edit after upload

### LINKED ISSUE #1398: Redirect to appropriate batch upload page
When uploading files that have default works created for them, the page should redirect to a batch upload page where metadata can be added and subsequently applied to all the default works created during the upload process.

A workaround was introduced in 1995d19ff21ef6ee684a2c1748e50edab00aada0 which should be remove once this ticket is closed.

Relates to: projecthydra-labs/curation_concerns#541 and #1397

jcoyne:
Fixes #1398


app/views/curation_concerns/file_sets/upload/_script_templates.html.erb
```diff
@@ -1,8 +1,6 @@
-<!-- TODO: make the upload_set_id available on the presenter -->
 <div class="alert alert-info hide" id="redirect-loc" type="text/x-jquery-tmpl">
   <%# Script redirects here once the uploads are complete %>
-  <%# TODO: change this once multifile upload issues have been resolved %>
-  <%= sufia.dashboard_files_path %>
+  <%= CurationConcerns::FileSetsController.upload_complete_path(SecureRandom.uuid) %>
```
mjgiarlo: What's the random UUID for?

grosscol: This is where the upload_set_id gets generated?

app/controllers/concerns/sufia/file_sets_controller_behavior.rb
```diff
@@ -98,12 +98,23 @@ def file_set_params
...
         super
       end
 
+      # If the user is creating a bunch of files, and not in a work context,
+      # create a work for each file.
+      def default_work
+        # Ensure the upload set exists, before trying to associate a work with it.
+        UploadSet.find_or_create(params[:upload_set_id])
+        GenericWork.create!(title: [params[:file_set][:files].first.original_filename],
+                            upload_set_id: params[:upload_set_id]) do |w|
```
awead: could you pass the UploadSet directly to create! since you'd already have it in-hand in the line above? It would eliminate an extra Fedora call, maybe.

jcoyne: Sure.

app/controllers/concerns/sufia/file_sets_controller_behavior.rb
```diff
@@ -28,7 +28,7 @@ module FileSetsControllerBehavior
     end
 
     def new
-      @upload_set_id = ActiveFedora::Noid::Service.new.mint
+      @upload_set_id = SecureRandom.uuid
```
mjgiarlo: Given the size of the UUID identifier space this is very unlikely buuuuut couldn't this line generate a UUID already allocated in Fedora? Do we worry about that, and hook into AF here? Or add a method to AF to spit out an unused UUID?

spec/spec_helper.rb
```diff
@@ -115,6 +115,10 @@ def main_app
 
   config.use_transactional_fixtures = false
 
+  config.before :suite do
```
mjgiarlo: :eyes:

PR #1438, opened 2015-12-10
Don't monkeypatch events onto models

jcoyne:


mjgiarlo:
OK, that's dope, and was long-needed!

PR #1439, opened 2015-12-10
Clear up Files/Works distinction in FilesController and My::FilesController

awead:
This is a series of commits that address a couple of issues and generally change the notion of "files" in Sufia to either "file sets" or "works" as appropriate. It doesn't completely clear up files versus works issues, but is a step in the process.


app/controllers/concerns/sufia/breadcrumbs.rb
```diff
@@ -27,10 +27,10 @@ def trail_from_referer
 
     def add_breadcrumb_for_controller
       case controller_name
-      when 'file_sets'.freeze, 'my/files'.freeze, 'batch_edits'.freeze
-        add_breadcrumb I18n.t('sufia.dashboard.my.files'), sufia.dashboard_files_path
+      when 'file_sets'.freeze, 'my/works'.freeze, 'batch_edits'.freeze
+        add_breadcrumb I18n.t('sufia.dashboard.my.works'), sufia.dashboard_works_path
       when 'stats'.freeze
```
mjgiarlo: Could we add `'stats'.freeze` to line 30 instead of keeping an extra branch that calls the same action as the prior condition?

spec/routing/route_spec.rb
```diff
@@ -78,8 +78,8 @@
       expect(get: '/dashboard/activity').to route_to(controller: 'dashboard', action: 'activity')
     end
 
-    it "routes to my works tab" do
-      expect(get: '/dashboard/files').to route_to(controller: 'my/files', action: 'index')
+    it "routes to my files tab" do
```
mjgiarlo: Shouldn't this still be works?

mjgiarlo:
Two questions. Other than those, :+1:

mjgiarlo:
Travis is :unamused:. :ghost:

spec/routing/route_spec.rb
awead: yep, sure should be.

app/controllers/concerns/sufia/breadcrumbs.rb
awead: :+1:

jcoyne:
If you rebase, travis should be fixed.

PR #1440, opened 2015-12-10
Put homepage text into i18n

### LINKED ISSUE #1421: Recently Uploaded should say "Work Details" instead of "File Details"

jcoyne:
Changed "File Details" to "Details"
Fixes #1421 
Removed variables we're no longer using.
Moved controller to the Sufia namespace.


app/views/sufia/homepage/_recent_document.html.erb
```diff
@@ -1,18 +1,13 @@
...
+      <td class="col-sm-2">
+        <%= link_to_profile recent_document.depositor("no depositor value") %>
+        <%= link_to curation_concerns_generic_work_path(recent_document) do %>
+          <%= render_thumbnail_tag recent_document, { width: 45 }, false %>
+        <% end %>
+      </td>
       <td>
         <h3>
           <span class="sr-only">Title</span><%= link_to truncate(recent_document.title_or_label, length: 28, separator: ' '), curation_concerns_file_set_path(recent_document)  %>
-          <% if display_access %>
```
mjgiarlo: Why'd you nuke this block?

jcoyne: It was put there way back when this partial was called from different places and we wanted it to show in one place and not another.  Now it's only called from one place, so we don't need the conditional.

mjgiarlo:
Routing errors make travis sad.

awead:
Looks good to me... but Travis doesn't seem to think so.

mjgiarlo:
Travis and @jcoyne have now reconciled their differences.

PR #1446, opened 2015-12-10
Merges sufia-models back into sufia. Fixes #1423

### LINKED ISSUE #1423: sufia-models no longer needs to be a gem
Remove the bits around managing the sufia-models gem (such as rake tasks), and reorganize the code such that we no longer have a sufia-models directory.

mjgiarlo:
This is rebased on the latest from `master`, and includes some streamlining of `sufia.rb` and the deletion of ChecksumAuditLog-related code which already lives in CurationConcerns but was hanging around causing me problems.

If I might ask an indulgence... if there are concerns related to code organization -- i.e., ones that are fairly easily fixable -- let me know and I will put in follow-up issues and tackle those today/tomorrow. I'd prefer that to holding up this very large PR which will be a merge conflict thorn in our sides until it's merged.


jcoyne:
:tada: :balloon: :clap: :cake: :high_brightness:

jcoyne:
Why did the coverage drop so drastically?

mjgiarlo:
more generators in `lib/generators`? I suspect we should omit that entire dir in spec_helper. For the next PR?

PR #1447, opened 2015-12-11
Tidy up the /lib directory

mjgiarlo:
Done for a few reasons. First, to get coverage back up over 90%. Second, to move stuff that should be in `app/` out of `lib/`. Third, to break the huge `sufia.rb` up so that the Engine stuff lives in its own file `lib/sufia/engine.rb` which seems common in engine gems.


.rubocop.yml
```diff
@@ -103,7 +103,7 @@ Style/GlobalVars:
     - 'spec/**/*'
     - 'spec/jobs/event_jobs_spec.rb'
     - 'lib/generators/sufia/templates/config/redis_config.rb'
-    - 'lib/sufia/redis_event_store.rb'
+    - 'app/models/concerns/sufia/redis_event_store.rb'
```
jcoyne: RedisEventStore is not a concern. It is a Class and it's not a model.

mjgiarlo: Is `lib` the best place for it?

mjgiarlo:
Sweeeeeeet coverage.

.rubocop.yml
jcoyne: I can't think of a better place off hand

mjgiarlo:
Do not merge unti lhttps://github.com/projecthydra-labs/curation_concerns/pull/550 is merged.

.rubocop.yml
mjgiarlo: Moved back to `lib/` and pushed. This needs to be merged before this PR can go in: https://github.com/projecthydra-labs/curation_concerns/pull/550

spec/test_app_templates/Gemfile.extra
```diff
@@ -1,6 +1,6 @@
 # Use this file to reference specific commits of gems.
 
-gem 'curation_concerns', github: 'projecthydra-labs/curation_concerns', branch: 'master'
+gem 'curation_concerns', github: 'projecthydra-labs/curation_concerns', branch: 'toggle_popover'
```
mjgiarlo: This is the bit I need to revert once https://github.com/projecthydra-labs/curation_concerns/pull/550 is merged, and then I'll force-push what should be the last change required for the `lib/` tidy-up.

spec/spec_helper.rb
```diff
@@ -30,8 +30,16 @@
   SimpleCov.root(File.expand_path('../..', __FILE__))
   SimpleCov.formatter = Coveralls::SimpleCov::Formatter
   SimpleCov.start('rails') do
+    add_filter '/.internal_test_app'
+    add_filter '/bin'
+    add_filter '/db'
+    add_filter '/config'
```
jcoyne: I'm pretty sure /config is already filtered by the 'rails' profile See https://github.com/colszowka/simplecov#profiles

spec/spec_helper.rb
```diff
@@ -30,8 +30,16 @@
...
+    add_filter '/jetty'
+    add_filter '/lib/generators'
+    add_filter '/public'
+    add_filter '/solr_conf'
```
jcoyne: I'm not sure we should filter solr_conf, public, jetty, db, bin, vendor, because we don't have any ruby files in there anyway.

PR #1469, opened 2015-12-11
Require CurationConcerns 0.4

jcoyne:


mjgiarlo:
Something about FileSets not accepting the "file_format" message. NEEDS MOAR UPDOG

PR #1476, opened 2015-12-15
Allows the File Set Show page to display. 

### LINKED ISSUE #1464: FileSet show page should render
When I follow these steps:

1. Create a new work from the dashboard.
1. Select a few files and upload them.
1. Click a FileSet show link from the Work show page.

I see a `NoMethodError in CurationConcerns::FileSets#show`:

```
Showing /home/mjg/workspace/sufia/app/views/shared/_citations.html.erb where line #13 raised:
undefined method `first' for nil:NilClass
11  <meta name="twitter:data1" content="<%= @presenter.tag.join(', ') %>"/>
12  <meta name="twitter:label1" content="Keywords"/>
13  <meta name="twitter:data2" content="<%= @presenter.rights.first %>"/>
14  <meta name="twitter:label2" content="Rights"/>
15 <% end %>
```

Additionally, even though I selected a value for rights and it appears in the edit form, it does not render in the presenter. These are likely related. (I also have given this Work a tag which appears in the form but does not render in the presenter so we should make sure that all values in the form are in the presenter.)

hectorcorrea:
There are still several TODOs that need to be addressed. Fixes #1464

This is how it looks so far.
screen shot 2015-12-15 at 10 20 02 am

### DIFF
--- a/app/helpers/file_set_helper.rb
+++ b/app/helpers/file_set_helper.rb
@@ -15,16 +15,17 @@ module FileSetHelper
   end
 
   def display_multiple(value)
+    return if value.nil?
     auto_link(value.join(" | "))
   end
 
   private
 
-    def render_visibility_badge
-      if can? :edit, @file_set
-        render_visibility_link @file_set
+    def render_visibility_badge(presenter)
+      if can? :edit, presenter.solr_document
+        render_visibility_link presenter.solr_document
       else
-        render_visibility_label @file_set
+        render_visibility_label presenter.solr_document
       end
     end
 end
--- a/app/presenters/sufia/file_set_presenter.rb
+++ b/app/presenters/sufia/file_set_presenter.rb
@@ -2,7 +2,7 @@ module Sufia
   class FileSetPresenter < ::CurationConcerns::FileSetPresenter
     include Sufia::CharacterizationBehavior
 
-    delegate :depositor, :tag, :date_created, to: :solr_document
+    delegate :depositor, :tag, :date_created, :date_modified, to: :solr_document
 
     def tweeter
       user = ::User.find_by_user_key(depositor)
@@ -11,6 +11,17 @@ module Sufia
       else
         I18n.translate('sufia.product_twitter_handle')
       end
+    end
+
+    def rights
+      return if solr_document.rights.nil?
+      solr_document.rights.first
+    end
+
+    def related_files
+      # TODO: support related_files. Maybe.
+      # See https://github.com/projecthydra/sufia/issues/1478
+      []
     end
 
     # Add a schema.org itemtype
--- a/app/views/curation_concerns/base/_show_actions.html.erb
+++ b/app/views/curation_concerns/base/_show_actions.html.erb
@@ -1,3 +1,7 @@
+<%# TODO: set these variables based on the current user
+  # See https://github.com/projecthydra/sufia/issues/1477 %>
+<% collector = true %>
+<% editor = true %>
 <% if collector || editor %>
   <div class="form-actions">
     <% if Sufia.config.analytics %>
@@ -14,6 +18,8 @@
     <%= link_to 'Export to Zotero', sufia.static_path('zotero'), id: 'zoteroLink', name: 'zotero', class: 'btn btn-default' %>
     <%= link_to 'Export to Mendeley', sufia.static_path('mendeley'), id: 'mendeleyLink', name: 'mendeley', class: 'btn btn-default' %>
     <!-- COinS hook for Zotero -->
-      <span class="Z3988" title="<%= export_as_openurl_ctx_kev(@presenter) %>"></span>
+    <!-- TODO: Fix issue with export_as_openurl_ctx_kev
+         See  https://github.com/projecthydra/sufia/issues/1479 -->
+      <span class="Z3988" title="<%= 'TODO: export_as_openurl_ctx_kev(@presenter)' %>"></span>
   </div>
 <% end %>
--- a/app/views/curation_concerns/file_sets/_show_collections.html.erb
+++ b/app/views/curation_concerns/file_sets/_show_collections.html.erb
@@ -1,5 +1,7 @@
     <h2 class="non lower">Collections</h2>
-    <% if @file_set.parent_collections.blank? %>
+    <!-- TODO: remove hard code "true" and fix collections display
+         See https://github.com/projecthydra/sufia/issues/1480 -->
+    <% if true || @file_set.parent_collections.blank? %>
         <p><%= t('sufia.file.collections_list.empty') %></p>
     <% else %>
         <p><%= t('sufia.file.collections_list.heading') %></p>
--- a/app/views/curation_concerns/file_sets/_show_details.html.erb
+++ b/app/views/curation_concerns/file_sets/_show_details.html.erb
@@ -1,16 +1,16 @@
 <h2>File Details</h2>
 <dl class="dl-horizontal file-show-term file-show-details">
   <dt>Depositor</dt>
-  <dd itemprop="accountablePerson" itemscope itemtype="http://schema.org/Person"><span itemprop="name"><%= link_to_profile @file_set.depositor %></span></dd>
+  <dd itemprop="accountablePerson" itemscope itemtype="http://schema.org/Person"><span itemprop="name"><%= link_to_profile @presenter.depositor %></span></dd>
   <dt>Date Uploaded</dt>
-  <dd itemprop="datePublished"><%= @file_set.date_uploaded %></dd>
+  <dd itemprop="datePublished"><%= @presenter.date_uploaded %></dd>
   <dt>Date Modified</dt>
-  <dd itemprop="dateModified"><%= @file_set.date_modified %></dd>
+  <dd itemprop="dateModified"><%= @presenter.date_modified %></dd>
   <dt>Audit Status</dt>
   <dd><%= @presenter.audit_status %></dd>
-  <% unless @file_set.related_files.empty? %>
+  <% unless @presenter.related_files.empty? %>
   <dt>Related Files</dt>
-      <% @file_set.related_files.each do |f| %>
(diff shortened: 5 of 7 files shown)


mjgiarlo:
@hectorcorrea Not sure if this helps document/break down your TODOs, but you can construct task lists in GitHub comments like so:

- [ ] Do a thing
- [ ] Frobulate the widgetizers
- [ ] Burn rubocop to the ground
- [ ] Cuss @mjgiarlo out on GitHub

(It's also cool to have TODOs in code comments, naturally.)

app/helpers/file_set_helper.rb
```diff
@@ -15,12 +15,20 @@ def render_collection_list(fs)
   end
 
   def display_multiple(value)
-    auto_link(value.join(" | "))
+    # TODO: figure out why sometimes we get an string and some others an array
```
mjgiarlo: This is a good TODO to leave in. I think the next few lines of code are a good candidate to take a "tell, don't ask" approach. Like so:

```ruby
string_value = Array(value).join(" | ")
auto_link(string_value)
```

app/helpers/file_set_helper.rb
```diff
@@ -15,12 +15,20 @@ def render_collection_list(fs)
...
+      string_value = value
+    end
+    auto_link(string_value)
   end
 
   private
 
     def render_visibility_badge
+      # TODO: figure out why #file_set is sometimes nil
+      return unless @file_set
```
mjgiarlo: Yay for guard clauses.

app/presenters/sufia/file_set_presenter.rb
```diff
@@ -13,6 +13,16 @@ def tweeter
       end
     end
 
+    def rights
+      return nil if solr_document.rights.nil?
```
mjgiarlo: Can omit the explicit `nil` here, like you did in `render_visibility_badge` above.

app/presenters/sufia/file_set_presenter.rb
```diff
@@ -13,6 +13,16 @@ def tweeter
...
+      solr_document.rights.first
+    end
+
+    def related_files
```
mjgiarlo: I wonder if related files still makes sense in Sufia 7.0.0. Related files in Sufia < 7 was a way to say "all these GFs/FSs that were uploaded as a single Batch/UploadSet are probably related," but FSs are not collected in USs in Sufia 7. I'll make a strawperson proposal.

We should split the notion of related files:

 1. FileSets within a single Work should be related such that you can browse to other FSs in a Work without going back up/out to the Work show page.
 1. Works should be related by virtue of having been uploaded together in an UploadSet (same idea as GFs/Batch in Sufia < 7).

Does that make sense? (This also clearly goes above and beyond this ticket.)

app/views/curation_concerns/base/_show_actions.html.erb
```diff
@@ -1,3 +1,6 @@
+<%# TODO: set these variables based on the current user %>
```
mjgiarlo: I don't get what these roles mean here. So these will be replaced with an ability check or what?

app/views/curation_concerns/base/_show_actions.html.erb
```diff
@@ -14,6 +17,7 @@
     <%= link_to 'Export to Zotero', sufia.static_path('zotero'), id: 'zoteroLink', name: 'zotero', class: 'btn btn-default' %>
     <%= link_to 'Export to Mendeley', sufia.static_path('mendeley'), id: 'mendeleyLink', name: 'mendeley', class: 'btn btn-default' %>
     <!-- COinS hook for Zotero -->
-      <span class="Z3988" title="<%= export_as_openurl_ctx_kev(@presenter) %>"></span>
+    <!-- TODO: Fix issue with export_as_openurl_ctx_kev -->
```
mjgiarlo: Is there an issue for that in GitHub, or do you think that should be part of this issue?

app/views/curation_concerns/file_sets/_show_collections.html.erb
```diff
@@ -1,5 +1,6 @@
     <h2 class="non lower">Collections</h2>
-    <% if @file_set.parent_collections.blank? %>
+    <!-- TODO: remove hard code "true" and fix collections display -->
```
mjgiarlo: What needs fixing, and should that be part of this issue or another issue?

app/views/curation_concerns/file_sets/_show_details.html.erb
```diff
@@ -1,16 +1,16 @@
...
 <dl class="dl-horizontal file-show-term file-show-details">
   <dt>Depositor</dt>
-  <dd itemprop="accountablePerson" itemscope itemtype="http://schema.org/Person"><span itemprop="name"><%= link_to_profile @file_set.depositor %></span></dd>
+  <dd itemprop="accountablePerson" itemscope itemtype="http://schema.org/Person"><span itemprop="name"><%= link_to_profile @presenter.depositor %></span></dd>
   <dt>Date Uploaded</dt>
-  <dd itemprop="datePublished"><%= @file_set.date_uploaded %></dd>
+  <dd itemprop="datePublished"><%= @presenter.date_uploaded %></dd>
   <dt>Date Modified</dt>
-  <dd itemprop="dateModified"><%= @file_set.date_modified %></dd>
+  <dd itemprop="dateModified"><%= "TODO:date_modified" %></dd>
```
mjgiarlo: Huh, is `date_modified` not available in the presenter?

app/views/curation_concerns/file_sets/show.html.erb
```diff
@@ -11,14 +11,13 @@
...
     <div class="col-xs-12 col-sm-4">
-      <%= render 'generic_files/media_display' %>
+      <%= media_display @presenter %>
       <%= render 'show_actions' %>
       <%= render 'show_collections' %>
     </div>
     <div itemscope itemtype="<%= @presenter.itemtype %>" class="col-xs-12 col-sm-8">
       <h1 class="visibility"><%= display_multiple @presenter.title %> <%= render_visibility_badge %></h1>
       <p class="genericfile_description"><%= display_multiple @presenter.description %></p>
-      <%= render 'show_descriptions' %>
```
mjgiarlo: Are you deleting this because there's currently no way to edit the descriptions that would be shown here? If so, I'd say we should use `<%# TODO: render 'show_descriptions' %>` instead, since there are discussions about allowing FileSet descriptions in Sufia 7. (If not in the 7.0.0 release.)

mjgiarlo:
Looking good so far, @hectorcorrea! Left some comments about the nitty-gritties.

app/helpers/file_set_helper.rb
hectorcorrea: No changes were needed to this method after all.

app/helpers/file_set_helper.rb
hectorcorrea: I was able to remove the TODO and the guard clause for this once I figure out what this code was doing.

app/presenters/sufia/file_set_presenter.rb
hectorcorrea: Good idea. Done.

app/presenters/sufia/file_set_presenter.rb
hectorcorrea: Added new ticket.

app/views/curation_concerns/base/_show_actions.html.erb
hectorcorrea: Added new ticket.

app/views/curation_concerns/base/_show_actions.html.erb
hectorcorrea: Added new ticket.

app/views/curation_concerns/file_sets/_show_collections.html.erb
hectorcorrea: Added new ticket.

app/views/curation_concerns/file_sets/_show_details.html.erb
hectorcorrea: Yeah, it wasn't. I added it and we are now good.

hectorcorrea:
This is ready to merge. Missing items have been entered as new tickets (See links above)

app/helpers/file_set_helper.rb
```diff
@@ -15,16 +15,16 @@ def render_collection_list(fs)
   end
 
   def display_multiple(value)
-    auto_link(value.join(" | "))
+    auto_link(value)
```
hectorcorrea: hum...hold on one sec. I shouldn't have changed this. Resubmitting in a second.

app/views/curation_concerns/file_sets/show.html.erb
```diff
@@ -11,14 +11,13 @@
...
-      <%= render 'generic_files/media_display' %>
+      <%= media_display @presenter %>
       <%= render 'show_actions' %>
       <%= render 'show_collections' %>
     </div>
     <div itemscope itemtype="<%= @presenter.itemtype %>" class="col-xs-12 col-sm-8">
-      <h1 class="visibility"><%= display_multiple @presenter.title %> <%= render_visibility_badge %></h1>
+      <h1 class="visibility"><%= @presenter.title %> <%= render_visibility_badge @presenter %></h1>
       <p class="genericfile_description"><%= display_multiple @presenter.description %></p>
-      <%= render 'show_descriptions' %>
```
mjgiarlo: Why are we removing this partial again?

mjgiarlo:
Hi @hectorcorrea, I have just one outstanding question about a partial rendering deleted in the FS show view.

app/views/curation_concerns/file_sets/show.html.erb
hectorcorrea: Done.

mjgiarlo: :clap:

PR #1482, opened 2015-12-16
Makes OpenUrlFormatter more forgiving when the work... 

### LINKED ISSUE #1479: Call to export_as_openurl_ctx_kev
View `views/curation_concerns/base/_show_actions.html.erb` calls `export_as_openurl_ctx_kev` that passes `FileSetPresenter` as the parameter but the function expects a `work` which causes an error. See PR #1476 

Not sure if we need to pass a different value, refactor `export_as_openurl_ctx_kev` to allow a presenter, remove the call to it from this view, or something else.

hectorcorrea:
...does not have all the fields. In particular `resource_type`, `identifier`, and `based_near` are not in FileSets. Fixes #1479


mjgiarlo:
:+1:

PR #1485, opened 2015-12-17
Makes sure the date_uploaded is initialized...

### LINKED ISSUE #1467: Date uploaded is not being displayed for works on the dashboard
http://localhost:3000/dashboard/works

hectorcorrea:
...when creating a default_work. Fixes #1467

### DIFF
--- a/app/controllers/concerns/sufia/file_sets_controller_behavior.rb
+++ b/app/controllers/concerns/sufia/file_sets_controller_behavior.rb
@@ -93,19 +93,15 @@ module Sufia
       end
 
       def find_parent_by_id
-        return default_work unless parent_id.present?
+        return default_work_from_params unless parent_id.present?
         super
       end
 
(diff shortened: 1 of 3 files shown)


grosscol:
I thought that GenericWork was the default work created by the sufia install process.  To what end are we defining a DefaultWork class that doesn't inherit from any of the PCDM or Hydra::Works classes?

jcoyne:
I have the same concerns as @grosscol

jcoyne:
I see. DefaultWork is a service class. @hectorcorrea can you move it
out of app/models? I don't think it belongs there.

hectorcorrea:
@grosscol @jcoyne Good idea. I made the class a service instead of a model.

mjgiarlo:
You'll want to rebase this on `master` once #1490 is merged.

hectorcorrea:
@mjgiarlo rebased

PR #1491, opened 2015-12-18
Don't generate auto_link on a non-url rights statement, fixes project…

hackmastera:
…hydra/sufia#1356


hackmastera:
this has problems. will follow-up on Monday.

Issue #1356, opened 2015-11-23
'All rights reserved' should not display twice on record show page


hackmastera:
![all-rights-twice](https://cloud.githubusercontent.com/assets/845363/11339535/f3bc802a-91c7-11e5-9f45-2654c9c81936.png)

Here's a code trace:
* cc_licenses hash is designed to hold a label => url, but for 'all rights reserved' it just holds the label in both places, since there isn't a url. Note the url is the part that's saved as metadata: https://github.com/projecthydra/sufia/blob/master/sufia-models/lib/generators/sufia/models/templates/config/sufia.rb#L20
* template calls iconify_auto_link on the url: https://github.com/projecthydra/sufia/blob/master/app/views/records/show_fields/_rights.html.erb#L2
* which just passes a block to create the icon: https://github.com/projecthydra/sufia/blob/master/app/helpers/sufia/sufia_helper_behavior.rb#L116
* in the the rails_autolink [1] plugin.

Autolink doesn't do anything if it doesn't find any links; There's no link in 'all rights reserved' so it just returns the text. Blocks passed to autolink are applied on every link found; not on the text as a whole.

So I know what's happening but I don't really know the best way to fix it. Ideally we'd have a url for 'all rights reserved.' Otherwise, I guess the best place to fix it is in show_fields/_rights.html.erb with a hack that makes sure the value is a url before calling iconify_auto_link. I guess with a simple regex?

[1] https://github.com/tenderlove/rails_autolink/blob/master/lib/rails_autolink/helpers.rb

mjgiarlo:
@hackmastera How about this URL I chanced upon the other day? https://github.com/projecthydra-labs/curation_concerns/blob/master/lib/generators/curation_concerns/templates/config/authorities/rights.yml#L18

hackmastera:
@mjgiarlo that looks like a great find. not my area of expertise, though. @catlu what do you think?

catlu:
@hackmastera @mjgiarlo I think that would work great!

mjgiarlo:
@hackmastera @catlu :clap:

mjgiarlo:
@hackmastera @catlu This should be fixed in Sufia 7.0.0. Do you think we need a fix in the 6.x line or meh?

hackmastera:
@mjgiarlo We've talked about this more here. 

[...]

PR #1492, opened 2015-12-21
Don't generate auto_link on a non-url rights statement, fixes project…

hackmastera:
…hydra/sufia#1356

### DIFF
--- a/app/helpers/sufia/sufia_helper_behavior.rb
+++ b/app/helpers/sufia/sufia_helper_behavior.rb
@@ -113,12 +113,17 @@ module Sufia
       link_to(display, link_url)
     end
 
-    def iconify_auto_link(text, showLink = true)
+    def iconify_auto_link(text, show_link = true)
       # this block is only executed when a link is inserted;
       # if we pass text containing no links, it just returns text.
       auto_link(text) do |value|
-        "<i class='glyphicon glyphicon-new-window'></i>#{('&nbsp;' + value) if showLink}<br />"
+        "<i class='glyphicon glyphicon-new-window'></i>#{('&nbsp;' + value) if show_link}<br />"
       end
+    end
+
+    def iconify_link(text)
+      # different from above in that if text is not a url, return nil.
+      iconify_auto_link(text, false) if is_url?(text)
     end
 
     def link_to_profile(login)
@@ -189,6 +194,10 @@ module Sufia
       return user_key if user.nil?
 
       user.respond_to?(:name) ? "#{user.name} (#{user_key})" : user_key
+    end
+
+    def is_url?(str)
+      str =~ /\A#{URI.regexp(['http', 'https'])}\z/
     end
 
     private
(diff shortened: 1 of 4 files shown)


app/views/records/show_fields/_rights.html.erb
```diff
@@ -1,4 +1,4 @@
 <% record.rights.each do |r| %>
-  <%=link_to_field('rights',r, Sufia.config.cc_licenses_reverse[r])%> <%= iconify_auto_link(r,false) %><br />
+  <%=link_to_field('rights',r, Sufia.config.cc_licenses_reverse[r])%><% if is_url?(r) %> <%=iconify_auto_link(r,false) %><% end %><br />
```
jcoyne: Can we put this code into the `iconify_auto_link` helper method instead of the view?

hackmastera: discussed with @mjgiarlo last week. i was of the opinion making the change here would have a more limited impact; changing it in iconify_auto_link would be changing the behavior of a public method. But if y'all believe that is the intended behavior of that method more broadly then we can certainly move it there.

jcoyne: Maybe a compromise approach?  Make a new helper that is for linking licenses?  Mainly I don't want to see a lot of conditionals in the views.

hackmastera: Sure that sounds reasonable. I will push a new version in a bit.

hackmastera:
@jcoyne I moved it as you suggested. taking the conditional out of the template means there's a trailing space when the link isn't filled in, e.g. 'all rights reserved '. Doesn't seem like something that should be put in the helper but if there's a nice way to take care of that in the template I don't know it.

app/helpers/sufia/sufia_helper_behavior.rb
```diff
@@ -121,6 +121,11 @@ def iconify_auto_link(text, showLink = true)
       end
     end
 
+    def iconify_link(text, showLink = true)
```
jcoyne: Does this need to take a second argument?  Seems like it would only be used where showLink is false.  Also, variables should be snake case instead of camel case.

hackmastera: sure i'll leave out the 2nd arg and someone can add it back in in the future if there's another use case. happy to fix the camel case var in the method above while i'm in there.

hackmastera:
updated; ready

spec/helpers/sufia_helper_spec.rb
```diff
@@ -186,4 +186,16 @@ def create_models(model, user1, user2)
       conn.commit
     end
   end
+
+  describe "#is_url?" do
+    it "returns false for non-url string" do
+      url = 'all rights reserved'
+      expect(helper.is_url?(url)).to be_falsey
```
jcoyne: can you switch this to `be false` This indicates the method returns a true boolean value which is more specific than `falsey`.

spec/helpers/sufia_helper_spec.rb
```diff
@@ -186,4 +186,16 @@ def create_models(model, user1, user2)
...
+    end
+
+    it "returns true for url" do
+      url = 'http://creativecommons.org/licenses/by/3.0/us/'
+      expect(helper.is_url?(url)).to be_truthy
```
jcoyne: And `be true` here.

spec/helpers/sufia_helper_spec.rb
jcoyne: I see, it will actually be returning `nil` if not true, and `0` if true.

mjgiarlo:
Fixes #1356.

PR #1499, opened 2016-01-05
add resource_type delegate

### LINKED ISSUE #1498: Users should be allowed to assert a resource type on collection
Missing Resource Type error on create a new Collection and view a collection

njaffer:
closes #1498

### DIFF
--- a/spec/presenters/sufia/collection_presenter_spec.rb
+++ b/spec/presenters/sufia/collection_presenter_spec.rb
@@ -8,14 +8,19 @@ describe Sufia::CollectionPresenter do
                             :language, :identifier, :based_near, :related_url] }
   end
 
-  let(:collection) { Collection.new(id: '111', description: 'a nice collection', title: 'A clever title') }
-  let(:solr_doc) { SolrDocument.new(collection.to_solr) }
+  let(:collection) { Collection.new(id: '111', description: 'a nice collection', title: 'A clever title', resource_type: ['Collection']) }
   let(:ability) { double }
   let(:presenter) { described_class.new(solr_doc, ability) }
+  let(:solr_doc) { SolrDocument.new(collection.to_solr) }
+
+  describe "#resource_type" do
+    subject { presenter.resource_type }
+    it { is_expected.to eq ['Collection'] }
+  end
 
   describe "#terms_with_values" do
     subject { presenter.terms_with_values }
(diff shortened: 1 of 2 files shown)


jcoyne:
Please add a test to this PR.

njaffer:
@jcoyne, the test already has :resource_type in the terms. I checked the work presenter specs and don't see any additional coverage for testing the delegate. Can you please specify what do you mean by adding a test?

jcoyne:
@njaffer the test should check that the code you added works as you intend.  So I'd probably write a test like this:

```ruby
let(:presenter) { CollectionPresenter.new(solr_doc) }
let(:solr_doc) { double("Mock Solr document", resource_type: 'Collection') }
subject { presenter.resource_type }
it { is_expected.to eq 'Collection' }
```

You should have a test that fails before you make changes to the code and then succeeds after making the changes.  In this way no-one will accidentally revert your changes without seeing a test fail.

njaffer:
@jcoyne, this code gives a lot of errors, can we just check the resource type as part of the terms? I have added the test for that, let me know what you think

@mjgiarlo, I looked at adding the other delegates, but did not seem to justify adding all the extra ones when they are not needed. May be I am missing something?

spec/presenters/sufia/collection_presenter_spec.rb
```diff
@@ -8,14 +8,19 @@
                             :language, :identifier, :based_near, :related_url] }
   end
 
-  let(:collection) { Collection.new(id: '111', description: 'a nice collection', title: 'A clever title') }
-  let(:solr_doc) { SolrDocument.new(collection.to_solr) }
+  let(:collection) { Collection.new(id: '111', description: 'a nice collection', title: 'A clever title', resource_type: ['Collection']) }
   let(:ability) { double }
   let(:presenter) { described_class.new(solr_doc, ability) }
+  let(:solr_doc) { SolrDocument.new(collection.to_solr, resource_type: 'Collection') }
```
jcoyne: is it necessary to provide `resource_type: 'Collection'` here?  Shouldn't `collection.to_solr` take care of that?

mjgiarlo:
@njaffer: I agree that we shouldn't add the others until/unless we're sure they're necessary.

njaffer:
@jcoyne , yep don't need resource_type: 'Collection'. Updated the PR

PR #1500, opened 2016-01-06
fix new collections page styling

njaffer:


config/locales/sufia.en.yml
```diff
@@ -238,7 +238,7 @@ en:
         publisher: "The person or group making the file available. Generally this is the institution."
         rights: "Licensing and distribution information governing access to the file. Select from the provided drop-down list. &lt;em&gt;This is a required field&lt;/em&gt;."
       collection:
-        creator_html: 'The person or group responsible for the collection. Usually this is the author of the collected content. Personal names should be entered with the last name first, e.g. "Smith, John." <em>This is a required field</em>.'
+        creator_html: 'The person or group responsible for the collection. Usually this is the author of the collected content. Personal names should be entered with the last name first, e.g. "Smith, John." &lt;em&gt;This is a required field&lt;/em&gt;.'
```
jcoyne: This should not be necessary when the i18n key is suffixed by _html. Please explain what the issue is.

jcoyne: I see, it's putting that html into a `data-content` html tag attribute, which necessitates the extra escape.

PR #1503, opened 2016-01-07
Allows a user to feature/unfeature a work. Fixes #1049

### LINKED ISSUE #1049: Mark a Work as Featured Work

hectorcorrea:
Fixes #1049

Below is how the Featured Works show in the home page. This display would require some polish but I suggest the UI changes are handled in a separate ticker.

screen shot 2016-01-07 at 11 48 40 am

### DIFF
--- a/app/presenters/sufia/work_show_presenter.rb
+++ b/app/presenters/sufia/work_show_presenter.rb
@@ -16,6 +16,14 @@ module Sufia
       end
     end
 
+    def display_feature_link?
+      user_can_feature_works? && solr_document.public? && FeaturedWork.can_create_another? && !featured?
+    end
+
+    def display_unfeature_link?
+      user_can_feature_works? && solr_document.public? && featured?
+    end
+
     # Add a schema.org itemtype
     def itemtype
       # Look up the first non-empty resource type value in a hash from the config
@@ -24,5 +32,18 @@ module Sufia
     rescue
       'http://schema.org/CreativeWork'
     end
+
+    private
+
+      def featured?
+        if @featured.nil?
+          @featured = FeaturedWork.where(generic_work_id: solr_document.id).exists?
+        end
+        @featured
+      end
+
+      def user_can_feature_works?
+        current_ability.can?(:create, FeaturedWork)
+      end
   end
 end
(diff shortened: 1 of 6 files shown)


app/presenters/sufia/work_show_presenter.rb
```diff
@@ -24,5 +32,16 @@ def itemtype
     rescue
       'http://schema.org/CreativeWork'
     end
+
+    private
+
+      def featured?
+        @featured ||= !FeaturedWork.find_by_generic_work_id(solr_document.id).nil?
+        @featured
```
awead: not needed? Isn't `@featured` going to be returned anyway?

app/presenters/sufia/work_show_presenter.rb
```diff
@@ -24,5 +32,16 @@ def itemtype
     rescue
       'http://schema.org/CreativeWork'
     end
+
+    private
+
+      def featured?
+        @featured ||= !FeaturedWork.find_by_generic_work_id(solr_document.id).nil?
```
jcoyne: This isn't going to memoize if `@featured` is `false`.  You have to explicitly check to see if `@featured.nil?`  Also, why not `FeaturedWork.where(generic_work_id: solr_document.id).exists?` That should avoid the overhead of instantiation.

awead:
One very minor comment, otherwise :+1:

app/presenters/sufia/work_show_presenter.rb
```diff
@@ -24,5 +32,16 @@ def itemtype
...
+        @featured
+      end
+
+      def user_can_feature
```
jcoyne: should this be named `user_can_feature?` (with a question mark)

app/presenters/sufia/work_show_presenter.rb
hectorcorrea: Good catch. Thanks!

hectorcorrea:
Addressed the comments. Added a few tests.

mjgiarlo:
:clap: @hectorcorrea

PR #1505, opened 2016-01-07
Change visibility_link path to polymorphic_path

### LINKED ISSUE #1103: Visibility badge/link on "My Files" dashboard
The "My Files" page on the dashboard has been changed to show works instead of files.  On the old "My Files" page, each file had a visibility badge that was a link to the "permissions_display" tab on the edit form.  We'll need a similar visibility badge for the GenericWorks on the page.

Acceptance criteria for this story:
- [x] Add a visibility badge for each GenericWork listed on the "My Files" page.
- [ ] The visibility badge should be a link to the permissions edit page for the GenericWork 

The visibility badge renders A-OK for each GW in the "My Works" view, but the link is wrong. Currently it's pointing at `/concern/file_sets/{id}/edit#permissions_display` instead of `/concern/generic_works/d791sg234/edit#permissions_display`. On top of that, the latter URL is no longer a tabbed display, the so `#permissions_display` could/should be an anchor on the `<h2>Permissions</h2>` instead.

njaffer:
closes #1103

### DIFF
--- a/app/helpers/sufia/sufia_helper_behavior.rb
+++ b/app/helpers/sufia/sufia_helper_behavior.rb
@@ -117,7 +117,7 @@ module Sufia
 
     def render_visibility_link(document)
       link_to render_visibility_label(document),
-              main_app.edit_curation_concerns_file_set_path(document, anchor: "permissions_display"),
+              edit_polymorphic_path([main_app, document], anchor: "permissions_display"),
               id: "permission_" + document.id, class: "visibility-link"
     end
 
--- a/app/views/curation_concerns/base/_permission_form.html.erb
+++ b/app/views/curation_concerns/base/_permission_form.html.erb
@@ -3,7 +3,7 @@
 <% public_perm = true if params[:controller] == 'batch' %>
 <% registered_perm = f.object.permissions.map { |perm| perm.access if perm.agent_name == "registered"}.compact.first %>
 
-<h2><% if params[:controller] == 'batch' %>Bulk <% end %>Permissions <% if params[:controller] == 'batch' %>
+<h2 id="permissions_display"><% if params[:controller] == 'batch' %>Bulk <% end %>Permissions <% if params[:controller] == 'batch' %>
       <small>(applied to all files just uploaded)</small><% end %>
 </h2>
 
--- a/spec/views/my/_list_works.html.erb_spec.rb
+++ b/spec/views/my/_list_works.html.erb_spec.rb
@@ -4,7 +4,7 @@ describe 'my/_index_partials/_list_works.html.erb' do
   let(:work_title) { 'Work Title' }
   let(:coll_title) { 'Collection Title' }
   let(:collection) { FactoryGirl.build(:collection, id: '3197z497t', title: coll_title) }
-  let!(:work) { FactoryGirl.build(:work, id: '3197z511f', title: [work_title]) }
+  let!(:work) { FactoryGirl.create(:generic_work, id: '3197z511f', title: [work_title]) }
   let(:doc) { SolrDocument.new(work.to_solr) }
 
   let(:config) { My::WorksController.new.blacklight_config }
--- a/app/views/shared/_attributes.html.erb
+++ b/app/views/shared/_attributes.html.erb
@@ -4,7 +4,7 @@
           <td><%= work.title %></td>
         </tr>
         <tr>
-          <th><span class="attribute-label h4">Depositor:</span></div>
+          <th><span class="attribute-label h4">Depositor:</span></th>
           <td><%= link_to_profile work.depositor("no depositor value") %></td>
         </tr>
         <tr>


jcoyne:
Instead of saying "fixed ..." can you write a more descriptive commit message that says what you changed?

app/helpers/sufia/sufia_helper_behavior.rb
```diff
@@ -117,7 +117,7 @@ def search_form_action
 
     def render_visibility_link(document)
       link_to render_visibility_label(document),
-              main_app.edit_curation_concerns_file_set_path(document, anchor: "permissions_display"),
+              main_app.edit_curation_concerns_generic_work_path(document, anchor: "permissions_display"),
```
jcoyne: This is concerning to me.  Are we sure that document always represents a GenericWork and never a FileSet?  I'm pretty sure the FileSet page calls this method too.

njaffer: I was thinking about that too, and I looked at the file set visibility badges, and they are not clickable, so they only show the visibility status and does not call the method render_visibility_link. The only way to change the visibility of the FileSet is to go through the Edit option

jcoyne: When it's called from here https://github.com/projecthydra/sufia/blob/60326e5de1262176f332fb26b329fe3610097b62/app/helpers/file_set_helper.rb#L19-L25 via https://github.com/projecthydra/sufia/blob/master/app/views/curation_concerns/file_sets/show.html.erb#L19 I'm pretty sure it's a FileSet.

njaffer:
@jcoyne , yep thats true, but its strange that when I have created a work and added files to edit, then it should show me the visibility badge as a link, but it does not..
I can see your point here though, so should there be 2 separate methods? one for file_set and one for generic_work?

jcoyne:
@njaffer I wonder if it couldn't be a polymorphic path. `[:main_app, :edit, :curation_concerns, document]` or something like that?

mjgiarlo:
@jcoyne :+1: to polymorphic path

jcoyne:
try just `[:edit, document]` first. That's what i see in CurationConcerns.

njaffer:
@jcoyne: I have added the plymorphic_path, but not sure how to add the anchor

njaffer:
The test is failing because it does not have all the required attributes for document. I compared the 'document' fields and here is the gist: https://gist.github.com/njaffer/9b790581a93f9688b056

Any ideas? Should I set all the fields to be the same as actual one?

njaffer:
I have changed the FactoryGirl.build to FactoryGirl.create,  build is not persistent but create is. With '.create' the document is persistent and hence the test is passing!

I think it is ready to go! Thank you all for your help @jcoyne , @grosscol , @hectorcorrea, @mjgiarlo

app/views/curation_concerns/base/_permission_form.html.erb
```diff
@@ -3,7 +3,7 @@
 <% public_perm = true if params[:controller] == 'batch' %>
 <% registered_perm = f.object.permissions.map { |perm| perm.access if perm.agent_name == "registered"}.compact.first %>
 
-<h2><% if params[:controller] == 'batch' %>Bulk <% end %>Permissions <% if params[:controller] == 'batch' %>
+<h2><% if params[:controller] == 'batch' %>Bulk <% end %><a name="permissions_display">Permissions </a><% if params[:controller] == 'batch' %>
```
jcoyne: Why are we putting an anchor here?

jcoyne: Why don't we just put `id="permissions_display"` on this tag: https://github.com/projecthydra/sufia/blob/5f7098d70849358828680cd7468570d577098176/app/views/curation_concerns/base/_permission_form.html.erb#L18

jcoyne: Or on the `<h2>`?

njaffer: It is on <h2>: https://github.com/projecthydra/sufia/blob/master/app/views/curation_concerns/base/_permission_form.html.erb#L6

jcoyne: My question is why are we adding a new anchor tag. There's no reason to do this.  Just stick the id attribute somewhere, such as the h2 at the beginning of the line.

njaffer: ok ...

spec/views/my/_list_works.html.erb_spec.rb
```diff
@@ -4,7 +4,7 @@
   let(:work_title) { 'Work Title' }
   let(:coll_title) { 'Collection Title' }
   let(:collection) { FactoryGirl.build(:collection, id: '3197z497t', title: coll_title) }
-  let!(:work) { FactoryGirl.build(:work, id: '3197z511f', title: [work_title]) }
+  let!(:work) { FactoryGirl.create(:generic_work, id: '3197z511f', title: [work_title]) }
```
jcoyne: Can we switch this back to build and stub out `persisted?: true` ?  This will make the test faster.  This is not a *big* deal, but the slowness adds up, so we should nip it in the bud.

njaffer: @jcoyne , when I use .build with persisted?: true, the travis build still failed.. can we see the history of the PR?

njaffer:
@jcoyne, @mjgiarlo: I would be interested in looking at your solution for .build instead of .create.
Here are my gist of errors and the code for .build with persisted: true
https://gist.github.com/njaffer/5eaeb8fa44e0149bfb39
https://gist.github.com/njaffer/f7d3bdebc36edb30ec8c

PR #1507, opened 2016-01-12
Allows users to highlight works

### LINKED ISSUE #1140: Allow a user to highlight a work from the my work listing
The action in the my works listing has a todo.
We need to put this back in and allow users the ability to highlight works from the dashboard.

hectorcorrea:
Fixes #1140

### DIFF
--- a/app/controllers/concerns/sufia/users_controller_behavior.rb
+++ b/app/controllers/concerns/sufia/users_controller_behavior.rb
@@ -68,17 +68,17 @@ module Sufia::UsersControllerBehavior
   end
 
   def toggle_trophy
-    unless current_user.can? :edit, params[:parent_id]
+    work_id = params[:work_id]
+    unless current_user.can? :edit, work_id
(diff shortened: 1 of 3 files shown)


app/controllers/concerns/sufia/users_controller_behavior.rb
```diff
@@ -68,17 +68,17 @@ def update_directory?
   end
 
   def toggle_trophy
-    unless current_user.can? :edit, params[:parent_id]
+    work_id = params[:work_id]
+    unless current_user.can? :edit, GenericWork.find(work_id)
```
jcoyne: I think you can omit `GenericWork.find`, just `can? :edit, work_id` ought to do the right thing and will be faster.

hectorcorrea: Fixed. I didn't realize that passing the ID will just look it up from Solr.

mjgiarlo:
:+1:

PR #1511, opened 2016-01-13
display a message when no works found

### LINKED ISSUE #1221: Collection without permission
When I do not have access to see the items in a collection, the user should see in the view below a message that they do not currently have access to some of the items in the collection.

https://scholarsphere-staging.dlt.psu.edu/collections/8623j0533

screen shot 2015-07-01 at 9 48 48 am

njaffer:
closes #1221

### DIFF
--- a/app/views/collections/_document_list.html.erb
+++ b/app/views/collections/_document_list.html.erb
@@ -1,3 +1,6 @@
+<% if documents.empty? %>
+  <div class="alert alert-warning" role="alert"><%= t('sufia.collection.no_visible_works') %></div>
+<% end %>
 <% @document_list = documents %>
 <% @batch_size_on_other_page = 0 %>
 <% @all_checked = false %>
--- a/config/locales/sufia.en.yml
+++ b/config/locales/sufia.en.yml
@@ -166,6 +166,7 @@ en:
         title: "Filename will be the default title. Please provide a more meaningful title, and filenames will still be preserved by the system."
         resource_type: "You may select multiple types to apply to all files"
     collection:
+      no_visible_works: "The collection is either empty or does not contain items to which you have access."
       is_part_of: "Is part of"
       select_form:
         title: "Add to collection"


app/views/collections/_document_list.html.erb
```diff
@@ -1,3 +1,5 @@
+<% if documents.empty? %> <h3><%= t('sufia.collection.no_works') %></h3><br>
```
mjgiarlo: Can you put a line break before the `<h3>`?

@mtribone do you think `<h3>` is the right mark-up for that message?

config/locales/sufia.en.yml
```diff
@@ -166,6 +166,7 @@ en:
         title: "Filename will be the default title. Please provide a more meaningful title, and filenames will still be preserved by the system."
         resource_type: "You may select multiple types to apply to all files"
     collection:
+      no_works: "The collection is either empty or does not contain items to which you have access."
```
mjgiarlo: Would you mind renaming to `no_visible_works`?

app/views/collections/_document_list.html.erb
jcoyne: I'm not sure why you'd want the `<br>` in there.  If it needs a margin, give the `h3` an id or class attribute and add the margin to the style sheet.

mtribone:
Yeah the break tag could go and add some spacing via CSS. I'll look at the context of the page and see if the H3 is appropriate.

mtribone:
@njaffer Could we see a screen shot? It will help me with context and the message. Thanks!

njaffer:
screen shot 2016-01-14 at 10 36 13 am

njaffer:
@mtribone , I took the h3 tag from this one: https://github.com/projecthydra/sufia/blob/fa0190067b76f825d2e017ceb1dca7fb95cbe250/app/views/admin/stats/_top_data.html.erb#L2

but I will happy to change with whatever is the best option!

mtribone:
@njaffer Thanks for the screen shot. It helps! So if "Items in this Collection" is an H2 then a H3 would be semantically correct. Another option would be to use Bootstrap's "alert" if we would like to communicate the information a little differently.

http://getbootstrap.com/components/#alerts

njaffer:
screen shot 2016-01-14 at 12 32 51 pm

njaffer:
screen shot 2016-01-14 at 12 37 38 pm

app/views/collections/_document_list.html.erb
mjgiarlo: No, sorry. This is what I had in mind:

```ruby
<% if documents.empty? %>
<h3><%= t('sufia.collection.no_works') %></h3>
<% end %>
```

Not an HTML line break but a CRLF.

njaffer:
@mtribone , @mjgiarlo , can this be merged?

mjgiarlo:
Thanks, @njaffer!

PR #1512, opened 2016-01-14
Displays option to highlight work in profile when viewing a collection

hectorcorrea:
This PR address an extra TODO on highlighting works that I found after submitting PR #1507


jcoyne:
Please add some tests. :smile_cat:

hectorcorrea:
@jcoyne Added a view test to make sure the link is rendered. Notice that I am only testing that the link is rendered but not the logic (show it vs not show it, show highlight vs show un-highlight) since that is already being tested in the trophy helper (https://github.com/projecthydra/sufia/blob/master/spec/helpers/trophy_helper_spec.rb)

PR #1515, opened 2016-01-15
Prevents rendering of the extra visibility textbox in the Generic …

### LINKED ISSUE #1513: Generic Works Edit Page: Redundant Visibility Text box
The generic_works edit page has a Visibility text box under 'Describe Your Work' section. Since there is Visibility under permissions section, this seems to be redundant. Also when I tried to enter some text and saved the work, but the text was not saved for Visibility text box. 

screen shot 2016-01-14 at 1 52 40 pm

hectorcorrea:
…Work edit form. Visibility is already handled via the Persmissions form. Fixes #1513

Notice that this fix manually skips the visibility field when rendering `_form`. 

Another approach to fix this issue is to either remove the "visibility" field from `CurationConcerns::Forms::WorkForm` (https://github.com/projecthydra-labs/curation_concerns/blob/master/app/forms/curation_concerns/forms/work_form.rb#L17) or at least from the Sufia version of this form (https://github.com/projecthydra/sufia/blob/master/app/forms/sufia/forms/work_form.rb) 

Thoughts?

Anyway, below is how the form looks with the fix. Notice the extra visibility textbox reported on #1513 is gone. 
screen shot 2016-01-15 at 9 41 21 am

### DIFF
--- a/app/forms/sufia/forms/work_form.rb
+++ b/app/forms/sufia/forms/work_form.rb
@@ -5,7 +5,7 @@ module Sufia::Forms
     def rendered_terms
       terms - [:files, :visibility_during_embargo, :embargo_release_date,
                :visibility_after_embargo, :visibility_during_lease,
-               :lease_expiration_date, :visibility_after_lease]
+               :lease_expiration_date, :visibility_after_lease, :visibility]
     end
   end
 end
--- a/sufia.gemspec
+++ b/sufia.gemspec
@@ -35,7 +35,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'acts_as_follower', '>= 0.1.1', '< 0.3'
   spec.add_dependency 'carrierwave', '~> 0.9'
   spec.add_dependency 'oauth2', '~> 0.9'
-  spec.add_dependency 'google-api-client', '~> 0.7'
+  spec.add_dependency 'google-api-client', '~> 0.7', '< 0.9'
   spec.add_dependency 'legato', '~> 0.3'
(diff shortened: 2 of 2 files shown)


app/views/curation_concerns/base/_form.html.erb
```diff
@@ -7,7 +7,11 @@
     <p><small class="pull-right"><%= link_to 'Switch to Batch Upload', sufia.new_curation_concerns_file_set_path %></small></p>
     <div class="well">
       <% f.object.rendered_terms.each do |term| %>
-        <%= render_edit_field_partial(term, f: f) %>
+        <% if term == :visibility %>
```
jcoyne: Can't we remove `:visibility` from what `rendered_terms()` returns?

jcoyne: See https://github.com/projecthydra/sufia/blob/5f7098d70849358828680cd7468570d577098176/app/forms/sufia/forms/work_form.rb#L5-L9

njaffer:
@hectorcorrea, that's interesting solution. My thoughts was to update this line: https://github.com/projecthydra/sufia/blob/5f7098d70849358828680cd7468570d577098176/app/forms/sufia/forms/work_form.rb#L6

and add :Visibility

hectorcorrea:
Thanks @jcoyne @njaffer that was a much better solution. I've updated the PR with the suggestion.

hectorcorrea:
Pinned it to a specific version of google-api-client for now.

PR #1516, opened 2016-01-15
add featured? methods to file_set_presenter

gordonleacock:
Add methods *display_feature_link?* and *display_unfeature_link?* to file sufia/app/presenters/sufia/file_set_presenter.rb, so that app/views/curations_concerns/base/_show_actions.html.erb not longer dies with a method not found error when viewing the file page.

### DIFF
--- a/app/presenters/sufia/file_set_presenter.rb
+++ b/app/presenters/sufia/file_set_presenter.rb
@@ -17,6 +17,14 @@ module Sufia
       end
     end
 
+    def display_feature_link?
+      false
+    end
+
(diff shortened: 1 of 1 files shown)


jcoyne:
Files can't be featured. It will be more efficient, and probably easier to understand if there's a `app/views/curation_concerns/file_sets/_show_actions.html.erb` that is specific to `FileSet`s rather than try to shoehorn `Works` and `FileSets` into using one partial.

gordonleacock:
Thanks @jcoyne, I'll do what you suggest.

Issue #1520, opened 2016-01-20
Update Sufia 6.x to be compatible with rubocop 0.36


elrayle:
See PR https://github.com/projecthydra/sufia/pull/1514 which made the same changes for Sufia 7 in the master branch.

awead:
Fixed in 895a1ed98b182217b0da7139f6c5690ac66afd15

PR #1522, opened 2016-01-20
Updating to Rubocop 0.36

awead:


hectorcorrea:
Fixes #1520

app/helpers/batch_edits_helper.rb
```diff
@@ -6,6 +6,6 @@ def batch_delete
   end
 
   def render_check_all
-    render partial: 'batch_edits/check_all' unless params[:controller].match("my/collections")
+    render partial: 'batch_edits/check_all' unless params[:controller] =~ /my\/collections/
```
hectorcorrea: @awead In Sufia 7 we had to use an "==" to fix this because there seems to be a bug in Rubocop regarding this. See https://github.com/projecthydra/sufia/commit/c59b40f3648d2620b8a0a667d493fd0998b8413d

awead: @hectorcorrea ok. That issue makes sense, but the regex should still work. Although, it should probably be `/^my\/collections$/` if we want to be exact. I don't have any opinions one way or the other... we can add an exception to Rubocop, or update the regex.

hectorcorrea: Well, it looks like it works OK in this case. I wonder what caused the bug in Sufia 7. :ghost:

hectorcorrea:
:+1:

PR #1524, opened 2016-01-21
fixes #1056

### LINKED ISSUE #1056: On the user show page the labels are duplicative and should be I18N
The Labels for the Tabs are duplicated within the tabs, and do not necessarily match.

We should remove the duplication and change the labels to be I18N.

awead:
Backports fix from 7 to 6.


mjgiarlo:
:+1:

PR #1529, opened 2016-01-21
WIP: Updating Sufia 6 to use hydra-head 9.6

awead:
I'm guessing we'll need a generator to upgrade folks from 6.5 to 6.6. Does this sound right, @mjgiarlo ?

### DIFF
--- a/app/controllers/concerns/sufia/homepage_controller.rb
+++ b/app/controllers/concerns/sufia/homepage_controller.rb
@@ -7,7 +7,7 @@ module Sufia::HomepageController
     include Blacklight::SearchContext
     include Sufia::Controller
     include Blacklight::SearchHelper
-    include Hydra::Controller::SearchBuilder
+    include Blacklight::AccessControls::Catalog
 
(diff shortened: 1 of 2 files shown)


jcoyne:
@awead alternatively, we can fix hydra-head, so that it includes the correct modules instead of just warning.

awead:
@jcoyne I defer to you since you're more up-to-date with that particular gem than I am.

mjgiarlo:
@awead When @jcoyne says things like "we can fix {**x**}," I'm generally in support of fixing **x**. If not, yeah, a generator to facilitate the upgrade process would be awful convenient for folks.

jcoyne:
Regardless of what we do to HydraHead / migration tools, this PR is great. :+1:

mjgiarlo:
Travis is :angry:

mjgiarlo:
Merged all the 6.x PRs I saw with passing tests, @awead. This one is red.

awead:
@mjgiarlo yeah, this is to get Sufia 6 with HH 9.6, and therefore Blacklight 6... It still needs some work.

Issue #1530, opened 2016-01-22
Enabling Sufia 6 with the Hydra Head 9.6


awead:
There are some gem incompatibilities that restrict how Sufia can be upgraded to work with HH 9.6. The problem centers around Blacklight 5 versus Blacklight 6. Here are some options from least amount of work, to more work, at least as far as I can tell.
## Option 1: Restrict Sufia 6 to Hydra Head 9.5

Nothing is required, but could impact upgrade paths to Sufia 7, which will presumably use hydra-head 9.6 or later and Blacklight 6.
## Option 2: Update but stay with Blacklight 5

To update to hydra-head 9.6, blacklight-access-controls 0.2 must be used, which presently has projectblacklight/blacklight-access_controls#13 blocking it from working properly under Sufia. **Note:** No one version of BAC will work with both Blacklight 5 or 6. To fix,
- split BAC into two branches:
  1. a 0.2-x branch to support Blacklight 5
  2. master branch to support Blacklight 6
- potentially fix it in hydra-head?
## Option 3: Update and move to Blacklight 6

For this, blacklight-access-controls 0.3 must be used, which has no issues; however, updating Sufia 6 to use Blacklight 6 presents a significant amount of work:
- update to Blacklight 6 (some work has already been done psu-stewardship/sufia@b086ef0a8930691f2b5b5c72f0d164b11c620c77)
- update to hydra-derivatives 3.0 because hydra-collections 6 is needed for BL 6 compatibility:

``` bash
    sufia was resolved to 6.5.0, which depends on
      hydra-batch-edit (~> 1.1) was resolved to 1.1.0, which depends on
        hydra-collections was resolved to 6.0.0, which depends on
          hydra-works (~> 0.4) was resolved to 0.4.0, which depends on
            hydra-derivatives (~> 3.0)
```

mjgiarlo:
For Sufia 7, I've put in #1534 to pin to hydra-head 9.4.1 (because there's no 9.5 release) to get the build working again. I've also decided to make Blacklight 6 support required for the Sufia 7 release. If we can smooth the migration path by taking option 3, that'd be ideal, but that is obviously dependent on someone having cycles to take it on.

awead:
:+1: sounds like a plan!

mjgiarlo:
Seems we've moved past this by now.

PR #1531, opened 2016-01-22
Pin Sufia 6 to hydra-head 9.5

awead:
Sufia 6 is presently broken with the latest hydra-head. Fixing this presents some issues, which I discuss in #1530.


awead:
@mjgiarlo can you merge this, since #1534 is also?

mjgiarlo:
@awead I would have but @jcoyne is less timezone challenged than I am, relative to Eastern anyway. ;) Thx for the PR!

PR #1534, opened 2016-01-25
Do not depend upon components that use BL6.

mjgiarlo:
Hydra-head 9.6.0 and CurationConcerns master have both migrated to Blacklight 6, which is not yet supported in Sufia. We should not depend on these new components until we've made Sufia's master branch support Blacklight 6.


mjgiarlo:
Related to #1531 and #1530.

Issue #1388, opened 2015-12-04
FileSet show page should have a link to "Analytics"


mjgiarlo:
Confirm that https://github.com/projecthydra/sufia/blob/master/spec/views/stats/file.html.erb_spec.rb is linked to from the FileSet show view and tested in a view spec.

https://github.com/projecthydra/sufia/blob/master/app/views/generic_files/_show_actions.html.erb#L5-L8

*Migrated from https://github.com/projecthydra-labs/sufia-core/issues/167*

barmintor:
@mjgiarlo We need to test whether this is satisfied generally by:
https://github.com/projecthydra/sufia/blob/242efd0a53b88ddaca82b7c7c637264de1f872a4/app/views/curation_concerns/base/_show_actions.html.erb#L3-L5

mjgiarlo:
@barmintor That looks about right to me, though does that mean works and filesets will have analytics pages? I'm not sure we have sufficient plumbing to make that be the case at this point.

I believe we may have also discussed this in the context of migration paths and concluded that, if we need to choose between the work and fileset, the work should hold analytics (under the assumption that a GF *becomes* a Work and a new FileSet is *created* to hold the GF's binary content, i.e., the Work takes the GF's identifier so that routes can be setup to transparently 301 from /files/ID to the new Work URI). Though this is complicated because our analytics strategy current includes both pageviews (useful for both filesets and works) and downloads (useful only for filesets, though these numbers could be summed for Work-level download stats).

In short... there's this ticket ("fix the show page") and then there's the all the other work that this ticket may be hiding.

grosscol:
I was under the impression that they were both going to have analytics with:
Work having page views
FileSet having downloads

awead:
I agree with @grosscol

mjgiarlo:
@awead Hrmm, how is what @grosscol said different from what I said?

```
Though this is complicated because our analytics strategy current includes both pageviews (useful for both filesets and works) and downloads (useful only for filesets, though these numbers could be summed for Work-level download stats).
```

:arrow_heading_up: implies works have page views and filesets have downloads, right? Sorry if I misunderstood.

awead:
I guess it isn't. I was just agreeing with him.

mjgiarlo:
OK, just checking!

hectorcorrea:
Are we not going to have views for FileSets (in addition to downloads)?

mjgiarlo:
This is part of #1455.

[...]

PR #1535, opened 2016-01-26
Render File_set page and add dependencies for Analytics

njaffer:
Related to #1388

### DIFF
--- a/lib/sufia/engine.rb
+++ b/lib/sufia/engine.rb
@@ -23,6 +23,9 @@ module Sufia
     initializer 'requires' do
       require 'activerecord-import'
       require 'hydra/derivatives'
+      require 'sufia/analytics'
+      require 'sufia/pageview'
+      require 'sufia/download'
     end
 
     initializer 'configure' do
--- a/app/helpers/sufia/sufia_helper_behavior.rb
+++ b/app/helpers/sufia/sufia_helper_behavior.rb
@@ -114,7 +114,7 @@ module Sufia
 
     def render_visibility_link(document)
       link_to render_visibility_label(document),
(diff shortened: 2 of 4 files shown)


=.bundle/ruby/2.2.0/bundler/gems/curation_concerns-96071bbf6aaf
```diff
@@ -0,0 +1 @@
+Subproject commit 96071bbf6aaf06098fbd75a55905066550cf5ea0
```
jcoyne: This should not be included.

=.bundle/ruby/2.2.0/bundler/gems/kaminari-7f7108e12bdd
```diff
@@ -0,0 +1 @@
+Subproject commit 7f7108e12bdd64eb5a76177633768feb14df5453
```
jcoyne: This should not be included.

lib/sufia/engine.rb
```diff
@@ -23,6 +23,9 @@ class Engine < ::Rails::Engine
     initializer 'requires' do
       require 'activerecord-import'
       require 'hydra/derivatives'
+      require 'sufia/analytics'
```
jcoyne: These should be autoloaded an not required in an initializer.

njaffer: @jcoyne but it does not autoload and without this the Legato fails and gives an error " undefined method sufia__download". If I take out this line require 'sufia/analytics' only, even then I get an error. I have compared this with Sufia 6 and this is what was added in one of the PRs. Yesterday we looked at this with @mjgiarlo

PR #1536, opened 2016-01-26
Render file_set page and add dependencies for Analytics

njaffer:
Related to #1388

### DIFF
--- a/app/views/curation_concerns/base/_show_actions.html.erb
+++ b/app/views/curation_concerns/base/_show_actions.html.erb
@@ -1,7 +1,10 @@
 <div class="form-actions">
   <% if Sufia.config.analytics %>
-    <%= link_to "Analytics", sufia.stats_work_path(@presenter), id: 'stats', class: 'btn btn-default' %>
-  <% end %>
+    <% if @presenter.model_name.name ==  "GenericWork" %>
+      <%= link_to "Analytics", sufia.stats_work_path(@presenter), id: 'stats', class: 'btn btn-default' %>
+    <% else %>
+      <%= link_to "Analytics", sufia.stats_file_path(@presenter), id: 'stats', class: 'btn btn-default' %>
+    <% end %>
   <% if Sufia.config.citations %>
     <%= link_to "Citations", sufia.citations_work_path(@presenter), id: 'citations', class: 'btn btn-default' %>
   <% end %>
@@ -13,16 +16,18 @@
   <%= link_to 'Export to Zotero', sufia.static_path('zotero'), id: 'zoteroLink', name: 'zotero', class: 'btn btn-default' %>
   <%= link_to 'Export to Mendeley', sufia.static_path('mendeley'), id: 'mendeleyLink', name: 'mendeley', class: 'btn btn-default' %>
 
-  <% if @presenter.display_feature_link? %>
-    <%= link_to "Feature", sufia.featured_work_path(@presenter, format: :json),
-      data: {behavior: 'feature'}, id: 'featureLink', name: 'featureLink', class: 'btn btn-default' %>
+  <% if @presenter.model_name.name ==  "GenericWork" %>
+    <% if @presenter.display_feature_link? %> 
+      <%= link_to "Feature", sufia.featured_work_path(@presenter, format: :json),
+        data: {behavior: 'feature'}, id: 'featureLink', name: 'featureLink', class: 'btn btn-default' %>
+    <% end %>
+  
+    <% if @presenter.display_unfeature_link? %>
+      <%= link_to "Unfeature", sufia.featured_work_path(@presenter, format: :json),
+        data: {behavior: 'unfeature'}, id: 'unfeatureLink', name: 'unfeatureLink', class: 'btn btn-default' %>
+    <% end %>
   <% end %>
-
-  <% if @presenter.display_unfeature_link? %>
-    <%= link_to "Unfeature", sufia.featured_work_path(@presenter, format: :json),
-      data: {behavior: 'unfeature'}, id: 'unfeatureLink', name: 'unfeatureLink', class: 'btn btn-default' %>
-  <% end %>
-
+ 
   <!-- COinS hook for Zotero -->
   <span class="Z3988" title="<%= export_as_openurl_ctx_kev(@presenter) %>"></span>
 </div>
--- a/lib/sufia/engine.rb
+++ b/lib/sufia/engine.rb
@@ -23,6 +23,9 @@ module Sufia
     initializer 'requires' do
       require 'activerecord-import'
       require 'hydra/derivatives'
+      require 'sufia/analytics'
+      require 'sufia/pageview'
+      require 'sufia/download'
     end
 
     initializer 'configure' do
--- a/app/models/file_view_stat.rb
+++ b/app/models/file_view_stat.rb
@@ -12,7 +12,7 @@ class FileViewStat < ActiveRecord::Base
   # Sufia::Download is sent to Sufia::Analytics.profile as #sufia__download
   # see Legato::ProfileMethods.method_name_from_klass
   def self.ga_statistics(start_date, file_id)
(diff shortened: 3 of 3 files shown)


app/views/curation_concerns/base/_show_actions.html.erb
```diff
@@ -1,7 +1,10 @@
 <div class="form-actions">
   <% if Sufia.config.analytics %>
-    <%= link_to "Analytics", sufia.stats_work_path(@presenter), id: 'stats', class: 'btn btn-default' %>
-  <% end %>
+    <% if @presenter.model_name.name ==  "GenericWork" %>
```
jcoyne: Instead of hardcoding `GenericWork` can you flip the conditional and check for `FileSet` instead?  I want this view to work for other types of works that are not `GenericWorks`.

app/views/curation_concerns/base/_show_actions.html.erb
```diff
@@ -13,16 +16,18 @@
...
 
-  <% if @presenter.display_feature_link? %>
-    <%= link_to "Feature", sufia.featured_work_path(@presenter, format: :json),
-      data: {behavior: 'feature'}, id: 'featureLink', name: 'featureLink', class: 'btn btn-default' %>
-  <% end %>
-
-  <% if @presenter.display_unfeature_link? %>
-    <%= link_to "Unfeature", sufia.featured_work_path(@presenter, format: :json),
-      data: {behavior: 'unfeature'}, id: 'unfeatureLink', name: 'unfeatureLink', class: 'btn btn-default' %>
+  <% if @presenter.model_name.name ==  "GenericWork" %>
```
jcoyne: Check for `!= "FileSet"` please

lib/sufia/engine.rb
```diff
@@ -23,6 +23,9 @@ class Engine < ::Rails::Engine
     initializer 'requires' do
       require 'activerecord-import'
       require 'hydra/derivatives'
+      require 'sufia/analytics'
```
jcoyne: These are not necessary as these classes are on already on the autoload path.

app/views/curation_concerns/base/_show_actions.html.erb
```diff
@@ -1,7 +1,10 @@
 <div class="form-actions">
   <% if Sufia.config.analytics %>
-    <%= link_to "Analytics", sufia.stats_work_path(@presenter), id: 'stats', class: 'btn btn-default' %>
-  <% end %>
+    <% if @presenter.model_name.name ==  "GenericWork" %>
+      <%= link_to "Analytics", sufia.stats_work_path(@presenter), id: 'stats', class: 'btn btn-default' %>
```
mjgiarlo: I might like to see this code in the presenter instead of the view. So the view would just have:

```ruby
<%= link_to "Analytics", @presenter.stats_path, id: 'stats', class: 'btn btn-default' %>
```

And the presenter would have, e.g.:

```ruby
def stats_path
  case model_name.name
  when 'GenericWork'
    sufia.stats_work_path(self)
  when 'FileSet'
    sufia.stats_file_path(self)
  else
    raise 'handle this better'
  end
end
```

app/views/curation_concerns/base/_show_actions.html.erb
mjgiarlo: Instead of checking the name of the model (identity), can you check the behavior? So replace this line with:

```ruby
<% if @presenter.respond_to(:display_feature_link?) %
```

That gives us more flexibility down the line, allowing users to generate other curation concerns with their own presenters and display the feature/unfeature links if supported.

lib/sufia/engine.rb
njaffer: @jcoyne, these are needed for viewing the pageviews and downloads. If I do not put these here, I get an error for Legato::Models

mjgiarlo: @jcoyne The problem this line is trying to solve is that we need the Legato library available in three different files: analytics.rb, pageview.rb, and download.rb. These lines make sure this line is available to those three files: https://github.com/projecthydra/sufia/blob/d515cd314a2d5e9e8179302e69824299e83e43b2/app/services/sufia/analytics.rb#L3

I'm sure @njaffer is open to others ways to handle this.

jcoyne: How about we just `require 'legato'` here and remove the require from `app/services/sufia/analytics.rb`?

mjgiarlo: That'd work for me. @njaffer ?

njaffer: @jcoyne , @mjgiarlo I tried it but it does not work!
I removed the 3 lines from engine.rb and added require 'legato' in engine.rb
and removed require 'legato' from analytics.rb. Getting the same error:
undefined method `sufia__download' for #<Legato::Management::Profile:0x007f85f68b12e8>

mjgiarlo: @njaffer What if you move the require out of engine.rb and put it in this block o' requires: 

https://github.com/projecthydra/sufia/blob/master/lib/sufia.rb#L1-L23

:question:

njaffer: @mjgiarlo , nope it does not work either :-(

njaffer: @jcoyne , @mjgiarlo , in rails console you can check if you get good data for Sufia::Analytics.Profile.sufia__download

I get data only when I add the three lines in engine.rb ( I am not saying that there might not be another way) but this is what is working for me and in Sufia 6 (https://github.com/projecthydra/sufia/blob/6.x-stable/sufia-models/lib/sufia/models/engine.rb#L71)

mjgiarlo: Hmmm. What if we add `#{config.root}/app/models/sufia` and `#{config.root}/app/services/sufia` to `config.autoload_paths` here:

https://github.com/projecthydra/sufia/blob/master/lib/sufia/engine.rb#L8-L12

app/views/curation_concerns/base/_show_actions.html.erb
njaffer: @mjgiarlo , which presenter file_set or works ?

mjgiarlo: @njaffer Oh, right. Derp. Below is what I should have typed.

Let's put the stats path code in the presenters, like so:

```ruby
# WorkShowPresenter
def stats_path
    sufia.stats_work_path(self)
end
```

and 

```ruby
# FileSetPresenter
def stats_path
    sufia.stats_file_path(self)
end
```

njaffer: @mjgiarlo , how about polymorphic_path? with the code above, it fails with the error: undefined local variable or method `sufia'

mjgiarlo: @njaffer Ah, right, you won't have access to that "sufia." object outside of views and controllers. I tweaked the code above to reference the paths properly.

lib/sufia/engine.rb
jcoyne: I think everything under config/app is already in the autoload paths. Certainly app/models.

mjgiarlo: Everything, as in, even stuff in random subdirectories? I thought that was true too but in https://github.com/projecthydra/sufia/blob/master/lib/sufia/engine.rb#L8-L12 we have `app/models/concerns` listed. Could be cruft of course.

jcoyne: I think that was for compatiblity with older versions of rails (Pre  4.2?)  that didn't have the concept of concerns.

mjgiarlo: @jcoyne @njaffer OK, I worked on this a bit and wrote it up. See here:

https://github.com/projecthydra/sufia/pull/1540#issue-129226479

Note that adding `app/models/sufia` did **not** help at all.

njaffer:
related PR #1540 is merged

PR #1537, opened 2016-01-26
Displays processing icon on works...

### LINKED ISSUE #1123: Processing on a work
Should a work have the concept of in process in the dashboard?

If so do we collect up the file processing.  Does the work have it on it's own?

hectorcorrea:
...that are still being processed. Fixes #1123

### DIFF
--- a/app/controllers/concerns/sufia/works_controller_behavior.rb
+++ b/app/controllers/concerns/sufia/works_controller_behavior.rb
@@ -17,6 +17,12 @@ module Sufia
       super
     end
 
+    def edit
+      work = GenericWork.find(params[:id])
+      throw "Cannot edit a work that still is being processed" if work.processing?
+      super
+    end
+
     protected
 
       def show_presenter
--- a/app/views/collections/_show_document_list_row.html.erb
+++ b/app/views/collections/_show_document_list_row.html.erb
@@ -1,12 +1,10 @@
 <% id = document.id %>
+<% presenter = document.is_a?(GenericWork) ? Sufia::WorkShowPresenter.new(document, nil) : nil %>
 <tr id="document_<%= id %>" class="<%= cycle("", "zebra") %>">
   <td>&nbsp;
-    <%# TODO is there a processing concept for a work ?? see #1122 & #1123 %>
-    <%# This block is for adding/removing the magic wand while batch updates are processing %>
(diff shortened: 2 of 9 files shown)


hectorcorrea:
When viewing works inside a collection
screen shot 2016-01-26 at 2 56 50 pm

When viewing a work (that still is being processed)
screen shot 2016-01-26 at 2 57 04 pm

When viewing "My Works"
screen shot 2016-01-26 at 2 57 18 pm

hectorcorrea:
@mjgiarlo @jcoyne: ping

app/views/collections/_show_document_list_row.html.erb
```diff
@@ -1,12 +1,10 @@
 <% id = document.id %>
+<% presenter = document.is_a?(GenericWork) ? Sufia::WorkShowPresenter.new(document, nil) : nil %>
```
grosscol: Is there already a presenter defined here that is the presenter of the collection?

hectorcorrea: What makes this code tricky is that this view might render Collections or Works.

I updated to code to use a presenter when rendering Works, but didn't mess with the code that render Collections and that part remains untouched.

grosscol: So we'll need another ticket to presenter-ify this code for a collection presenter?

app/controllers/concerns/sufia/works_controller_behavior.rb
```diff
@@ -17,6 +17,12 @@ def new
       super
     end
 
+    def edit
+      work = GenericWork.find(params[:id])
+      throw "Cannot edit a work that still is being processed" if work.processing?
```
mjgiarlo: @hectorcorrea @grosscol @jcoyne Should this be `raise` instead of `throw`? I don't see a `catch` block around it. Is there a test that makes sure `"Cannot edit a work that still is being processed"` is ever used?

hectorcorrea: I'll submit a separate PR with the test and update this to a raise.

mjgiarlo: @hectorcorrea :clap:

PR #1538, opened 2016-01-27
fix call to get path to generic work in partial.

grosscol:
Fixes an error in rendering the edit profile page when a user has highlighted works.  Call signature for generic_file_path just needed to be updated.

### DIFF
--- a/app/views/users/_trophy_edit.html.erb
+++ b/app/views/users/_trophy_edit.html.erb
@@ -6,7 +6,7 @@
   <dl>
     <% trophies.each do |t| %>
       <dd>
-        <%= link_to t, generic_work_path(t) %>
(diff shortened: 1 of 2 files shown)


mjgiarlo:
Travis is sad. Can you also write a test that fails if you revert your one-line fix so that we can make sure we don't run into regressions later? Thx for the PR, @grosscol

grosscol:
@mjgiarlo I'm been working on that currently. The trophy form does not render properly in the edit profile page.  So a test to ensure proper rendering will follow as soon as the rendering is rendered properly.

grosscol:
Couldn't get the UI to look pretty despite dropping a large number of F :bomb: while hacking at it.  Since the UI is getting a once over this year, I'll make a backlog ticket for fixing the UI.

Issue #1455, opened 2015-12-11
Revisit analytics implementation to account for new FileSet/GenericWork split


jcoyne:
now that we've moved to works/file_sets, what analytics should be collected? On what events?

awead:
Related tickets #1419 #1388

mjgiarlo:
Here's where I stand: we should collect pageviews for the work (on hits to the work show page) and we should collect downloads for on the fileset (on clicks that result in downloads, whether from the work show page, fileset show page, or catalog pages if downloading is accessible there). We should continue to cache analytics in the application database. We should present all analytics information in the `/works/{id}/stats` view. Let me know if I missed anything.

gordonleacock:
Okay, I'm going to start poking at this one. However, I can not currently assign myself to it.

mjgiarlo:
@gordonleacock I've added you to the GitHub team now. Thanks!

gordonleacock:
Jose and I worked on this today and discovered that the _show_actions.html.erb partial was trying to use if @presenter.display_feature_link? and if @presenter.display_unfeature_link?. Those were causing sufia to die with a method not found error. For now, we have commented them out. Do folks agree that they don't belong in this partial code in sufia 7?

jcoyne:
@gordonleacock I think @hectorcorrea just added that code, so I suspect it does belong in Sufia 7.

gordonleacock:
@mjgiarlo I did a fresh pull this afternoon and it was dying (no method found) on line 16: 
<% if @presenter.display_feature_link? %>

We'll refresh our code tomorrow and try it again. Thanks.

hectorcorrea:
@gordonleacock this is the code where the `display_feature_link?` and `display_unfeature_link?` were introduced a few days ago: https://github.com/projecthydra/sufia/pull/1503/files

gordonleacock:
We didn't touch "app/presenters/sufia/work_show_presenter.rb". We changed the partial in "app/views/curations_concerns/base/_show_actions.html.erb".

gordonleacock:
I've pulled down the code and rechecked the problem. It still exists. Unlike sufia/app/presenters/work_show_presenter.rb, the file sufia/app/presenters/file_set_presenter.rb does not have the methods *display_feature_link?* and *display_unfeature_link?*.

[...]

PR #1517, opened 2016-01-15
Revised file_set presenter and show per @jcoyne's suggestion

gordonleacock:
1. Added new show_actions partial under views/curation_concerns/file_sets directory.
2. Disabled (for now): @stats = FileUsage.new(params[:id]), in stats_controller.rb.
Result: when analytics button is clicked on the file_sets show page, one is taken to a mostly blank analytics page for file_sets.

PR #1540, opened 2016-01-27
Render file_set page and add dependencies for Analytics. Refs #1388 #1455

mjgiarlo:
Supersedes #1536.

### DIFF
--- a/app/views/curation_concerns/base/_show_actions.html.erb
+++ b/app/views/curation_concerns/base/_show_actions.html.erb
@@ -1,6 +1,6 @@
 <div class="form-actions">
   <% if Sufia.config.analytics %>
-    <%= link_to "Analytics", sufia.stats_work_path(@presenter), id: 'stats', class: 'btn btn-default' %>
+    <%= link_to "Analytics", @presenter.stats_path, id: 'stats', class: 'btn btn-default' %>
   <% end %>
   <% if Sufia.config.citations %>
     <%= link_to "Citations", sufia.citations_work_path(@presenter), id: 'citations', class: 'btn btn-default' %>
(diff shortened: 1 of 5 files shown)


app/views/curation_concerns/base/_show_actions.html.erb
```diff
@@ -12,17 +12,14 @@
   <%= link_to 'Export to EndNote', polymorphic_path([main_app, @presenter], format: 'endnote'), class: 'btn btn-default' %>
   <%= link_to 'Export to Zotero', sufia.static_path('zotero'), id: 'zoteroLink', name: 'zotero', class: 'btn btn-default' %>
   <%= link_to 'Export to Mendeley', sufia.static_path('mendeley'), id: 'mendeleyLink', name: 'mendeley', class: 'btn btn-default' %>
-
-  <% if @presenter.display_feature_link? %>
-    <%= link_to "Feature", sufia.featured_work_path(@presenter, format: :json),
-      data: {behavior: 'feature'}, id: 'featureLink', name: 'featureLink', class: 'btn btn-default' %>
+  <% if @presenter.respond_to?(:display_feature_link?) && @presenter.display_feature_link? %>
```
jcoyne: Why are we checking respond_to? here?  Are there any presenters that should work that don't implement this?

lib/sufia/engine.rb
```diff
@@ -11,6 +11,11 @@ class Engine < ::Rails::Engine
       #{Hydra::Engine.root}/app/models/concerns
     )
 
+    config.eager_load_paths += %W(
```
jcoyne: Please add a comment about why this is necessary. e.g. `Force these models to be added to Legato's registry in development mode`

sufia.gemspec
```diff
@@ -16,7 +16,7 @@ Gem::Specification.new do |spec|
   spec.version       = version
   spec.license       = 'Apache2'
 
-  spec.add_dependency 'curation_concerns', '~> 0.4'
+  spec.add_dependency 'curation_concerns', '~> 0.5', '>= 0.5.0'
```
jcoyne: Isn't that third argument just a less restrictive version of the second?

app/views/curation_concerns/base/_show_actions.html.erb
mjgiarlo: The FileSetPresenter does not implement these methods. @gordonleacock started working on that PR here: https://github.com/projecthydra/sufia/pull/1517

lib/sufia/engine.rb
mjgiarlo: Done.

sufia.gemspec
mjgiarlo: I mistakenly thought that '~> 0.5' might allow 0.4, but no, that's not right. Removing that argument.

app/views/curation_concerns/base/_show_actions.html.erb
jcoyne: How about we use different views then. I just heard Sandi Metz talk. Now I'm condition averse :)

mjgiarlo:
Travis will be :unamused: until Capybara 2.6.2 comes out (2.6.1 broke stuff). This is why: https://github.com/jnicklas/capybara/issues/1644#issuecomment-175891194

app/views/curation_concerns/base/_show_actions.html.erb
mjgiarlo: Can you issue-fy that for another day? ;)

mjgiarlo:
Capybara 2.6.2 is out. Restarting the builds.

mjgiarlo:
@jcoyne @njaffer :shipit: ?

app/views/curation_concerns/base/_show_actions.html.erb
njaffer: @mjgiarlo, @jcoyne: how about we just revert this back to checking  !=FileSet

mjgiarlo: @njaffer My quibble with checking the class name, rather than checking the behavior, is that we haven't said "FileSets will never be able to be highlighted," so why not build in the extra flexibility? I've seen over time that the assumptions I make don't necessarily jibe with those made by other Sufia adopters -- so if some adopter wants to make FileSets featurable, this lets them do that without requiring that they override one or more views.

njaffer: @mjgiarlo , ok that's understandable, but at the same time thinking about the simplicity, it might be good to keep it simple for now and when future adopters might need to update, that can be done later

mjgiarlo: I thought it was simpler, @njaffer! Ha. Maybe it's one of those things. :)

PR #1541, opened 2016-01-28
update background workers doc

elrayle:
Please review for accuracy.  Added additional information on message queues in general, resque message queue specifically as used in Sufia, configuration, and validating running processes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,8 +26,13 @@
     * [Install Sufia](#install-sufia)
     * [Database tables and indexes](#database-tables-and-indexes)
     * [Solr and Fedora](#solr-and-fedora)
-    * [Start background workers](#start-background-workers)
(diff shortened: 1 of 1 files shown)


hectorcorrea:
:+1:

jcoyne:
I'm wondering if we want to continue to be Resque specific. Might someone want to use sidekiq instead?

elrayle:
@jcoyne I think we could definitely expand our documentation more.  If this documentation had existed, it would have saved me a lot of time in debugging.  If someone uses sidekiq and wants to add similar documentation for that, it would be great.

mjgiarlo:
:+1: and I would welcome -- nay, **encourage** -- someone who swapped out Resque for Sidekiq to contribute said documentation back to the project!

PR #1545, opened 2016-01-28
Added controller test for editting works in process.

hectorcorrea:
Also replaced `throw` with `raise` since that's more inline with the way exceptions are handled in Ruby:
http://stackoverflow.com/questions/51021/what-is-the-difference-between-raising-exceptions-vs-throwing-exceptions-in-ruby

These changes were suggested after merging PR #1537.

@mjgiarlo as you suggested.


awead:
:+1:

PR #1548, opened 2016-02-01
Escape property values before presenting them with auto_link - Sufia 6

### LINKED ISSUE #1546: Various fields are not displayed correctly if they contain a less than character
Several fields are not shown correctly in the details page if they contain data with a less than character (<). Fields that are having problems include at least description, publisher and language and possibly others. In Sufia 6 title is also not shown correctly in the details page. Sufia 7 has no issues with title containing a less than character but the other fields still are having problems.

In our Sufia 6 based system we have a Generic File with title "something something with <30kHz frequency something" and it's displayed as "something something with" in the details page. The title is fine in the record itself if you look at it from the console. And it's displayed correctly in the Blacklight based pages too.

The problem seems to be ActionView::Helpers::TextHelper#auto_link. It will choke on the < character and drop off anything following that. You can pass sanitize: false to auto_link and then it'll accept the < character but then the output will not be html_safe. 

I think this same thing affects CurationConcerns.

ojlyytinen:
Fixes #1546 for Sufia 6.

### DIFF
--- a/app/helpers/generic_file_helper.rb
+++ b/app/helpers/generic_file_helper.rb
@@ -28,7 +28,7 @@ module GenericFileHelper
   end
 
   def display_multiple(value)
-    auto_link(value.join(" | "))
+    auto_link(html_escape(value.join(" | ")))
   end
 
   private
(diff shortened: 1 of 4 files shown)


mjgiarlo:
Looks good to me. :+1:

mjgiarlo:
Thanks, @ojlyytinen

PR #1554, opened 2016-02-04
Comply with Rubocop 0.37

mjgiarlo:


.rubocop.yml
```diff
@@ -49,6 +49,12 @@ Metrics/ModuleLength:
     - 'app/controllers/concerns/sufia/files_controller_behavior.rb'
     - 'app/helpers/sufia/sufia_helper_behavior.rb'
 
+Performance/RedundantMerge:
+  Enabled: false
+
+Style/ZeroLengthPredicate:
+  Enabled: false
+
```
awead: Seems reasonable. What do we miss by turning this off?

mjgiarlo: If we leave this on and run autocorrect, we get test failures. These are both arguably buggy corrections because of the assumptions made. Rubocop is seeing `whatever[:something]` and assuming `whatever` is a hash, which causes the RedundantMerge violation. And ZeroLengthPredicate assumes that if an object supports `#size` or `#length` that it also supports `#empty?`, and that's not always true. @jcoyne has already created GitHub issues in the Rubocop repo for both of these new violations.

awead:
@mjgiarlo thanks for the explanations. I'm :+1:

mjgiarlo:
@awead :shipit: ?

Issue #1552, opened 2016-02-03
Transfer work gives an error if cancelled


njaffer:
1. I tried to transfer the work ownership to another user and then cancelled the request 
2. Logged in as the second user, it showed me the notification that the work transfer request was initiated and cancelled
3. When I click on the work name, I get an error: undefined method `human_readable_type' for nil:NilClass in unauthorized.html.erb

njaffer:
![screen shot 2016-02-03 at 2 59 15 pm](https://cloud.githubusercontent.com/assets/12487929/12795249/078d9722-ca87-11e5-8354-91e1bcd0bdf0.png)

mjgiarlo:
Good find! Thanks, @njaffer

njaffer:
closed with #1557

PR #1557, opened 2016-02-05
Do not show 'work link' if work transfer is cancelled'

njaffer:
related to #1552

### DIFF
--- a/app/helpers/sufia/sufia_helper_behavior.rb
+++ b/app/helpers/sufia/sufia_helper_behavior.rb
@@ -31,7 +31,7 @@ module Sufia
     end
 
     def show_transfer_request_title(req)
-      if req.deleted_work?
+      if req.deleted_work? || req.canceled?
         req.to_s
       else
         link_to(req.to_s, curation_concerns_generic_work_path(req.generic_work_id))
--- a/app/models/proxy_deposit_request.rb
+++ b/app/models/proxy_deposit_request.rb
@@ -75,6 +75,10 @@ class ProxyDepositRequest < ActiveRecord::Base
     save!
   end
 
+  def canceled?
+    self.status == 'canceled'
+  end
+
   def deleted_work?
     !GenericWork.exists?(generic_work_id)
   end
--- a/spec/helpers/sufia_helper_spec.rb
+++ b/spec/helpers/sufia_helper_spec.rb
@@ -1,6 +1,22 @@
 require 'spec_helper'
 
 describe SufiaHelper, type: :helper do
+  describe "show_transfer_request_title" do
+    let(:sender) { create(:user) }
+    let(:user) { create(:user) }
+    let(:work) do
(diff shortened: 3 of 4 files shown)


app/helpers/sufia/sufia_helper_behavior.rb
```diff
@@ -31,7 +31,7 @@ def error_messages_for(object)
     end
 
     def show_transfer_request_title(req)
-      if req.deleted_work?
+      if req.deleted_work? || req.status != 'Canceled'
```
jcoyne: Have you considered making this a method on the `req` object?

njaffer: Sure, I will update the code

app/models/proxy_deposit_request.rb
```diff
@@ -75,6 +75,10 @@ def cancel!
     save!
   end
 
+  def is_canceled?
+    self.status == 'canceled'
```
jcoyne: Let's make `'canceled'` a constant (and use it in `cancel!`) too.  Also, can you add a test to this PR?

app/models/proxy_deposit_request.rb
```diff
@@ -75,6 +75,10 @@ def cancel!
     save!
   end
 
+  def is_canceled?
```
jcoyne: The ruby convention is just `canceled?` without having `is_` at the front.

njaffer:
@jcoyne: The test already covers the canceled status here: https://github.com/projecthydra/sufia/blob/master/spec/models/proxy_deposit_request_spec.rb#L63

and for making 'canceled' as a constant, we will have to create all other status as constants also, how about we make that as a separate ticket?

jcoyne:
There is no test of the behavior of the new method or how it affects the view.

njaffer:
@jcoyne: Is there an existing test that needs to be modified? I cannot seem to find where would this test go?

jcoyne:
@njaffer I'd put one test in here: https://github.com/njaffer/sufia/blob/transfer/spec/models/proxy_deposit_request_spec.rb  and then one in here: https://github.com/projecthydra/sufia/blob/master/spec/helpers/sufia_helper_spec.rb  I did mistype before, I said view, but I actually meant helper.

awead:
:ok_hand: to me, but @jcoyne might want a test in `sufia_helper_spec.rb`.

njaffer:
@awead , @jcoyne : I do not see any relevant test for testing the link for work name in sufia_helper_spec.rb 

The test I added in proxy_deposit_request_spec.rb tests the code I have added in the proxy_deposit_request file

jcoyne:
@njaffer Do you think it might be important to test the behavior of show_transfer_request_title when the status of the request is canceled? If you don't add a test here your change might be reverted without anyone knowing this was important behavior.

njaffer:
@jcoyne: added the test to make sure no one can put title as a link when the transferred work is cancelled!

PR #1566, opened 2016-02-10
fix analytics path for file_set and work

njaffer:


jcoyne:
This looks good, but per the guidelines in CONTRIBUTING.md, tests are required.  Would you please add some?

njaffer:
@jcoyne: added the tests

spec/presenters/sufia/file_set_presenter_spec.rb
```diff
@@ -4,7 +4,17 @@
   let(:solr_document) { SolrDocument.new(file.to_solr) }
   let(:ability) { double "Ability" }
   let(:presenter) { described_class.new(solr_document, ability) }
-  let(:file) { build(:file_set).tap { |f| f.apply_depositor_metadata(user) } }
+  let(:file) do
+    build(:file_set).tap do |f|
+      f.id = '123abc'
+      f.apply_depositor_metadata(user)
```
jcoyne: Instead of this block, just pass args to `build(:file_set, id: '123abc', user: user)`

spec/presenters/sufia/file_set_presenter_spec.rb
```diff
@@ -4,7 +4,16 @@
   let(:solr_document) { SolrDocument.new(file.to_solr) }
   let(:ability) { double "Ability" }
   let(:presenter) { described_class.new(solr_document, ability) }
-  let(:file) { build(:file_set).tap { |f| f.apply_depositor_metadata(user) } }
+  let(:file) do
+    build(:file_set, id: '123abc').tap do |f|
```
jcoyne: You can supply `user: user` as a parameter and then drop the block.  It needs to respond to `user_key` though, so `user: double(user_key: 'sarah')` should work.

njaffer: @jcoyne, If I do that, the stats_path test passes but other tests fails

PR #1569, opened 2016-02-11
fix visibility badge link and anchor

jcoyne:
add plymorphic_path

update spec test

Supercededs #1505


jcoyne:
@njaffer I rewrote the test in this version such that it doesn't hit Fedora.

njaffer:
@jcoyne: Thank you!

PR #1582, opened 2016-02-16
Update README.md to quickstart with actual app instead of test app

barmintor:


barmintor:
User report:

I got "NameError: uninitialized constant SolrWrapper" after running rake hydra:server

awead:
@barmintor try adding "require solr_wrapper"

barmintor:
Looks like user error.

PR #1590, opened 2016-02-19
add processing method skeleton to FileSet Presenter

njaffer:
Fixes error: 
undefined method `processing?' for #<Sufia::FileSetPresenter:0x007ff9b4532ab8>

The processing method was added for UploadSet but not for FileSet. I have added a method, with comments as TODO, for now the app will not crash on viewing FileSets 

Error generated on this line: https://github.com/projecthydra/sufia/blob/master/app/views/curation_concerns/base/_show_actions.html.erb#L8

* Add a skeleton method for processing in FileSetPresenter
* 
* 

### DIFF
--- a/app/presenters/sufia/file_set_presenter.rb
+++ b/app/presenters/sufia/file_set_presenter.rb
@@ -36,6 +36,10 @@ module Sufia
       'http://schema.org/CreativeWork'
     end
 
+    def processing?
+      # TODO: Refactor this away per https://github.com/projecthydra/sufia/pull/1592
+    end
+    
     def stats_path
       Sufia::Engine.routes.url_helpers.stats_file_path(self)
     end
--- a/spec/presenters/sufia/file_set_presenter_spec.rb
+++ b/spec/presenters/sufia/file_set_presenter_spec.rb
@@ -6,6 +6,11 @@ describe Sufia::FileSetPresenter do
   let(:presenter) { described_class.new(solr_document, ability) }
   let(:file) { build(:file_set, id: '123abc', user: user) }
 
+  describe 'processing?' do
+    let(:user) { double(user_key: 'sarah') }
+    it { expect(presenter.processing?).to eq nil }
+  end
+
   describe 'stats_path' do
     let(:user) { double(user_key: 'sarah') }
     it { expect(presenter.stats_path).to eq Sufia::Engine.routes.url_helpers.stats_file_path(id: file) }


jcoyne:
@njaffer can you say what calls processing? and causes the  `undefined method 'processing?' for #Sufia::FileSetPresenter:0x007ff9b4532ab8`?

njaffer:
@jcoyne: The view calls presenter.processing? in this file: https://github.com/projecthydra/sufia/blob/master/app/views/curation_concerns/base/_show_actions.html.erb#L8

mjgiarlo:
Can you check in Sufia 6.x to see which classes used the processing? method? I thought it was just UploadSets, but I may be wrong (or this may have changed in Sufia master). I'd dig myself but I've got some other things on my plate today.

jcoyne:
This seems like a good fix, but I'd like to see a unit test added to the PR too.

njaffer:
@mjgiarlo: I don't see processing? used by generic_file presenter (the upload_set used to be generic file??)
@jcoyne: adding the test now..

njaffer:
@mjgiarlo,  @jcoyne: the tests are passing!

mjgiarlo:
As far as I can see, GenericWorks are the only objects that support `#processing?`. And the impl is to check the status of any UploadSets to which the work belongs. I suppose it's OK to allow FileSets to be asked if they're processing, if the method always returns `nil`.

mjgiarlo:
What seems odd to me is that we have two impls of `#processing?`. First it's implemented in a module that gets mixed into every GenericWork:

https://github.com/projecthydra/sufia/blob/master/app/models/concerns/sufia/belongs_to_upload_sets.rb#L8-L11

And then it's implemented in the WorkShowPresenter:

https://github.com/projecthydra/sufia/blob/master/app/presenters/sufia/work_show_presenter.rb#L36-L40

@jcoyne Do we want/need two impls here, or should one delegate to the other? They more or less are doing the same function.

jcoyne:
@mjgiarlo Presenters contain SolrDocuments, not Works. So, yes it's a bit odd. I'm not sure what can be done about it.

jcoyne:
I know this is not the first bug that we've seen with app/views/curation_concerns/base/_show_actions.html.erb being used for FileSets and Works.  Maybe there should just be a app/views/curation_concerns/file_sets/_show_actions.html.erb.  Then we wouldn't have to have the WorkPresenter and FileSetPresenter conform to the same interface.  They really seem pretty different to me.

mjgiarlo:
I'd rather see two different sets of similar templates than one set of templates with a lot of conditionals or an artificial coercion of two different presenters to be similar when they're really not, so I'm :+1: on that.

mjgiarlo:
@jcoyne It looks like the only place in the codebase where we ask GenericWorks (vs. SolrDocuments) directly whether they are processing or not is here:

https://github.com/projecthydra/sufia/blob/master/app/controllers/concerns/sufia/works_controller_behavior.rb#L20-L24

We could ask the presenter there, like: `show_presenter.new(curation_concern, current_ability).processing?` Or does that entangle the concerns in a way that's worrisome? Either way, I think where I'm going with this is verging on work that's not directly related to helping @njaffer get past what she's seeing...

jcoyne:
@mjgiarlo yeah, that seems gross, plus it would have to be:
```ruby
show_presenter.new(SolrDocument.new(curation_concern.to_solr), current_ability).processing?
```

mjgiarlo:
Yeah, gross indeed. OK, I'm now fine with the duplication. ;)

njaffer:
@mjgiarlo: Can you please recap your conversation with @jcoyne. Are you saying duplication of views?

mjgiarlo:
@njaffer You bet! Yeah, I was :+1:ing @jcoyne's suggestion here:

> I know this is not the first bug that we've seen with app/views/curation_concerns/base/_show_actions.html.erb being used for FileSets and Works. Maybe there should just be a app/views/curation_concerns/file_sets/_show_actions.html.erb. Then we wouldn't have to have the WorkPresenter and FileSetPresenter conform to the same interface. They really seem pretty different to me.

What do you think about that?

njaffer:
@mjgiarlo: Yes I agree with @jcoyne that this is not the first bug we have seen with app/views/curation_concerns/base/_show_actions.html.erb, but at the same time I think we have worked hard before at least 2 times I remember, one was for stats path and probably the other one for citation and we worked through it, so may be it would be a good idea for now to just add a processing? method as an empty method and refactor it later after more discussion. As of now, Sufia has a bug which needs to be fixed soon and for refactoring we can create an issue.

mjgiarlo:
@njaffer I'm fine with that too. Would you mind putting in a new issue for the backlog to consider refactoring this at a later date? (Feel free to link into this discussion.) That way we unblock your PR and flag this for later work.

njaffer:
@mjgiarlo: sounds great!
here is the issue for refactoring: https://github.com/projecthydra/sufia/issues/1592

app/presenters/sufia/file_set_presenter.rb
```diff
@@ -36,6 +36,10 @@ def itemtype
       'http://schema.org/CreativeWork'
     end
 
+    def processing?
+      # TODO: Add Status to FileSet to track the staus of processing
```
mjgiarlo: Can you tweak this comment to be:

`# TODO: Refactor this away per https://github.com/projecthydra/sufia/pull/1592`

:question:

Issue #1287, opened 2015-08-20
Rights field should be single-valued by default, and overrideable


awead:

jcoyne:
I'm not sure this is *always* true. Probably for the 99% though.

mjgiarlo:
I suspect it's always true for folks who aren't overriding any of the select values for that field.  And if they're already overriding that... :smile: 

I don't think it makes any sense to let folks select multiple CC licenses, or to select a CC license and "All rights reserved." The default select values suggest single-valued field to me.

vantuyls:
The only places I've seen someone assign multiple (CC) licenses to an object, I'm pretty sure they screwed up. I've been racking my brain for the past week trying to think of a case where one could actually assign multiple CC licenses and i've come up with nothing.

jcoyne:
Yes, my exception case was licenses from the library, not CC licenses.

mjgiarlo:
I spent a day or two hacking on this but I need to move onto other things.

vantuyls:
Hrm. We just made this happen locally.

mjgiarlo:
Maybe the magic incantation is in here: https://github.com/osulp/Scholars-Archive/commit/746e02a0e5fccc85217450ae560d0191500900b9

Thanks for the heads-up, @vantuyls 

@straleyb if you want to toss a fix upstream, that'd be great!

straleyb:
Yeah, I'll update and send a PR. If not this afternoon, tomorrow morning. +1

vantuyls:
INTERNET!

mjgiarlo:
:clap: :fireworks:  @vantuyls @straleyb

mjgiarlo:
Are you still interested in working on this @straleyb?

mjgiarlo:
Paging @straleyb and @vantuyls. :)

If you are interested in contributing this work back, please do issue the PR against the `6.x-stable` branch.

newmanld:
We're taking a look at making this field single valued in our local implementation.  Multiple rights statements didn't make sense to us, and seemed a good use case for multi-part works.  I'll refer the developer who has taken our local issue to this thread.

PR #1596, opened 2016-02-22
update rights input field selector

njaffer:
The right property does not get saved when a New Work is created or edited. 

Here is the debugger output before the change
byebug
    7:     # set the @files ivar then remove the files attribute so it isn't set by default.
=>  8:     files && attributes.delete(:files)
    9:     self.raw_attributes = attributes.dup
   10:     # Files must be attached before saving in order to persist their relationship to the work
   11:     assign_pid && interpret_visibility && attach_files && super && assign_representative
   12:   end
(byebug) attributes
{"title"=>["Test Work 333"], "creator"=>[], "contributor"=>[], "description"=>[], "tag"=>[], "publisher"=>[], "date_created"=>[], "subject"=>[], "language"=>[], "identifier"=>[], "based_near"=>[], "related_url"=>[], "visibility_during_embargo"=>"restricted", "embargo_release_date"=>"2016-02-23", "visibility_after_embargo"=>"open", "visibility_during_lease"=>"open", "lease_expiration_date"=>"2016-02-23", "visibility_after_lease"=>"restricted", "visibility"=>"restricted", "resource_type"=>[]}

Here is the debugger output after the change
(byebug) attributes
{"title"=>["Test Work 333"], "creator"=>[], "contributor"=>[], "description"=>[], "tag"=>[], "rights"=>["http://creativecommons.org/licenses/by-nc/3.0/us/"], "publisher"=>[], "date_created"=>[], "subject"=>[], "language"=>[], "identifier"=>[], "based_near"=>[], "related_url"=>[], "visibility_during_embargo"=>"restricted", "embargo_release_date"=>"2016-02-23", "visibility_after_embargo"=>"open", "visibility_during_lease"=>"open", "lease_expiration_date"=>"2016-02-23", "visibility_after_lease"=>"restricted", "visibility"=>"restricted", "resource_type"=>[]}

* override multiple? in WorkForm, set rights to true
* 
* 

### DIFF
--- a/app/forms/sufia/forms/work_form.rb
+++ b/app/forms/sufia/forms/work_form.rb
@@ -7,5 +7,10 @@ module Sufia::Forms
                :visibility_after_embargo, :visibility_during_lease,
                :lease_expiration_date, :visibility_after_lease, :visibility, :thumbnail_id, :representative_id, :ordered_member_ids]
     end
+
+    def self.multiple?(term)
+      return true if term == :rights
+      super
+    end
   end
 end
--- a/spec/form/work_form_spec.rb
+++ b/spec/form/work_form_spec.rb
@@ -1,0 +1,28 @@
+require 'spec_helper'
+
+describe CurationConcerns::GenericWorkForm do
+  let(:work) { stub_model(GenericWork, id: 'abc123') }
+  let(:curation_concern) { create(:work) }
+  let(:ability) { nil }
+  let(:form) { described_class.new(curation_concern, ability) }
+
+  describe '.model_attributes' do
+    let(:params) { ActionController::Parameters.new(
+      title: ['foo'],
+      description: [''],
+      visibility: 'open',
+      admin_set_id: '123',
+      representative_id: '456',
+      thumbnail_id: '789',
+      tag: ['derp'],
+      rights: ['http://creativecommons.org/licenses/by/3.0/us/']) }
+    subject { described_class.model_attributes(params) }
+    it 'permits parameters' do
+      expect(subject['title']).to eq ['foo']
+      expect(subject['description']).to be_empty
+      expect(subject['visibility']).to eq 'open'
+      expect(subject['rights']).to eq ['http://creativecommons.org/licenses/by/3.0/us/']
+      expect(subject['tag']).to eq ['derp']
+    end
+  end
+end


app/views/records/edit_fields/_rights.html.erb
```diff
@@ -1,4 +1,4 @@
-<%= f.input :rights, as: :select_with_modal_help,
```
cam156: @njaffer The modal help allows for you to display more information about the licenses: 
screen shot 2016-02-22 at 11 05 25 am

I'm not sure why that is not working, but removing it seems like a bad idea.

cam156: You are removing the help ? from the screen by removing that class.
screen shot 2016-02-22 at 11 07 11 am

app/views/records/edit_fields/_rights.html.erb
```diff
@@ -1,4 +1,4 @@
...
+<%= f.input :rights, as: :select,
     collection: RightsService.select_options,
-    input_html: { class: 'form-control', multiple: true } %>
+    input_html: { class: 'form-control' } %>
```
cam156: @njaffer Has the rights changed from a multi select value to a single value? There is no indication in sufia 6.0 that this has changed: https://github.com/projecthydra/sufia/blob/6.x-stable/sufia-models/app/models/concerns/sufia/generic_file/metadata.rb#L41 

@mjgiarlo was this a decision that was made to move from multiple rights statements to a singular value?

mjgiarlo: Not that I can recall, @cam156. There was a ticket for this, but we deferred it: https://github.com/projecthydra/sufia/issues/1287

cam156:
@njaffer looks much better!  Thanks! :+1:   Anyone else want to chime in?

app/forms/sufia/forms/work_form.rb
```diff
@@ -7,5 +7,17 @@ def rendered_terms
                :visibility_after_embargo, :visibility_during_lease,
                :lease_expiration_date, :visibility_after_lease, :visibility, :thumbnail_id, :representative_id, :ordered_member_ids]
     end
+
+    # override Curation Concerns Work Form, Sufia allows multiple values for rights
+    def self.multiple?(term)
+      case term.to_s
+      when 'rights'
+        true
+      when 'ordered_member_ids'
```
jcoyne: Why is this required?

spec/form/work_form_spec.rb
```diff
@@ -0,0 +1,46 @@
+require 'spec_helper'
+
+describe Sufia::Forms::WorkForm do
+  before do
+    class PirateShip < ActiveFedora::Base
```
jcoyne: why do this? You can use GenericWork instead.

spec/form/work_form_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+      include CurationConcerns::RequiredMetadata
+      include CurationConcerns::BasicMetadata
+      include CurationConcerns::HasRepresentative
+    end
+
+    class PirateShipForm < described_class
```
jcoyne: we don't need this either. you can just use `Sufia::Forms::WorkForm`

app/forms/sufia/forms/work_form.rb
njaffer: @jcoyne: I am not sure I understand what are you referring to here, can be more specific

jcoyne: How does this clause solve the issue you reported?  It doesn't seem releated to me. If it's not related, it shouldn't go in this PR.

spec/form/work_form_spec.rb
```diff
@@ -0,0 +1,29 @@
+require 'spec_helper'
+
+describe Sufia::Forms::WorkForm do
```
jcoyne: This should be testing `Sufia::Forms::GenericWorkForm` instead.  `WorkForm` is an abstract class.

njaffer: ok so this is confusing, the form I am testing is work_form and when I use your suggestion, I get an  error: uninitialized constant Sufia::Forms::GenericWorkForm (NameError)

jcoyne: Sorry, I see it's actually called `CurationConcerns::GenericWorkForm`

njaffer: ok so in my understanding, I am overriding the multiple? in sufia's  work_form and I should be testing sufia work_form?

jcoyne: That's up to you. I don't think there's a hard rule for how to test abstract classes.

spec/form/work_form_spec.rb
jcoyne: Sorry, I meant to say `CurationConcerns::GenericWorkForm`

njaffer:
The PR can be closed. I tested the latest Sufia master branch, and the rights metadata field is getting saved now.

njaffer:
fixes #1645

mjgiarlo:
Thanks for the PR, @njaffer. I see there are some unresolved questions. I'll highlight those.

app/forms/sufia/forms/work_form.rb
mjgiarlo: @njaffer This discussion is unresolved.

spec/form/work_form_spec.rb
mjgiarlo: This one is unresolved too.

spec/form/work_form_spec.rb
mjgiarlo: This is unresolved.

spec/form/work_form_spec.rb
mjgiarlo: This is unresolved.

mjgiarlo:
@njaffer If it'd help to pair up on some of this, let me know. (Or if you'll be at LDCX/HDC next week, maybe that'd be a good place to take this up.)

app/forms/sufia/forms/work_form.rb
njaffer: @mjgiarlo: I think this was about the comment line I put previously, which is removed now, but I might be wrong.

njaffer:
@mjgiarlo: I have updated the spec test, still there is one unresolved discussion. I reopened the PR on Friday in reference to discussion with @hectorcorrea on #1645. Sorry I had to leave early on Friday, so couldn't cleanup the PR. I will be there at HDC, for sure if there is more discussion needed on this, we can take it up there.

app/forms/sufia/forms/work_form.rb
hectorcorrea: @jcoyne @mjgiarlo the reason this method override is required is because without it the `rights` field is being rejected by the controller. This is the method in CC that Nabeela is overriding: https://github.com/projecthydra-labs/curation_concerns/blob/master/app/forms/curation_concerns/forms/work_form.rb#L44

I suspect the fact that the field is defined as multi-value but the CC `multiple?` implementation marks it as single-value causes a mismatch that prevents the controller from allowing this field during the HTTP POST request.

@njaffer I think writing the override as follow might make it clearer that the intention is to handle a specific issue for the `rights` field:

```
def self.multiple?(term)
  return true if term == :rights
  super
end
```

mjgiarlo:
Great :thought_balloon:, @njaffer!

app/forms/sufia/forms/work_form.rb
njaffer: @hectorcorrea: thank you for the detailed note.. will update it!

PR #1607, opened 2016-02-25
Gemspec: cap Curation Concerns below 0.9.0

dunn:
Sufia is still using the `fits_to_desc_mapping` method which was removed in https://github.com/projecthydra-labs/curation_concerns/commit/a7a7c898883af146f234f9382ece502d6a1afca5.
Cap the version of CC to a compatible version until this can be fixed.

With CC 0.9.0, the following error occurs when starting `rails s`: https://gist.github.com/anonymous/1018c66fbaa126c12200

### DIFF
--- a/sufia.gemspec
+++ b/sufia.gemspec
@@ -16,7 +16,7 @@ Gem::Specification.new do |spec|
   spec.version       = version
   spec.license       = 'Apache2'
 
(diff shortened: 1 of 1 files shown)


dunn:
This isn't actually a Sufia problem; the issue is just that the initializer generated by CC 0.8.0 isn't compatible with 0.9.0.

Issue #1603, opened 2016-02-25
Error when updating a Work's visibility


hectorcorrea:
### Descriptive summary

In Sufia 7, when editing a work, changing its visibility value, and saving the updates, Sufia throws error

    NoMethodError at /concern/permissions/bk1289888/confirm
    undefined method `copy_curation_concern_permission_path' for ...

### Expected behavior
Not to throw an error.

### Actual behavior
Throws an error.

### Steps to reproduce the behavior

1. Edit a work with at least one file.
1. Change it's visibility (e.g. from public to private)
1. Save your changes
1. What Sufia crash.

### More info
The problem seems to be in view `sufia/app/views/curation_concerns/permissions/confirm.html.erb` line `Sufia::Engine.routes.url_helpers.copy_curation_concern_permission_path([sufia,curation_concern])`

hectorcorrea:
Fixed by PR #1615

Issue #1606, opened 2016-02-25
Confirm Permission view renders incorrectly


hectorcorrea:
### Descriptive summary
The `Confirm Permission` view renders raw HTML (see screenshot below), it still renders with the Curation Concernss styles rather than Sufia's. 

In addition the links to "Yes please" and "No, I'll update manually" are being rendered wrong. For example, the link for "No, I'll update manually" renders as `http://localhost:3000/concern/generic_works/%23%3CActionDispatch::Routing::RoutesProxy:0x007ff2e8063028%3E%2Fx633f101s`

### Expected behavior
The view should render without the raw HTML, using the Sufia styles, and the links should not include the ActionDispatch text.

### Actual behavior
See screenshot.

### Steps to reproduce the behavior

1. Edit a view with at least one file.
1. Change the work permissions (e.g. from Public to Private) 
1. Save your changes and you'll be presented with the "Confirm Permission" view.

### Related work
Wait until issue #1603 has been fixed.

![screen shot 2016-02-25 at 3 52 02 pm](https://cloud.githubusercontent.com/assets/568286/13334130/7a1f819e-dbd8-11e5-93d3-2df518725015.png)

ilessing:
There is erb code is [config/locales/sufia.yml](https://github.com/projecthydra/sufia/blob/master/config/locales/sufia.en.yml#L59) which shouldn't be there.

hectorcorrea:
@ilessing thanks for the pointer. I've replaced the ERB with tokens that can be injected during rendering which I believe was the original idea of whomever put the ERB on the YML file. See https://github.com/projecthydra/sufia/pull/1615/files

hectorcorrea:
Fixed by PR #1615

PR #1615, opened 2016-03-01
Fixes render error in confirm.html view.

hectorcorrea:
Fixes problem that caused this view to crash.

It also replaces ERB from YML file it with the correct syntax `%{token}` and passes the expected values to the `I18n.t()` call.

Fixes issues #1603 and #1606 

Below is how the view now renders
screen shot 2016-03-02 at 3 24 18 pm


hectorcorrea:
Any ideas why this view (`app/views/curation_concerns/permissions/confirm.html.erb`) is not using the Sufia look and feel and it is using the Curation Concern one?

PR #1621, opened 2016-03-02
Make explicit the expected style guide, and address outstanding issues

cbeer:
I've tried to adapt the existing rubocop rules to distinguish between the style guide we want and the styles we're (not) enforcing because of legacy code. I've also added new styles rules that match existing practice (in e43b3ee). And, finally, in 826c55a, I've addressed the 9 outstanding style violations. 


mjgiarlo:
:clap:

PR #1626, opened 2016-03-04
Removes hard coded root path on the Cancel button

### LINKED ISSUE #1602: Cancel button from edit work page takes you to the homepage
### Descriptive summary

Clicking "Cancel" from the edit work page takes you to the homepage.

### Expected behavior

Clicking "Cancel" from the edit work page should take you back to the work show page.

### Actual behavior

Clicking "Cancel" from the edit work page takes you to the homepage.

hectorcorrea:
... and uses Rails :back to make sure the user is sent to the previous page (e.g. dashboard or work show) regardless of how they got here.

Fixes #1602

### DIFF
--- a/app/views/curation_concerns/base/_form.html.erb
+++ b/app/views/curation_concerns/base/_form.html.erb
@@ -15,7 +15,7 @@
 
     <div class="primary-actions">
       <%= f.submit 'Save', class: 'btn btn-primary', onclick: "confirmation_needed = false;", id: "with_files_submit", name: "save_with_files" %>
-      <%= link_to t(:'helpers.action.cancel'), main_app.root_path, class: 'btn btn-link' %>
+      <%= link_to t(:'helpers.action.cancel'), :back, class: 'btn btn-link' %>
     </div>
 
   </div>
--- a/spec/views/curation_concerns/base/_form.html.erb_spec.rb
+++ b/spec/views/curation_concerns/base/_form.html.erb_spec.rb
@@ -1,30 +1,34 @@
 require 'spec_helper'
 
 describe 'curation_concerns/base/_form.html.erb', :no_clean do
+  let(:work) do
+    stub_model(GenericWork, id: '456')
+  end
+  let(:ability) { double }
+
+  let(:form) do
+    CurationConcerns::GenericWorkForm.new(work, ability)
+  end
+
+  before do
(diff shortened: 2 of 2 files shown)


mjgiarlo:
:ok:

awead:
I line or two in a feature test ought to do it.

hectorcorrea:
@mjgiarlo we don't have a test for it but if I add one it would need to execute the whole page workflow (e.g. render page when coming from the dashboard and render the page when coming from work show) to prove that the logic does work. A test like that seems an overkill to test that a built-in Rails feature indeed works.

awead:
bummer, I thought we already had an existing feature test that we could piggy-back on.

mjgiarlo:
@hectorcorrea I'm not really interested in making sure the built-in Rails feature works -- though given how quickly Rails moves and we keep pace with it, it's not imprudent to build some firewalls around Rails in our test suite -- but it'd be good to test the behavior is as we want it to be. Else someone can come in in three weeks and revert this change and the green test suite would suggest to them that no one cared about the change. Unless there's a less heavyweight/slow way to test this, I'm OK with skipping coverage in this PR. For instance, a view spec (vs. a feature spec)? Thoughts, @cam156 @jcoyne?

hectorcorrea:
@mjgiarlo I don't think a view test will really validate the behavior that we want (some times render `/` some other times render `/curation_concerns/works/123`). I think we would need a feature test to make sure that logic works and that's what I thought would be an overkill for a Cancel button.

mjgiarlo:
@hectorcorrea Yeah, I think you're right.

cam156:
@hectorcorrea could we piggy back off the test here: https://github.com/projecthydra/sufia/blob/master/spec/views/curation_concerns/base/_form.html.erb_spec.rb

Add another line to check the link content?  It has to be outputting something...

cam156:
@hectorcorrea I'm not really asking for you to test all the possible ways back works, just make sure for one scenario it is there.  Then we do not run into the situation where someone deletes the cancel button and we do not notice, assuming it is important to us to be there...

awead:
I wonder if mocking `request.env['HTTP_REFERER']` in a view would work... might not even be there though.

hectorcorrea:
Per the group's request I've added a view test.

mjgiarlo:
:clap: 

Thanks, @hectorcorrea!

PR #1630, opened 2016-03-04
Remove setup_mail initializer and replace it with documentation about mailers

### LINKED ISSUE #1588: Dead code? generator template for setup_mail.rb
Doesn't seem to be used anywhere. Remove?

https://github.com/projecthydra/sufia/blob/master/lib/generators/sufia/templates/config/setup_mail.rb

cbeer:
Fixes #1588


awead:
:laughing: Yeah, that shouldn't be there... :+1:

PR #1632, opened 2016-03-04
Add test coverage for RedisEventStore and resque approriately

### LINKED ISSUE #1629: Sufia::RedisEventStore should have a logger
### Descriptive summary

Due to some recent Redis changes, I saw:

``` ruby
Failure/Error: logger.error("unable to create event: #{e}") if logger
NameError:
  undefined local variable or method `logger' for Sufia::RedisEventStore:Class
```

Let's use `CurationConcerns::NullLogger.new` instead.

### Expected behavior

It should just return nil.

### Actual behavior

It doesn't.

awead:
Fixes #1629

`Sufia::RedisEventStore` doesn't have any test coverage, so we're adding some. Let's also resque from the appropriate error and always have some kind of logger.

### DIFF
--- a/lib/sufia/redis_event_store.rb
+++ b/lib/sufia/redis_event_store.rb
@@ -1,5 +1,29 @@
 module Sufia
   class RedisEventStore
+    class << self
+      def for(key)
+        new(key)
+      end
+
+      # @return [Fixnum] the id of the event
+      def create(action, timestamp)
+        event_id = Redis.current.incr("events:latest_id")
+        Redis.current.hmset("events:#{event_id}", "action", action, "timestamp", timestamp)
+        event_id
+      rescue Redis::CommandError => e
+        logger.error("unable to create event: #{e}")
+        nil
+      end
+
+      def logger
+        Rails.logger || CurationConcerns::NullLogger.new
+      end
+    end
+
+    def initialize(key)
+      @key = key
+    end
+
     def fetch(size)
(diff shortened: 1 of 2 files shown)


lib/sufia/redis_event_store.rb
```diff
@@ -1,5 +1,29 @@
...
+        event_id = Redis.current.incr("events:latest_id")
+        Redis.current.hmset("events:#{event_id}", "action", action, "timestamp", timestamp)
+        event_id
+      rescue Redis::CommandError => e
+        logger.error("unable to create event: #{e}")
+        nil
+      end
+
+      def logger
+        @logger = (defined?(Rails) && Rails.logger) || CurationConcerns::NullLogger.new
```
jcoyne: How could Rails ever be undef here?

awead: I didn't know the context in which this class was used. For example, if it was used in Resque, but I guess that has Rails as well? Either way, I'm :+1: to removing it.

jcoyne: Rails is a dependency of Sufia, so it must exist.

jcoyne: also why are you assigning to `@logger`?

awead: ah, yeah. habit. it's only called once so no need.

mjgiarlo:
:+1: as soon as the changes to `RedisEventStore` discussed by @jcoyne and @awead are pushed.

PR #1634, opened 2016-03-09
pass presenter to show_actions

njaffer:
Pass presenter to show actions partials

In View Files, the app crashes with the error: Argument error. In CC the partial is rendered with passing the presenter locally. Copying the same pattern here.

* pass presenter to show partial 

### DIFF
--- a/spec/views/curation_concerns/file_set/show.html.erb_spec.rb
+++ b/spec/views/curation_concerns/file_set/show.html.erb_spec.rb
@@ -1,0 +1,40 @@
+require 'spec_helper'
+
+describe 'curation_concerns/file_sets/show.html.erb', type: :view do
+  let(:parent) { stub_model(GenericWork) }
+  let(:user) { double(user_key: 'sarah', twitter_handle: 'test') }
+  let(:file_set) { build(:file_set, id: '123', depositor: user.user_key, title: ['My Title'], user: user, visibility: 'open') }
+  let(:ability) { double }
+  let(:solr_doc) { SolrDocument.new(file_set.to_solr) }
+  let(:presenter) { Sufia::FileSetPresenter.new(solr_doc, ability) }
+  let(:mock_metadata) do
+    {
+      format: ["Tape"],
+      long_term: ["x" * 255],
+      multi_term: ["1", "2", "3", "4", "5", "6", "7", "8"],
+      string_term: 'oops, I used a string instead of an array',
+      logged_audit_status: "Audits have not yet been run on this file"
+    }
+  end
+  before do
+    allow(view).to receive(:parent).and_return(parent)
+    view.lookup_context.prefixes.push 'curation_concerns/base'
+    allow(view).to receive(:can?).with(:edit, SolrDocument).and_return(false)
+    allow(view).to receive(:can?).with(:edit, String).and_return(true)
+    allow(presenter).to receive(:audit_status).and_return(mock_metadata)
+    assign(:presenter, presenter)
+    assign(:document, solr_doc)
+    assign(:audit_status, "none")
+  end
+
+  describe 'title heading' do
+    before do
+      stub_template 'shared/_title_bar.html.erb' => 'Title Bar'
+      stub_template 'shared/_citations.html.erb' => 'Citation'
+      render template: 'curation_concerns/file_sets/show.html.erb', layout: 'layouts/curation_concerns'
+    end
+    it 'shows the title' do
+      expect(rendered).to have_selector 'h1', text: 'My Title'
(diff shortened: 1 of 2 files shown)


spec/views/curation_concerns/file_set/show.html.erb_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+
+  let(:parent) { stub_model(GenericWork) }
+
+  let(:depositor) do
+    stub_model(User,
+               user_key: 'bob',
+               twitter_handle: 'bot4lib')
+  end
+
+  let(:content) do
```
mjgiarlo: I'm not sure this variable is being used. Delete it?

spec/views/curation_concerns/file_set/show.html.erb_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+                        title: ['My Title'],
+                        description: ['Lorem ipsum lorem ipsum. http://my.link.com']
+              )
+  end
+
+  let(:ability) { double }
+  let(:solr_doc) { SolrDocument.new(file_set.to_solr) }
+  let(:presenter) { CurationConcerns::FileSetPresenter.new(solr_doc, ability) }
+
+  before do
```
mjgiarlo: This `before` block should be combined with the one on lines 4-6, I suspect.

spec/views/curation_concerns/file_set/show.html.erb_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+    allow(view).to receive(:can?).with(:edit, String).and_return(true)
+    assign(:presenter, presenter)
+  end
+
+  describe 'title heading' do
+    before do
+      stub_template 'shared/_title_bar.html.erb' => 'Title Bar'
+      render template: 'curation_concerns/file_sets/show.html.erb', layout: 'layouts/curation_concerns'
+    end
+    it 'shows the title' do
```
mjgiarlo: Does this test fail if you revert your change to `app/views/curation_concerns/file_sets/show.html.erb`?

mjgiarlo:
I left a few minor comments but I'm otherwise :ok: with the changes made. Travis is :red_circle:, at the moment, though.

mjgiarlo:
@njaffer Does this close or relate to an existing issue or does this PR represent an as yet unticketed bug?

njaffer:
No it does not close or relate to any existing issue.

On Wed, Mar 9, 2016 at 12:34 PM, Michael J. Giarlo <notifications@github.com

mjgiarlo:
OK, great -- thanks for finding a new bug!

njaffer:
@mjgiarlo: I think it should be good to go!

mjgiarlo:
:+1: works for me! Thanks, @njaffer

Thoughts, @jcoyne @awead @cam156 @hectorcorrea? Shall we merge?

spec/views/curation_concerns/file_set/show.html.erb_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+    allow(view).to receive(:can?).with(:edit, SolrDocument).and_return(true)
+    allow(view).to receive(:can?).with(:edit, String).and_return(true)
+    assign(:presenter, presenter)
+    assign(:document, solr_doc)
+  end
+
+  let(:parent) { stub_model(GenericWork) }
+  let(:user) { double(user_key: 'sarah', twitter_handle: 'test') }
+  let(:file_set) do
+    create(FileSet, id: '123',
```
jcoyne: I'd prefer that we don't create models in view specs.  Can you just produce the values that go into solr without hitting Fedora?

njaffer:
@jcoyne: changed create to stub_model..

mjgiarlo:
Wasn't `stub_model` breaking the PR before?

njaffer:
@mjgiarlo: yes, I will need to stub more values so that it does not hit Fedora and still pass the test! Create is the easy way but I remember conversation with @jcoyne about not using create..

njaffer:
@mjgiarlo, @jcoyne: Here is the gist of error with using build instead of create: https://gist.github.com/njaffer/ab64b6fc7e90daea515c

I browsed through a lot of code where FileSet.find is called and all the places are using create for spec tests. Do we need to change all of those 'create' calls to 'build' or is it only the  views specs?

njaffer:
@mjgiarlo , @jcoyne: tests are passing without hitting Fedora!!

mjgiarlo:
:clap:

PR #1641, opened 2016-03-17
Add Sufia::FormBuilder

jcoyne:

Previously the CurationConcerns::FormBuilder was used, but Sufia was the
only consumer of that class.  This moves the class into Sufia so it may
be removed from CurationConcerns.

### DIFF
--- a/app/builders/sufia/form_builder.rb
+++ b/app/builders/sufia/form_builder.rb
@@ -1,0 +1,16 @@
+module Sufia
+  class FormBuilder < SimpleForm::FormBuilder
+    def input_label(attribute_name, options = {})
+      options = options.dup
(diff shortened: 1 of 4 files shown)


awead:
What's CurationConcerns using now? Could it be used in Sufia instead of FormBuilder?

jcoyne:
@awead curation_concerns doesn't call the `input_label` method. That's only on views in sufia.

awead:
@jcoyne ah, ok. So CC is using SimpleForm as-is without any enhancements?

jcoyne:
@awead it used to be in Sufia, introduced here: https://github.com/projecthydra/sufia/commit/79a8e2d1c2bea48fee718cfbfc17419320e3954b  then it moved to CC, now I'm just putting it back.

jcoyne:
@awead yes, CC does not have need of a FormBuilder that provides any additional methods.

PR #1642, opened 2016-03-17
Use CurationConcerns collection code, fixes #1620

### LINKED ISSUE #1620: Disentangle and realign CurationConcerns' use of HydraCollections with Sufia's
CurationConcerns uses HydraCollections, and so does Sufia. Sufia extends CC, but some of the code related to collections is duplicated in both gems. We should take a look at the duplication and remove it where necessary, and make a cleaner distinction between where they differ.

For example, there are many duplicated view partials, as well as methods in Sufia::CollectionPresenter. Additionally, test coverage might be duplicated as well and we could trim back some of the spec tests that overlap.

awead:
Fixes #1620 

Aligns Sufia's collections code more with CurationConcerns.

### DIFF
--- a/app/controllers/concerns/sufia/collections_controller_behavior.rb
+++ b/app/controllers/concerns/sufia/collections_controller_behavior.rb
@@ -1,7 +1,6 @@
 module Sufia
   module CollectionsControllerBehavior
     extend ActiveSupport::Concern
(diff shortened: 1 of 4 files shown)


mjgiarlo:
TRAVIIIIIIIIIIIS

PR #1647, opened 2016-03-18
Enforce default statefile location

### LINKED ISSUE #1532: The minter statefile default should work in dev and prod environments
Building off the very good discussion in #1510: in Sufia, we want dev/test environments to "just work" and prod environments not to break on reboot, or on Cap deployment, or on `/tmp` autoclean.

Let's keep using /tmp in dev, and require in our documentation that folks either use the /var/sufia/ location for this file (a _new_ default in prod envs) or override the value of `config.minter_statefile` themselves. This change should also be in the release notes of the release that contains this change (7.0.0).

We should support the documentation change above by changing the Sufia initializer to something like: `config.minter_statefile = Rails.env.production? ? '/var/sufia/minter-state' : '/tmp/minter-state'`. The initializer should also check that the path exists and, if it doesn't, complain **and fail** during initialization at runtime.
- [ ] Change the initializer to set the statefile default depending on environment
- [ ] Add logic to the initializer to complain and fail in production if `/var/sufia/minter-state` doesn't exist
- [ ] Test the above with new specs
- [ ] Update README
- [ ] Update/create release notes text

awead:
Fixes #1532 

Enforces the correct default location of a statefile, depending on the environment.

ping @mjgiarlo  thoughts?

### DIFF
--- a/lib/sufia/engine.rb
+++ b/lib/sufia/engine.rb
@@ -1,4 +1,12 @@
 module Sufia
+  class Statefile
+    def self.default
+      return '/tmp/minter-state' if Rails.env.development? || Rails.env.test?
+      raise NotImplementedError, "can't use the default statefile for production" unless Dir.exist?('/var/sufia')
+      '/var/sufia/minter-state'
+    end
+  end
+
   class Engine < ::Rails::Engine
     engine_name 'sufia'
 
@@ -79,7 +87,7 @@ module Sufia
     # Noid identifiers
     config.enable_noids = true
     config.noid_template = '.reeddeeddk'
-    config.minter_statefile = '/tmp/minter-state'
+    config.minter_statefile = Statefile.default
     config.translate_uri_to_id = ActiveFedora::Noid.config.translate_uri_to_id
     config.translate_id_to_uri = ActiveFedora::Noid.config.translate_id_to_uri
 
--- a/spec/lib/sufia/statefile_spec.rb
+++ b/spec/lib/sufia/statefile_spec.rb
@@ -1,0 +1,28 @@
+require 'spec_helper'
+
+describe Sufia::Statefile do
+  let(:default) { '/tmp/minter-state' }
+  let(:production_default) { '/var/sufia/minter-state' }
+
+  subject { described_class.default }
+
+  it { is_expected.to eq(default) }
+
+  context "when in development" do
+    before { allow(Rails).to receive(:env).and_return(ActiveSupport::StringInquirer.new("development")) }
+    it { is_expected.to eq(default) }
+  end
+
+  context "when in production" do
+    before { allow(Rails).to receive(:env).and_return(ActiveSupport::StringInquirer.new("production")) }
+    context "with a missing default directory" do
+      it "raises an error" do
+        expect { subject }.to raise_error(NotImplementedError)
+      end
+    end
+    context "with a default directory" do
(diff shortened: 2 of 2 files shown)


escowles:
:+1: to moving the statefile out of `/tmp` for non-dev/test environments.  Maybe this could go in CC though?

jcoyne:
Yes, please move this to CC.

lib/sufia/engine.rb
```diff
@@ -1,4 +1,12 @@
 module Sufia
+  class Statefile
+    def self.default
+      return '/tmp/minter-state' if Rails.env.development? || Rails.env.test?
+      raise NotImplementedError, "can't use the default statefile for production" unless Dir.exist?('/var/sufia')
```
jcoyne: What if it is in production and I don't have `/var/sufia`, but I've specified an alternative location (lets say `/opt/sufia`).  Won't this raise an error?

awead: It shouldn't because if you'd specify that in config/initializers/sufia.rb, and it would skip this entirely. Of course, it wouldn't check to see if `/opt/sufia` was there, but that's not our problem, is it?

jcoyne: Are you certain about that? I thought it always ran the code in engine.rb, then ran your configuration over the top of that.

awead: Oh I see what you're saying... even if I have specified my own location, it would still execute `Sufia::Statefile.default`? I wonder if it can check to see if a value has been provided.

awead: @mjgiarlo see @jcoyne above. He's correct. Since the engine code gets executed first, Sufia.config doesn't exist yet, so we don't really know what the user has specified for their statefile. The only solution I can see right now is perhaps raising an error if `/tmp/minter-state` exists while running in production, or just returning `nil` as the default in production which would force the user to set it to something. Or both?

jcoyne: Yeah, I think just set `/var/sufia` as the default location. If they don't override it, they'll get an error when it can't be created.  We could also add a method to the initializer that validates the location exists and raise an exception at that point.

mjgiarlo:
Reading this and #1532 over now.

mjgiarlo:
I might recommend:

- [ ] fixing this in Sufia first
- [ ] doing a full extraction to CC next
- [ ] removing duplicate code from Sufia last

I'm not sure that *all* the Sufia minter stuff is in CC yet. (I'm looking at output of `ag minter_statefile curation_concerns/ sufia/`.)

lib/sufia/engine.rb
mjgiarlo: I believe the solution in the ticket was to make this change in the Sufia initializer (within the application) vs. in the Sufia engine:

https://github.com/projecthydra/sufia/issues/1532#issue-128606782

And we'd either need a generator (to run on upgrade) to automatically "fix" folks initializers, or skip the generator and include instructions in the release notes. (This latter part is already covered in #1532.)

mjgiarlo:
:-1: on the current implementation, but :clap: for taking this on!

I believe per #1532 that we want this change in the generated Sufia initializer not in the engine itself.

awead:
@mjgiarlo I'm un-assigning myself from this one.  I don't think I fully understand the requirements.

mjgiarlo:
Thanks for taking the time to put in this PR, @awead! If you're still interested in this work, I'd be happy to chat about this over Hangout or pair up or something. (Next week, after LDCX/Hydra Power Steering.)

PR #1661, opened 2016-03-25
Sends notification when proxy depositor specified

### LINKED ISSUE #1655: Proxy user designee notification.
### Descriptive summary
Proxy user designee does not get notified after they have been designated as a proxy by another user.
Sufia 7.x HEAD

### Expected behavior
User designated as a proxy should see a message in user notifications informing them they've been designated the proxy for some other user.

### Actual behavior
No notifications for designated user.

1. Create two users: first and other.
2. Log in as the first user.
3. Go to dashboard.
4. Designate other user as a proxy under "Manage Proxys" section.
5. Log out.
6. Log in as the other user.
7. Go to dashboard.

### Related work

Tangentially related to #1609

afred:
Fixes #1655 

Sends notification when proxy depositor specified

### DIFF
--- a/app/controllers/concerns/sufia/depositors_controller_behavior.rb
+++ b/app/controllers/concerns/sufia/depositors_controller_behavior.rb
@@ -24,6 +24,7 @@ module Sufia
         head :ok
       else
         grantor.can_receive_deposits_from << grantee
+        send_proxy_depositor_added_messages(grantor, grantee)
         render json: { name: grantee.name, delete_path: sufia.user_depositor_path(grantor.user_key, grantee.user_key) }
       end
     end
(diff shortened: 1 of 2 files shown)


njaffer:
@afred: Is it sending email or a notification message? I thought it was only the notification message, in that case the spec test description would not be 'fires off an email to both grantor and grantee'

afred:
@njaffer :+1: thanks.. I'll fix it.

app/controllers/concerns/sufia/depositors_controller_behavior.rb
```diff
@@ -24,6 +24,7 @@ def create
         head :ok
       else
         grantor.can_receive_deposits_from << grantee
+        send_proxy_depositor_added_messages(grantor, grantee)
```
jcoyne: Can you make this a subclass of EventJob and call it asynchronously (`perform_later`)? For example: https://github.com/projecthydra/sufia/blob/master/app/controllers/concerns/sufia/users_controller_behavior.rb#L91

mjgiarlo:
Thank you for your work, @afred!

app/controllers/concerns/sufia/depositors_controller_behavior.rb
afred: got merged before i read this. New ticket?

mjgiarlo: Nm, I'd say. I don't think we want to conflate user messaging with event jobs. Thanks for the PR!

PR #1664, opened 2016-03-27
Log an error when the terms component is not enabled

jcoyne:


app/models/system_stats.rb
```diff
@@ -97,11 +97,12 @@ def top_data(key, limit)
       query_url = "#{ActiveFedora.solr_config[:url]}/terms?terms.fl=#{key}&terms.sort=count&terms.limit=#{limit}&wt=json&omitHeader=true"
       # Parse JSON response (looks like {"terms":{"depositor_tesim":["mjg36",3]}} for depositor)
       json = open(query_url).read
-      begin
-        tuples = JSON.parse(json)['terms'][key]
-      rescue
-        []
+      unless json
```
mjgiarlo: I don't love `unless` as a block but `¯\_(ツ)_/¯`?

jcoyne: "Favor `unless` over `if` for negative conditions " https://github.com/bbatsov/ruby-style-guide#unless-for-negatives

mjgiarlo: Yeah, I wasn't saying it was a style violation. Not a problem.

Issue #1646, opened 2016-03-18
Disambiguate depositor and owner


grosscol:
### Descriptive summary

In the case of Works the depositor of the work is conflated to also mean owner.  The owner property is not currently being used.  This is relevant to proxy deposits and transfers of works where the original depositor is a different user than the eventual owner.

This affects Sufia 7.x
### Expected behavior

Using owner property to track the current owner of the work.
Transfer Work to user changes the owner, not the depositor.
### Actual behavior

Owner property is empty.
Transfer Work changes the depositor.
### Steps to reproduce the behavior

Will put in subsequent comments when I get the opportunity.

mjgiarlo:
Is this a bug or do you consider this surprising behavior? (I'm not sure what the owner property is or where it comes from.)

grosscol:
I would not say this is a requirement for a 7.0.  The property comes from CC-models work behavior: curation_concerns-models/app/models/concerns/curation_concerns/work_behavior.rb 
`property :owner, predicate: RDF::URI.new('http://opaquenamespace.org/ns/hydra/owner'), multiple: false`

This arose because it is problematic for tracking who did the action of depositing or uploading something and who currently owns it.  Work transfers and proxy deposits illuminated this.  The user that agrees to the depositor agreement does the uploading action.  That may be different from the owner of the work if it's a proxy deposit or was transferred to another user.

mjgiarlo:
This issue is deeper than Sufia -- there is no strong "owner" concept in underlying gems (AF, HH). I suggest this requires more discussion before we implement up at the Sufia/CC/Hyrax layer. Thanks for reporting, @grosscol, and let me know if I can help this conversation happen. Closing for now.

PR #1666, opened 2016-03-28
Display owner if work is transferred

njaffer:
refs #1646 

When a work is transferred from User A to User B, the depositor is displayed as User B. The proposed fix at DevCon was to display the User A as depositor and User B as Owner and User A as depositor.

screen shot 2016-03-28 at 11 54 38 am
 

* If proxy depositor is present on a work, display both owner and depositor
* 
* 


mjgiarlo:
@njaffer The build is :red_circle: but I dig the change. I think we should *always* display the owner, even if the value is identical to the depositor. Can we add that to this PR?

Thoughts, @projecthydra/sufia-code-reviewers ?

PR #1682, opened 2016-03-29
Remove unnecessary intermediate container-fluid element

cbeer:
Before (note the left and right margins):
screen shot 2016-03-29 at 4 37 38 pm

After:
screen shot 2016-03-29 at 4 37 17 pm

PR #1683, opened 2016-03-29
Remove unnessary page positioner

cbeer:
(Includes #1682; once the left and right gutters are gone, the gradient it provides is not visible.)


mjgiarlo:
Tagging @mtribone who added the page positioner so he can comment. (It was added in 2014 and may not be needed anymore.)

mtribone:
@cbeer @mjgiarlo It was a remnant of our transition from ScholarSphere 1.0 to 2.0. It isn't necessary. :+1:

Issue #1611, opened 2016-02-29
Restructure CSS


justcolin:
## Summary

While working on U-M's beta interface for our repository based on Sufia 7 we had a lot of trouble working with Sufia's current stylesheets. The following suggestions are centered around simplifying maintenance and making it easier for groups to customize Sufia's views as much or as little as they see fit.

**Note:** This issue should probably be dealt with as part of implementing whatever the UI working group comes up with instead of working on the current interface.

## Suggestions

1. Reduce the amount of CSS in Sufia to a bare minimum, instead relying on native Bootstrap wherever possible.
  - Bootstrap is a great tool that was made to be easier to customize. If we let Bootstrap do almost all the heavy lifting and trust it to have proper padding/margins it will be a lot easier for developers to customize Sufia to meet their needs
  - This is also reduce the amount of CSS which has to be maintained by the Sufia community.
2. Any custom CSS shouldn't have any id's in the selectors, and should only have at most two classes.
  - This will make it easier to overwrite styles and will ease longterm maintenance.
3. Add semantically named classes to most elements in Sufia, even when they aren't being used by stock Sufia's CSS.
  - This will make it easier for people to restyle Sufia without having the go in and edit the HTML.
4. Improve the accessibility of Sufia's stock styles.
  - There are a number of things that need improvement so that it is easier for more people to use Sufia, such as more obvious link styles, clear `:hover` and `:focus` actions, etc.

## Related
See #1594 for the start of this discussion
See #1608 for the discussion of front-end development best practices

justcolin:
@grosscol

jcoyne:
I think I'm pretty strongly against point one. I think that people want a repository that looks good out of the box.  Not every institution is Michigan, and has the resources to completely redo the CSS.

jcoyne:
I think point 4 got messed up (it's trailing point 3) 

> There are a number of things that need improvement

Please expand on these things.  Let's make all suggestions actionable.

grosscol:
I don't get the impression that many institutions that are not already in the mix will be doing a great deal of work in house.  I imagine that the hydra is a box project will provide a very stylish styling.

For the base Sufia, making easier to modify css should give us a more stable (fewer places to touch) base for UI work.

I do concur with a list of actionable items.

jcoyne:
@grosscol But what about the institutions that *are* already in the mix. Those are the ones I care most about right now.

grosscol:
[...]

PR #1690, opened 2016-03-30
Remove tagcloud widget from the home page

cbeer:


mtribone:
:clap:

jcoyne:
Is there a rationale for this?  I'm not arguing against this proposal, but I wonder if concerned parties agree with removing it.

cbeer:
@mjgiarlo enthusiastically suggested throwing it away.

mtribone:
We *may* need to keep it at Penn State, but might just be able to add it at the ScholarSphere level. However, it doesn't get a lot of hits when I take a quick look at analytics.

jcoyne:
It seems like it could be in a plugin if more than one person wants it.

mtribone:
Feb 29 through March 29 nobody clicked on the tag cloud in ScholarSphere...

mjgiarlo:
Refs #1611

PR #1691, opened 2016-03-30
Custom button styling should use bootstrap variables

cbeer:
I couldn't track down any consumers of the `.info` styling, introduced in 721ef7088100915ef2ec87b1aa08bdd7de2fc07b. Maybe it should have been `.btn-info`, but if no one has noticed in Sufia 6.x, I don't see why we'd introduce custom styling now.


mjgiarlo:
Refs #1611

PR #1712, opened 2016-03-30
Do not automatically install PhantomJS

mjgiarlo:
I'm not sure we need to, and the auto-install process is failing builds here and there. The problem: intermittent outages and rate limiting at BitBucket (where PhantomJS is hosted) cause the download and thus the build to fail.


mjgiarlo:
I'm going to restart the Travis builds here 3-4 times in a row to see if the failures go away. Please do not merge until I give the go sign.

mjgiarlo:
Workaround failed. Here's another way to deal with this: https://github.com/travis-ci/travis-ci/issues/3225#issuecomment-200965782

PR #1725, opened 2016-03-31
Use HTML5 date field instead of plain text elements for date values

cbeer:
screen shot 2016-03-31 at 11 05 25 am

### DIFF
--- a/app/views/curation_concerns/base/_form_progress.html.erb
+++ b/app/views/curation_concerns/base/_form_progress.html.erb
@@ -32,7 +32,7 @@
                 <div class="form-inline">
                   Embargo
                   <%= f.input :visibility_during_embargo, wrapper: :inline, collection: visibility_options(:restrict), include_blank: false %>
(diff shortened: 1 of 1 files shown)


app/views/curation_concerns/base/_form_progress.html.erb
```diff
@@ -32,8 +32,7 @@
                 <div class="form-inline">
                   Embargo
                   <%= f.input :visibility_during_embargo, wrapper: :inline, collection: visibility_options(:restrict), include_blank: false %>
-                  <%= f.input :embargo_release_date, wrapper: :inline, input_html: { value: f.object.embargo_release_date || Date.tomorrow, class: 'datepicker' } %>
-                  <%= f.input :visibility_after_embargo, wrapper: :inline, collection: visibility_options(:loosen), include_blank: false %>
```
mjgiarlo: Do we not need this line anymore?

cbeer: Looks like I lost a line break.

PR #1728, opened 2016-03-31
Use consistent user profile panel on dashboard and user page

cbeer:
With avatar image:

screen shot 2016-03-31 at 11 46 26 am

screen shot 2016-03-31 at 12 46 51 pm


mjgiarlo:
Looks much better, @cbeer! My only question is whether we should consistently align the number badges, i.e., shift views, downloads, followers, and following all the way over as well.

cbeer:
I think it might be more confusing if all the numbers were aligned right -- at the very least, it'd be harder to tell which number goes with what. I could be persuaded pretty easily about aligning followers/following, though.

mjgiarlo:
Works for me, @cbeer. Will merge when :green_heart:

Issue #1583, opened 2016-02-17
Toolbar


dt-ucsd:
# Wireframes
- [Toolbar with sub-menu items revealed](https://wiki.duraspace.org/pages/viewpage.action?pageId=73173754)
- [Placement on a page](https://wiki.duraspace.org/pages/viewpage.action?pageId=73173840)

# Demo
- [Bootstrap 3 w/ FA Icons (Responsive)](https://library.ucsd.edu/assets/temp/dt/sufia-toolbar.html)
  - [Gist](https://gist.github.com/dt-ucsd/d09baaa2c6626ef23b17)

# Usage
Easy access to site tools for logged in users.

# Implementation

## Visibility
- The toolbar is only visible to logged in users
- The "Admin" menu is only visible to logged in "Admin" users

## Placement
- The toolbar is always positioned at the top of the page (Upper left-hand corner at 0,0)
- The toolbar spans the entire width of the page (100%)

## Menu Items
- Top level menu items open their respective sub-menu
- Sub-menu items link to their respective Sufia content

## Mobile
- Use BS3 mobile breakpoints
- Top level menu items only display FA icon in mobile view

## Decorations
- The toolbar uses the following Font Awesome icons:
 - fa-gear (Admin)
 - fa-tachometer (Dashboard)
 - fa-cube (Works)
 - fa-cubes (Collections)
 - fa-user (User)
- The label for the "User" menu is the username of the logged in user

cbeer:
Should we consider pushing the Sufia branding into the toolbar as well, matching the default bootstrap placement?

dt-ucsd:
@cbeer -- This is an interesting wrinkle. Is the intention to merge the Masthead and Toolbar into a single unit?

When we designed the Toolbar, it was with the custom Masthead in mind. It was meant to be separate.

Also, to more directly answer your question, I suppose we could rename the "Admin" menu to "Sufia" and use the little globe icon there in place of the gear.

dt-ucsd:
Oh yeah, I also wanted to mention that Lynette and I started work on adding this feature in last week at the Dev Congress. You can find our work here: https://github.com/projecthydra/sufia/tree/FIX_1583-add_toolbar

mjgiarlo:
@elrayle @dt-ucsd may I delete [this branch](https://github.com/projecthydra/sufia/tree/FIX_1583-add_toolbar) now that this issue is closed?

dt-ucsd:
It’s okay with me.

[...]

PR #1732, opened 2016-03-31
Use default bootstrap navbar styling instead of custom masthead

cbeer:
First steps toward #1583.

screen shot 2016-03-31 at 2 44 46 pm

screen shot 2016-03-31 at 2 44 41 pm

### DIFF
--- a/app/assets/stylesheets/sufia/_header.scss
+++ b/app/assets/stylesheets/sufia/_header.scss
@@ -1,49 +1,10 @@
-#masthead {
-  background: $header-bg;
-  min-height: 4em;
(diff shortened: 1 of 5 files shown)


cbeer:
Note that the buttons are vertically centered, just a bad screenshot.

mtribone:
Very nice. I think that we can handle something like making the logo bigger, if needed, at an institutional level.

mjgiarlo:
@mtribone Indeed, and we should document how to do this in a way that minimizes complicating the upgrade path. :clap:

Issue #1649, opened 2016-03-22
Style errors on FileSet edit page.


jcoyne:
> the permission form's lease and embargo selectors are overrunning the text

Moved this issue from  #1604

atz:
Example:
![screen shot 2016-03-29 at 12 00 47 pm](https://cloud.githubusercontent.com/assets/109954/14120073/2219d9d0-f5a6-11e5-93a9-b117902880ac.png)

atz:
One of these overlaps comes from http://localhost:3000/assets/curation_concerns.self-4e4167184078d8e858eee542cdad2883700618fc5d8b5e68b33e416fd0f926f9.css?body=1 (line 9477)
```css
access-controls .form-inline label {
    padding-left: 0;
    display: inline;
}
```
Removing the CC `padding-left` reinstates the Sufia padding.   Due to #1680, perhaps the answer is to stop including CurationConcern's CSS?

atz:
Upstream updates are of the _one step forward, two steps back_ variety.

![screen shot 2016-03-31 at 6 33 40 pm](https://cloud.githubusercontent.com/assets/109954/14195539/3662399e-f76f-11e5-9f53-9388a8b70744.png)

atz:
Also, what the heck is "**Your File's Title**" doing in there above "Title" and... the actual editable title?

mjgiarlo:
@atz I haven't seen "Your File's Title" before, but it does look out of place.

PR #1738, opened 2016-04-01
Remove extraneous legend element

atz:
Relates to #1649 

Legends are *potentially* useful for accessibility.  They are not *automatically* beneficial.  This one is **detrimental** to accessibility, reiterating the one label of the one input field in the form.  Employing a screenreader does not make that helpful.

Tons more work to do here, but cutting cruft is a start.

### DIFF
--- a/app/views/curation_concerns/file_sets/_form.html.erb
+++ b/app/views/curation_concerns/file_sets/_form.html.erb
@@ -1,6 +1,5 @@
 <%= simple_form_for [main_app, curation_concern], html: { multipart: true } do |f| %>
   <fieldset class="required">
-    <legend>Your File&#8217;s Title</legend>
     <span class="control-label">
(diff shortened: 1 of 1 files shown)


mjgiarlo:
I believe this came over from worthwhile. I'm :+1: on removing it. Do you have thoughts, @jcoyne @mtribone?

jcoyne:
I don't care

mtribone:
I would say that it doesn't need a fieldset either if we're dealing with only one label for one input.

WebAIM says "Fieldset and legend should only be used to associate groups of controls when a higher level description (i.e., the legend) is necessary. Single checkboxes or basic radio buttons (such as male/female for gender) that make sense from their labels alone do not require fieldset and legend. "

http://webaim.org/techniques/forms/controls

PR #1747, opened 2016-04-04
Update checkbox when files are uploaded.

### LINKED ISSUE #1684: Create new works: Sidebar status
- [x] It should update the metadata complete after all required metadata is entered
- [x] It should update the files complete after at least one file has been attached.

Check that this works both for creating and editing.

### LINKED ISSUE #1688: Create new works: disable save until valid
Don't allow the user to save until the required metadata is entered and at least one file has been attached.

### LINKED ISSUE #1714: Active acceptance of deposit licence
https://wiki.duraspace.org/display/hydra/New+Work+-+Metadata+Tab+-+Phase+1+-+UI+Wireframes#NewWork-MetadataTab-Phase1-UIWireframes-VariantsofSaveWorkwidget

Requires new configurations for `config.active_deposit_agreement_acceptance`
  * if true, show checkbox and active acceptance message for Deposit Agreement (see variant); Also show in Requirements area and set to red exclamation if acceptance is not checked or green check if acceptance is checked. Needs to be able to be changed if user checks/unchecks acceptance.
  * if false, do not show checkbox and do show passive acceptance message for Deposit Agreement (see in main New Work wireframe)

Ref #1562

jcoyne:
* Use AMD modules to impose some namespacing on the Javascript.
* Sidebar shows when files are uploaded.
* Disable save button until files and required metadata are added.
* Active acceptance of deposit license.

Fixes #1684 
Fixes #1688 
Fixes #1714


lib/generators/sufia/templates/config/sufia.rb
```diff
@@ -120,7 +120,14 @@
   # NOTE: if you have always sent analytics to GA for downloads and page views leave this commented out
   # config.analytic_start_date = DateTime.new(2014,9,10)
 
-  config.geonames_username = ''
+  # Location autocomplete uses geonames to search for named regions.
+  # Specify the user for connecting to geonames:
+  # config.geonames_username = ''
+
+  # Should the acceptance of the licence agreement be active (checkbox), or
```
mjgiarlo: We'll also need a generator that knows how to add this on upgrade. Can you ticket that? Otherwise :+1: to this PR!

jcoyne: You don't need to add this on upgrade. It's already defaulted to the old behavior.

mjgiarlo: Folks who are upgrading should at least know this is now possible with 7.0.0, though, if they want to turn off active acceptance. We should at least include this in the release notes (or ticket doing same), and then if someone wants to add this to the upgrade generator, great, but not necessary.

mjgiarlo: Done: #1751.

PR #1760, opened 2016-04-05
We already track this information in GitHub releases

mjgiarlo:
And much of the history was just copied from the changelog, so this seems worthless now.


cbeer:
:clap:

PR #1765, opened 2016-04-05
Cleanup BlacklightOverride

jcoyne:

* Stage methods that we can remove with the next version of CC
* Move render_bookmarks_control? to the controller context.
* Remove implied hash braces


mjgiarlo:
@jcoyne :red_circle:

Issue #1562, opened 2016-02-08
UI Create _save_work_widget partial


elrayle:
### Wireframes:
* [in context of New Work](https://wiki.duraspace.org/display/hydra/New+Work+-+Metadata+Tab+-+Phase+1+-+UI+Wireframes)
* [in context of Edit Work](https://wiki.duraspace.org/display/hydra/Edit+Work+-+Metadata+Tab+-+Phase+1+-+UI+Wireframes)
* [in context of Batch Upload](https://wiki.duraspace.org/display/hydra/Batch+New+Works+-+Phase+1+-+UI+Wireframes#mockup_p1-works-batch_new-step1_upload_files)
* [variants](https://wiki.duraspace.org/display/hydra/New+Work+-+Phase+1+-+UI+Wireframes#NewWork-Phase1-UIWireframes-VariantsofSaveWorkwidget) - initial state, error state, success state, embargo options, lease options, active acceptance of deposit agreement

### Usage - Right Hand Side widget to control saving of new/changed works
* Add New Work -- menu:  top-bar-> Works -> New
* Edit Work -- menu: My Dashboard -> work's Action select-menu -> Edit
* Batch Upload -- menu: top-bar -> Works -> Batch New

NOTE: Primary differences between usage on these screens.
* if new_work or batch_upload, then do not show 'Last saved on' under widget; shown for editing only
* if new_work or edit_work, title=Save Work
* if batch_upload, title=Save Works

### Implementation:
* File: app/views/generic_works/_save_work_widget.html.erb
* create a partial view to display the Save Work widget

### Actions:
* Save button - Hook into the existing save process and save all information across all tabs.  Move files uploaded prior to save from tmp area to permanent location.  Finish uploading any files that did not complete prior to the save.  After save, go to the Show page for the new work.
 * Does work already exist?  
   * If yes, update existing work.
    * If no, create work.
 * Were files uploaded to tmp area? If yes, move to permanent location for work.
 * Have all files completed the upload process? Wait for files already partially uploaded to complete loading and move to permanent location of work.  Upload remaining files directly to work.  (Can this be done in background?)
 * Hook into the existing save process and save all information across all tabs.
 * After save, go to Show page for the new/changed work.

* Cancel button
 * Do not create or save new/changed work.  If new work was created as start of create work process, then remove it.
 * Remove any files uploaded to the tmp area.

elrayle:
NOTES:
[...]

PR #1778, opened 2016-04-06
My CSS-only interpretation of Bootstrap Affix for save/visibility

### LINKED ISSUE #1719: Affix the create a new work status panel to the screen
So that it doesn't move when you scroll.

Consider: 
- [ ] mobile screen sizes
- [ ] viewport is shorter than the status panel.

Ref #1562

mtribone:
Fixes #1719  ; refs #1562 

CSS-only interpretation of Bootstrap 3's affix.js to keep the save/visibility widget visible when scrolling vertically through the metadata. Has a tendency to cannonball into the footer when the *height* of the browser is short and the *width* hasn't hit the mobile media query.

*  added affix class to <article>
*  media queries to handle viewport width
*  CSS positioning

### DIFF
--- a/app/assets/stylesheets/sufia/_form-progress.scss
+++ b/app/assets/stylesheets/sufia/_form-progress.scss
@@ -1,3 +1,25 @@
+@media screen and (min-width: 400px) {
+  .affix {
+    position: static;
+    margin-top: 1em;
+  }
+}
+@media screen and (min-width: 768px) {
+  aside.form-progress {
+    margin-right: 0.5em;
+  }
+  .affix {
+    position: fixed;
+  }
+  .tab-content {
+    min-height: 55em;
+  }
+}
+@media screen and (min-width: 960px) {
+  aside.form-progress {
+    margin-right: 1.5em;
+  }
+}
 aside.form-progress {
   ul.requirements {
     list-style-type: none;
--- a/app/views/curation_concerns/base/_form_progress.html.erb
+++ b/app/views/curation_concerns/base/_form_progress.html.erb
@@ -1,4 +1,4 @@
-    <aside class="form-progress panel panel-default" id="form-progress">
+    <aside class="form-progress panel panel-default affix" id="form-progress">
       <div class="panel-heading">
         <h2>Save Work</h2>
       </div>


mjgiarlo:
Tagging @aeschylus so he can have a look

mjgiarlo:
Also: @cbeer

cbeer:
Thanks, I'm very interested in taking a look.

mjgiarlo:
Have at it, @cbeer. This PR isn't in my wheelhouse, but it looks :+1: to me.

cbeer:
@mtribone do you have an example using the bootstrap affix too? I'm curious what trade-offs we're making.

My main concern is making sure it's worth maintaining this code here in Sufia, or if it's worth extra effort or some PR/ticket filing in bootstrap to make it work for us.

app/assets/stylesheets/sufia/_form-progress.scss
```diff
@@ -1,3 +1,25 @@
+@media screen and (min-width: 400px) {
```
cbeer: Can we use bootstrap variables for the screen sizes?

app/assets/stylesheets/sufia/_form-progress.scss
```diff
@@ -1,3 +1,25 @@
...
+}
+@media screen and (min-width: 768px) {
+  aside.form-progress {
+    margin-right: 0.5em;
+  }
+  .affix {
+    position: fixed;
+  }
+  .tab-content {
+    min-height: 55em;
```
cbeer: What's the significance of `55em`? Should we extract it to a variable?

app/assets/stylesheets/sufia/_form-progress.scss
mtribone: Absolutely.

app/assets/stylesheets/sufia/_form-progress.scss
mtribone: Since I'm using CSS positioning, it pulls the save/visibility box out of the normal document flow. The minimum height of 55em ensures that the widget without the additional metadata fields doesn't dive behind the footer.

mtribone:
@cbeer Here's the Bootstrap version.

https://github.com/projecthydra/sufia/tree/upload_files_affix

I still think that both attempts require more work and trouble shooting. It's a bit of a tricky thing with the extra hidden metadata and the variable height of the widget based on which radio button one selects.

mjgiarlo:
Let's see if we can come to a decision on this after standup today.

mjgiarlo:
Concluded today: @mtribone will see if a polyfill gets us the desired sidebar behavior. If not, he'll bring this to the Bootstrap community for feedback to see if we're doing it wrong or if there may be a bug that needs fixing.

PR #1799, opened 2016-04-07
Exposes LibreOffice configuration path to Hydra::Derivatives

eefahy:


mjgiarlo:
do not merge!

@eefahy is in the process of submitting an iCLA. Whenl I confirm that it's in place, we can merge this PR. In the meantime, feel free to review the work here. :clap:

lib/sufia/engine.rb
```diff
@@ -72,6 +73,7 @@ class Engine < ::Rails::Engine
     config.temp_file_base = nil
     config.redis_namespace = "sufia"
     config.fits_path = "fits.sh"
+    config.libreoffice_path = "/opt/local/libreoffice_path/bin/soffice"
```
jcoyne: default should probably be `soffice` which requires it to be on your path. This will work for the largest number of systems.

jcoyne: `soffice` is the present default: https://github.com/projecthydra/hydra-derivatives/blob/252cabb1278f703d20805d813bdc5c23ca056a8c/lib/hydra/derivatives/config.rb#L13

eefahy: changed in both the default and the comment in `sufia.rb`

PR #1804, opened 2016-04-07
Adds dropdown controls to work edit form to allow user to select a representative media and a thumbnail for a work

### LINKED ISSUE #1601: Depositors should be able to select representative media and thumbnail for a Work
**Please note this ticket has been changed from a question to an enhancement. See the acceptance criteria: https://github.com/projecthydra/sufia/issues/1601#issuecomment-199954403**

If you upload multiple files as part of a work, one of them is chosen to be the representative file of the work. This seems to be the first FileSet object in the work. How does one go about changing that? It seems like it'd be a process of re-ordering the contained FileSets, but I don't see where that's done anywhere.

hectorcorrea:
Fixes #1601

screen shot 2016-04-07 at 3 42 44 pm

### DIFF
--- a/app/views/curation_concerns/base/_form_metadata.html.erb
+++ b/app/views/curation_concerns/base/_form_metadata.html.erb
@@ -18,6 +18,7 @@
                     'aria-expanded'=> "false",
                     'aria-controls'=> "extended-terms" %>
         <div id="extended-terms" class='collapse'>
+          <%= render 'form_media', f: f %>
           <% (f.object.rendered_terms - [:title, :creator, :tag, :rights]).each do |term| %>
             <%= render_edit_field_partial(term, f: f) %>
           <% end %>
--- a/spec/views/curation_concerns/base/_form.html.erb_spec.rb
+++ b/spec/views/curation_concerns/base/_form.html.erb_spec.rb
@@ -12,6 +12,7 @@ describe 'curation_concerns/base/_form.html.erb' do
 
   before do
     view.lookup_context.view_paths.push 'app/views/curation_concerns'
+    allow(work).to receive(:member_ids).and_return([1, 2])
     allow(view).to receive(:curation_concern).and_return(work)
     allow(controller).to receive(:current_user).and_return(stub_model(User))
     assign(:form, form)
@@ -41,6 +42,12 @@ describe 'curation_concerns/base/_form.html.erb' do
     describe 'when the work has two or more resource types' do
       it "only draws one resource_type multiselect" do
         expect(page).to have_selector("select#generic_work_resource_type", count: 1)
+      end
+      it "allows to change the thumbnail" do
+        expect(page).to have_selector("select#generic_work_thumbnail_id", count: 1)
+      end
+      it "allows to change the representative media" do
+        expect(page).to have_selector("select#generic_work_representative_id", count: 1)
       end
     end
 
--- a/spec/views/sufia/batch_uploads/_form.html.erb_spec.rb
+++ b/spec/views/sufia/batch_uploads/_form.html.erb_spec.rb
@@ -6,6 +6,7 @@ describe 'sufia/batch_uploads/_form.html.erb', :no_clean do
   let(:form) { Sufia::BatchUploadForm.new(work, ability) }
 
   before do
+    view.lookup_context.view_paths.push "#{CurationConcerns::Engine.root}/app/views/curation_concerns/base"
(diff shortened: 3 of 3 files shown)


mjgiarlo:
Looks great, @hectorcorrea! Can you add spec coverage? Looks like this may be the related spec in CC:

https://github.com/projecthydra-labs/curation_concerns/blob/43f728e52fe1011ba6aa219864beb6d1b2df64d5/spec/forms/work_form_spec.rb

Comments on the PR, @jcoyne?

hectorcorrea:
@mjgiarlo There is already a work_form_spec.rb in Sufia that targets the two new fields: https://github.com/projecthydra/sufia/blob/work_media_thumbnail/spec/forms/work_form_spec.rb#L28-L29

I wonder why it wasn't failing before I made my changes.

mjgiarlo:
That is an excellent question, @hectorcorrea!

app/views/curation_concerns/base/_form_metadata.html.erb
```diff
@@ -11,6 +11,12 @@
             <%= render_edit_field_partial(term, f: f) %>
           <% end %>
         </div>
+
+        <% if f.object.persisted? && f.object.member_ids.present? %>
```
jcoyne: Can you push this block down into the #extended-terms div?

app/views/curation_concerns/base/_form_representative.html.erb
```diff
@@ -0,0 +1,10 @@
+<div class="row">
```
jcoyne: There's no need to add these templates to Sufia if they're a direct copy out of CurationConcerns. It can inherit them.

app/views/curation_concerns/base/_form_representative.html.erb
```diff
@@ -0,0 +1,10 @@
...
+  <fieldset id="representative-media">
+    <legend>
+      Representative Media
+    </legend>
+    <p>Select the file with media that represents this <%= f.object.human_readable_type %>.</p>
+    <%= f.select :representative_id, @form.select_files, {}, { class: 'form-control' } %>
```
jcoyne: should use `f.object.select files` instead of `@form`

app/views/curation_concerns/base/_form_thumbnail.html.erb
```diff
@@ -0,0 +1,10 @@
+<div class="row">
+  <fieldset id="representative-image">
+    <legend>
+       Thumbnail
+    </legend>
+    <p>Select the file to be used as the thumbnail for this <%= f.object.human_readable_type %>.</p>
+    <%= f.select :thumbnail_id, @form.select_files, {}, { class: 'form-control' } %>
```
jcoyne: should use `f.object.select files` instead of `@form`

app/views/curation_concerns/base/_form_representative.html.erb
hectorcorrea: @jcoyne I copied the `f.object.select_files` from CurationConcerns. If I change the syntax here I won't be able to inherit from them as you suggested in your comment above.

app/views/curation_concerns/base/_form_representative.html.erb
hectorcorrea: @jcoyne They are a copy from CurationConcerns. I removed a `<div class="col-md-12">` but they would probably render OK with that div.

hectorcorrea:
@jcoyne @mjgiarlo I moved the rendering of the drop down lists to the extended terms section as Justin suggested and also let it render the default CurationConcerns partials instead of overwriting them in Sufia.

screen shot 2016-04-08 at 8 31 21 am

jcoyne:
Can you add a test that these fields are on the form (probably in here: https://github.com/projecthydra/sufia/blob/master/spec/views/curation_concerns/base/_form.html.erb_spec.rb#L36-L53)?

jcoyne:
Please rebase this too, that should fix many of the errors in CI

jcoyne:
I think you'll need to add this line:
```
view.lookup_context.view_paths.push "#{CurationConcerns::Engine.root}/app/views/curation_concerns/base"
```
to the before block in here: https://github.com/projecthydra/sufia/blob/master/spec/views/sufia/batch_uploads/_form.html.erb_spec.rb#L9

hectorcorrea:
Rebased and added test for the representative media and thumbnail.

PR #1805, opened 2016-04-07
Rubocop-related updates

mjgiarlo:

1. Restore rubocop-rspec (**yes, this must be in both .yml files -- I verified this**)
1. Pin versions more tightly in gemspec
1. Fail build earlier if there are style issues

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -1,3 +1,5 @@
+require: rubocop-rspec
+
 Metrics/CyclomaticComplexity:
   Exclude:
     - 'lib/sufia/arkivo/metadata_munger.rb'
--- a/sufia.gemspec
+++ b/sufia.gemspec
@@ -60,4 +60,6 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency "factory_girl_rails", '~> 4.4'
   spec.add_development_dependency "equivalent-xml", '~> 0.5'
   spec.add_development_dependency "jasmine", '~> 2.3'
+  spec.add_development_dependency 'rubocop', '~> 0.39'
+  spec.add_development_dependency 'rubocop-rspec', '~> 1.4'
 end
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,3 +1,4 @@
+require: rubocop-rspec
 inherit_from: .rubocop_todo.yml
 
 AllCops:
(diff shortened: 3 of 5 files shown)


.rubocop_todo.yml
```diff
@@ -1,3 +1,5 @@
+require: rubocop-rspec
```
escowles: Does `rubocop_todo.yml` need to be renamed to `rubocop.yml`?

jcoyne: I don't think so, I think the rubocop_todo.yml is the things we thing ought to get fixed.

jcoyne: can this line go into .rubocop.yml instead?

mjgiarlo: Nope. They're separate files. `.rubocop.yml` is where we set the styles we care about, and `.rubocop_todo.yml` is where we put temporary exceptions to appease the :cop:. I started off by having this `require` statement in the former file, but that causes complaints because the todo file makes a reference to rspec-related cops. I *could* put the `require` statement in both files, but that seems janky. (I don't love having the `require` in the TODO file which will eventually go away. I'm open to being told that having the `require` in both files is less janky, though.)

mjgiarlo:
This PR needs to be rebased on master once @jcoyne is done with a PR that gets Sufia working atop the latest CC (Now With Hydra::Collections :tm:).

.rubocop_todo.yml
mjgiarlo: OK, the `require` is now in both files. Less janky, I think.

sufia.gemspec
```diff
@@ -60,4 +60,6 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency "factory_girl_rails", '~> 4.4'
   spec.add_development_dependency "equivalent-xml", '~> 0.5'
   spec.add_development_dependency "jasmine", '~> 2.3'
+  spec.add_development_dependency 'rubocop', '~> 0.39'
```
mjgiarlo: I went back and forth on this: should we pin this more tightly to `0.39.0`? The downside is that we'll need to remember to update this on a regular basis, when Rubocop updates. The benefit is that we won't be hit so often with new Rubocop violations (which seems to always happen *after* you've worked up a PR unrelated to Rubocop).

PR #1812, opened 2016-04-08
Remove the well class

jcoyne:
This is sort of a matter of taste, but I feel we've used well gratuitously in the past. Usually this could be added by the application as a style if required.

### DIFF
--- a/app/views/advanced/index.html.erb
+++ b/app/views/advanced/index.html.erb
@@ -7,7 +7,6 @@
 
 <div class="col-md-6">
  <h2><i class=glyphicon glyphicon-pushpin"></i> Search tips</h2>
-  <div class="well">
 <ul id="advanced_search_tips">
     <li><i class="glyphicon glyphicon-chevron-right"></i> Select "match all" to require all fields.</li>
     <li><i class="glyphicon glyphicon-chevron-right"></i> Select "match any" to find at least one field.</li>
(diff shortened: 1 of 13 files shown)


mjgiarlo:
Can you include a screenshot of this UI change?

mjgiarlo:
Thoughts on this change? @mtribone @atz @ggeisler @aeschylus

ggeisler:
FYI: I don't have Sufia or HyBox running locally yet so can't provide an opinion on UI questions without seeing examples. (I plan to install HyBox locally but was hoping to focus on finishing the user research and design deliverables before worrying about my development environment.)

jcoyne:
Old way:
screen shot 2016-04-08 at 12 54 32 pm
screen shot 2016-04-08 at 12 54 41 pm
screen shot 2016-04-08 at 12 55 00 pm


New way:

screen shot 2016-04-08 at 12 56 05 pm
screen shot 2016-04-08 at 12 56 22 pm
screen shot 2016-04-08 at 12 56 32 pm

ggeisler:
Thanks @jcoyne. Yes, I'd agree that removing the well is a good idea. I don't think using a well for large/main sections of content really makes sense. It's better for more supplementary content, generally at least.

mjgiarlo:
Seeing at least four votes in favor, I'm merging. Thanks!

Issue #1815, opened 2016-04-09
Override CurationConcerns PermissionsController layout without copypasta


mjgiarlo:
Sufia uses CurationConcerns' PermissionsController to handle confirming changes to GenericWork permissions and copying those down to FileSets. We override this currently only to set the layout to `nil` which gets the controller to pick up Sufia's layout. This approach introduces technical debt -- whenever CC makes changes to its PermissionsController, we'll need to copy/paste that over. This is not good. CurationConcerns already contains some machinery to deal with theming and layouts within controllers -- we should either use this in Sufia or find a way to make this work in Sufia. Another approach would be to put in a PR in CurationConcerns that extracts PermissionsController to PermissionsControllerBehavior. That way we should be able to pick up changes to the behavior and have a tiny overridden PermissionsController in Sufia.

### Related work

Refs #1731 (See comments for more discussion.)

PR #1816, opened 2016-04-09
PermissionsController should use Sufia layout

### LINKED ISSUE #1731: Override CurationConcerns' PermissionsController to prevent funky layout
### Descriptive summary

When you change permissions and arrive at `/concern/permissions/{ID}/confirm`, you see a view that uses CurationConcerns' styling rather than Sufia's. We should provide a consistent look and feel.

Revert #1726 as part of this work and write a test that fails without the layout override in place. As part of this work, review the [same code in CurationConcerns](https://github.com/projecthydra-labs/curation_concerns/blob/master/app/controllers/curation_concerns/permissions_controller.rb) to see if there are any changes we should make to our local override. (Specifically, do we need to add `authorize! :edit, curation_concern` to the `#copy` method?)

Pertains to Sufia master.

### Expected behavior

Sufia looks like Sufia.

### Actual behavior

screen shot 2016-03-31 at 2 04 13 pm

### Related work

This was removed here: https://github.com/projecthydra/sufia/pull/1726

mjgiarlo:
Fixes #1731

This takes a workaround approach, introducing a small amount of technical debt (ticketed in #1815), namely overriding CC's PermissionsController via copypasta. Note that this approach is not new -- this commit is essentially reverting an earlier commit that introduced this bug.

Also: remove unused redlock cruft and add a .rb extension to a related view spec that hasn't been getting run.


cbeer:
Ew. Do we urgently need this, or can we fix it upstream?

mjgiarlo:
We don't urgently need this -- I was scanning for a ticket I knew I could close quickly, and saw this one. I'm fine if we want to use the #1815 implementation and scrap the workaround here.

mjgiarlo:
If you're itching to take on #1815 soon, maybe we could merge this first to get in the other changes?

PR #1817, opened 2016-04-09
Re-enable browse_everything

jcoyne:


mjgiarlo:
I see @awead started down this path in CC. (He's got a currently busted, possibly incomplete PR there.) Should this work be combined and put in at the CC layer instead of at the sufia layer? Cc: @jpstroop @tpendragon @escowles

escowles:
Yes, putting BrowseEverything in CC makes sense.

jcoyne:
The problem is that CC and Sufia have pretty different ways of doing an upload. Do we want to move the UploadsController to CC too?

njaffer:
I had the same situation of callback for browse_everything done method in the test app (.internal_test_app) but when Sufia is used as a gem in an app, the done method is called, and it works great. In the last DevCon I was working with @afred on this one, but we were not able to figure out why the done is not triggered in the test app. May be some path issues?

mjgiarlo:
@jcoyne I'd pose that to the CC developers to hear their assessment.

awead:
@jcoyne @escowles yeah, I started down the path of moving BE into CC, but ran out of time:
https://github.com/projecthydra-labs/curation_concerns/pull/690
I could pick it back up if it would solve this situation, but it sounds like maybe they should be kept separate?

mbklein:
Hey all. I'm just dotting this thread in case I can provide some BrowseEverything help.

jcoyne:
@mbklein Hey Michael. One problem we're having is that we can't the call to  BrowseEverything in the asset pipeline, because the BrowseEverything constant is declared on DOM load. https://github.com/projecthydra/browse-everything/blob/master/app/assets/javascripts/browse_everything/behavior.js.coffee#L1
@aeschylus is working on refactoring that.

mbklein:
OK

PR #1822, opened 2016-04-12
Refactor javascript.

jcoyne:

* Push each functional area to its own file. Thus sufia.js becomes just a
manifest.
* Leverage bootstrap's alert function instead of having our own.


mjgiarlo:
Looks good to me.

PR #1833, opened 2016-04-13
Remove inline HTML definition list, fix links

### LINKED ISSUE #1830: Inline HTML cannot also use markdown: some README links broken
We are trying to use markdown inside inline-HTML (inside markdown).  Specifically, see the `Terminology` definition list.  We should pick one format and use it.  If HTML is really needed, it should be broken out of our mega-README and into its own semantically coherent HTML.

Other links are possibly just badly written, like `[[http:link.to.place]]`, using double brackets.

atz:
Fixes #1830 .  Added enough formatting for github to display intended-clickable links as links.  

Note: removed inaccurate declaration of PhantomJS version dependency.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,16 +37,16 @@ After running these steps, browse to [localhost:3000](http://localhost:3000/) an
       * [Derivatives](#derivatives)
     * [Environments](#environments)
     * [Ruby](#ruby)
-  * [Creating a Sufia-based app](#creating-a-sufia-based-app)
+  * [Creating a Sufia\-based app](#creating-a-sufia-based-app)
     * [Rails](#rails)
-    * [Sufia's Ruby-related dependencies](#sufias-ruby-related-dependencies)
-      * [Pagination](#pagination)
+    * [Sufia's Ruby\-related dependencies](#sufias-ruby-related-dependencies)
     * [Install Sufia](#install-sufia)
     * [Database tables and indexes](#database-tables-and-indexes)
-    * [Start Solr](#start-solr) and [FCRepo](#start-fcrepo)
+    * [Start Solr](#start-solr)
+    * [Start FCRepo](#start-fcrepo)
     * [Start background workers](#start-background-workers)
     * [Spin up the web server](#spin-up-the-web-server)
-  * [Managing a Sufia-based app](#managing-a-sufia-based-app)
+  * [Managing a Sufia\-based app](#managing-a-sufia-based-app)
     * [Production concerns](#production-concerns)
       * [Identifier state](#identifier-state)
       * [Web server](#web-server)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -283,37 +287,41 @@ For the remainder of the background worker documentation, it is assumed that you
 
 ### Terminology
 
-<dl>
-<dt>Resque</dt>
-<dd>Resque is a <a href="https://en.wikipedia.org/wiki/Message_queue">message queue</a> that is used by Sufia to manage long-running or slow processes. (Sufia builds its jobs using Rails's ActiveJob framework, so you are free to use another queuing system, e.g., DelayedJob or Sidekiq.)</dd>
+##### Resque
+
```
mjgiarlo: Should this be an h4? The heading above is an h3.

mjgiarlo: Also applies to sister headers below.

mjgiarlo:
Tweak the header level, and can you also include a regenerated TOC per https://github.com/projecthydra/sufia/blob/master/README.md#regenerating-the-readme-toc

README.md
atz: I consider this ad-hoc replacement for definition list at the correct "h".  If we had enough terms to break into sections, section headings would be h4.

atz:
Happy to regenerate TOC, but also willing to defer that to a release-build step.  That way it only needs to be done once.

elrayle:
Can you make this change to the 'Run the test suite' section?

## Run the test suite

### Prerequisites
* Make sure all [basic prerequisites](https://github.com/projecthydra/sufia#prerequisites) are running.
* Additional prerequisite for tests: [PhantomJS](http://phantomjs.org/), version 1.9.x (NOTE: 2.1.1 has also been confirmed to work.)

### Test app
Generate the test app.  *NOTE: Run this only once.*
```
rake engine_cart:generate
```

This generates `sufia/.internal_test_app` directory.  The tests will run against this test app.

### Run tests
```
rake spec
```

### Testing FAQ
* Where is rake jetty?  It was retired.  Solr and Fedora are started on their own.
* How can I start a test instance of Solr?  **DO NOT USE FOR PRODUCTION**
```
# from sufia root in a separate terminal window
solr_wrapper -d solr/config/ --collection_name hydra-development
```
  * Test that Solr is running at: [[http://localhost:8983/]] 
  * Solr Core name:  hydra-development
* How can I start a test instance of Fedora?  **DO NOT USE FOR PRODUCTION**
```
# from sufia root in a separate terminal window
fcrepo_wrapper -p 8984
```
  * Test that Fedora is running at: [[http://localhost:8984/]] 
* The generated test app isn't doing what I expected after making changes to Sufia.  What can I do?  Generally, engine cart will pick up changes to Sufia.  If not, try the following to regenerate the test app.
```
rake engine_cart:clean
rake engine_cart:generate
```

elrayle:
I would also recommend removing solr_wrapper and fedora_wrapper commands from Start Solr and Start FCRepo sections.  You could point them to the Run the test suite section if they want to experiment with Sufia.  And this will avoid the risk of having the command different in two parts of the doc.  In Start Solr and Start FCRepo, the recommendation would be to use production approaches for starting Solr and Fedora.

mjgiarlo:
@elrayle sorry, can you create a new issue for those changes? I'd like to get this PR merged once its TOC is sorted out.

PR #1838, opened 2016-04-13
Accept resource_type in Batch Create

### LINKED ISSUE #1796: Batch upload form should let you set a resource type per uploaded file
Ref https://github.com/projecthydra/sufia/issues/1674

Similar to #1773

https://wiki.duraspace.org/display/hydra/Batch+New+Works+-+Phase+1+-+UI+Wireframes#mockup_p1-works-batch_new-step1_upload_files

hectorcorrea:
Fixes #1796

Allows user to select a different resource_type for each file when using Batch Create.

screen shot 2016-04-13 at 5 05 04 pm

I can get the screen to allow the user to select a resource type, and the selected resource type makes it to the controller, but I am having problems getting the value to the `BatchCreateJob.perform_later`. Any ideas?

### DIFF
--- a/app/views/sufia/batch_uploads/_form_files.html.erb
+++ b/app/views/sufia/batch_uploads/_form_files.html.erb
@@ -74,7 +74,19 @@
 </script>
 <!-- The template to display files available for download -->
 <script id="template-download" type="text/x-tmpl">
-{% for (var i=0, file; file=o.files[i]; i++) { %}
+{%
+
+    // I am not happy with this global declaration.
+    // Using "var setAllResourceTypes" or "function setAllResourceTypes()" does not work. Not sure why.
+    // Perhaps this kind of declaration is not a problem due to the way the JavaScript is executed???
+    setAllResourceTypes = function() {
+        var firstResourceType = $(".resource_type_dropdown")[0].value;
+        $(".resource_type_dropdown").each(function(index, element) {
+            element.value = firstResourceType;
+        });
+    };
+
+    for (var i=0, file; file=o.files[i]; i++) { %}
     <tr class="template-download fade">
         <td>
           <div class="row">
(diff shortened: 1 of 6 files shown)


jcoyne:
Try using https://github.com/projecthydra/sufia/blob/master/app/forms/sufia/batch_upload_form.rb instead.

hectorcorrea:
Screenshot with the button to set the same resource type to all files (still need to figure out why it is drawn for all files, it should be displayed only for the first file)
screen shot 2016-04-14 at 2 06 24 pm

mjgiarlo:
Looking good, @hectorcorrea!

app/views/sufia/batch_uploads/_form_files.html.erb
```diff
@@ -100,6 +112,18 @@
...
                   <input type="text" class="form-control" name="title[{%=file.id%}]" id="title_{%=file.id%}" value="{%=file.name%}">
                 </div>
+                <label for="resource_type_{%=file.id%}" class="col-sm-5 control-label">Resource Type</label>
+                <div class="col-sm-7">
+                  <select class="form-control resource_type_dropdown" name="resource_type[{%=file.id%}]" id="resource_type_{%=file.id%}" value="{%=file.name%}">
+                    <% Sufia.config.resource_types.each do |type| %>
+                        <option value="<%= type[0] %>"><%= type[1] %></option>
+                    <% end %>
+                  </select>
+                  <!-- TODO: Why is the button drawn for all files? -->
```
hectorcorrea: I cannot get the button to be displayed only for the first file. It shows for all of them. Any ideas?

I can add a separate issue to track this problem if you want to go ahead and merge the current PR in time for the demo tomorrow.

app/views/sufia/batch_uploads/_form_files.html.erb
```diff
@@ -74,7 +74,19 @@
 </script>
 <!-- The template to display files available for download -->
 <script id="template-download" type="text/x-tmpl">
-{% for (var i=0, file; file=o.files[i]; i++) { %}
+{%
+
+    // I am not happy with this global declaration.
+    // Using "var setAllResourceTypes" or "function setAllResourceTypes()" does not work. Not sure why.
+    // Perhaps this kind of declaration is not a problem due to the way the JavaScript is executed???
+    setAllResourceTypes = function() {
```
hectorcorrea: I am not sure why neither `function setAllResourceTypes() {...}` nor `var setAllResourceTypes = function() {...}` worked. Not a show stopper but I thought I would highlight this in case any one has ideas on why this is the case.

app/views/sufia/batch_uploads/_form_files.html.erb
mjgiarlo: Pls do. Thanks!

PR #1839, opened 2016-04-14
Dynamic cancel button

### LINKED ISSUE #1831: jquery-ui upload buttons dynamic behavior
Split from #1734.  

"Start Upload", "Cancel Upload" and "Delete" buttons are live, have hover effect, clickable, have "active" effect when clicked... when no files have been added. It is a whole lot of interface for nothing. These 3 buttons are only relevant after the list is populated. (Consider making them part of the populated list, make their visibility dependent on the list being populated or disable them until the list is populated.) Since each file also has start and cancel buttons, the two "list scope" counterparts are not even usefully distinct until after a **2nd file** is added!

How is `Cancel` different than `Delete`?  Neither actually touch the server side.

atz:
Dynamically reveal cancel button  …
Remove list scope other buttons.
Fixes #1831

TODO: consolidate templates between:
 `app/views/sufia/batch_uploads/_form_files.html.erb` and   `app/views/curation_concerns/base/_form_files.html.erb`

### DIFF
--- a/app/assets/javascripts/sufia/uploader.js
+++ b/app/assets/javascripts/sufia/uploader.js
@@ -34,9 +34,9 @@
         url: '/uploads/',
         type: 'POST'
       })
-      // .bind('fileuploadprocessfail', function (e, data) {
-      //   console.log('Processing ' + data.files[data.index].name + ' failed.');
-      // })
+      .bind('fileuploadadded', function (e, data) {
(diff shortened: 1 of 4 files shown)


mjgiarlo:
There are a couple specs that need tweaking. Our expectations have changed.

mjgiarlo:
@atz once this PR is :green_apple:, I'd like to merge it. Can you ticket the above template consolidation TODO separately?

PR #1850, opened 2016-04-14
Scroll no longer moves when clicking tabs.

### LINKED ISSUE #1794: Clicking a tab on the upload page should not cause the scroll to jump.
I think we need to invoke the tab javascript with preventDefault() as is shown here:
http://getbootstrap.com/javascript/#tabs-usage

jcoyne:

Fixes #1794

### DIFF
--- a/spec/support/proxies.rb
+++ b/spec/support/proxies.rb
@@ -1,6 +1,7 @@
 module ProxiesHelper
   def create_proxy_using_partial(*users)
     users.each do |user|
+      sleep(1)
       first('a.select2-choice').click
       find(".select2-input").set(user.user_key)
       expect(page).to have_css("div.select2-result-label")
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,19 @@
 language: ruby
-cache: bundler
 sudo: false
+
+cache:
+  bundler: true
+  directories:
+    - "travis_phantomjs"
+
+before_install:
+  - "phantomjs --version"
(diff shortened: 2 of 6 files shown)


jcoyne:
I can't duplicate this failure on my machine. Does it fail for anyone else?

hectorcorrea:
@jcoyne could you rebase this?

spec/support/proxies.rb
```diff
@@ -1,6 +1,7 @@
 module ProxiesHelper
   def create_proxy_using_partial(*users)
     users.each do |user|
+      sleep(1)
```
cbeer: Is this because of those tab fade-in animations? In other projects, I think we've had a way to disable them globally; maybe @mejackreed remembers off-hand?

jcoyne: That's a possibility. I don't know why.

mejackreed:
```javascript
jQuery.fx.off = true;
```
https://makandracards.com/coffeeandcode/7503-disable-jquery-animations-during-rails-tests
https://github.com/sul-dlss/sul-embed/search?utf8=%E2%9C%93&q=jQuery.fx.off+%3D+true%3B

jcoyne: Nope, it was not jquery fx. I think it's something that select2 is doing.

jcoyne:
@hectorcorrea rebased and ready to ship

PR #1867, opened 2016-04-15
Add human readable type for generated GenericWorks

### LINKED ISSUE #1858: Add human readable type for generated GenericWorks
### Descriptive summary

This [Sufia install generator step](https://github.com/projecthydra/sufia/blob/master/lib/generators/sufia/install_generator.rb#L86-L90) should also inject `self.human_readable_type = 'Work'` into the application's `app/models/generic_work.rb`.

Pertains to Sufia master.

### Expected behavior

Instances of `GenericWork` in the UI are referred to as "works" instead of "generic works".

atz:
Closes #1858

### DIFF
--- a/lib/generators/sufia/install_generator.rb
+++ b/lib/generators/sufia/install_generator.rb
@@ -85,7 +85,8 @@ module Sufia
 
     def inject_sufia_generic_work_behavior
       insert_into_file 'app/models/generic_work.rb', after: 'include ::CurationConcerns::BasicMetadata' do
-        "\n  include Sufia::WorkBehavior"
+        "\n  include Sufia::WorkBehavior" \
+          "\n  self.human_readable_type = 'Work'"
       end
     end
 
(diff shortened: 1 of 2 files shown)


lib/generators/sufia/install_generator.rb
```diff
@@ -85,7 +85,8 @@ def inject_sufia_collection_behavior
 
     def inject_sufia_generic_work_behavior
       insert_into_file 'app/models/generic_work.rb', after: 'include ::CurationConcerns::BasicMetadata' do
-        "\n  include Sufia::WorkBehavior"
+        "\n  include Sufia::WorkBehavior" \
+          "\n  self.human_readable_type = 'Work'"
```
cbeer: What is `human_readable_type`, and should it actually be an i18n string?

jcoyne: It's a relic from Curate. Yes, it probably should have been i18n.

mjgiarlo:
@atz one expectation to fix in `curation_concerns/base/edit.html.erb`

PR #1876, opened 2016-04-15
Fix margin on collection form. Fixes #1874

### LINKED ISSUE #1874: new collection page doesn't have correct styling.
might be missing a `.row` class

jcoyne:


atz:
Definitely support removing form-horizontal in almost all the places I've seen it so far.

Issue #1884, opened 2016-04-18
Browse-everything updates unable to be pushed.


aeschylus:
There are a number of changes that can enable the browse-everything feature to be initialised without placing javascript inside the template.

These require changes in curation-concerns, the browse-everything engine, and here (sufia). The change has to be made in [the initialisation of the BE component](https://github.com/projecthydra/browse-everything/blob/master/app/assets/javascripts/browse_everything/behavior.js.coffee#L195-L203) (use a constructor); the initialisation boilerplate in curation_concerns/browse-everything, and an initialisation function needs to be added to the sufia app.js. The reorganisation will work because the [blacklight.onload function](https://github.com/projectblacklight/blacklight/blob/d55d6f566d203a092f7f40987a37d8988df5500d/app/assets/javascripts/blacklight/core.js) already binds to both the pageload (turbolinks) event and the domready events.

If someone could add me as a collaborator on these repos I can send a PR without forking the repos.

PR #1885, opened 2016-04-18
[wip] Browse everything

aeschylus:
Fixes #issuenumber ; refs #issuenumber

More detailed description, if necessary. Try to be as descriptive as you can: even if you think that the PR content is obvious, it may not be obvious to others. Include tracebacks if helpful, and be sure to call out any bits of the PR that may be work-in-progress.

Description can have multiple paragraphs and you can use code examples inside:

``` ruby
class PostsController
  def index
    respond_with Post.limit(10)
  end
end
```

- 
- 
- 

### DIFF
--- a/app/assets/javascripts/sufia/app.js
+++ b/app/assets/javascripts/sufia/app.js
@@ -8,21 +8,16 @@ Sufia = {
   },
 
   browse_everything: function() {
-    $('#browse-btn').browseEverything({
-      onSelection: function(data) {
+    $('#browse-btn').browseEverything()
+      .done(function(data) {
         $('#status').html(data.length.toString() + " <%= t('sufia.upload.browse_everything.files_selected')%>");
         $('#submit-btn').html("Submit "+data.length.toString() + " selected files");
         var evt = {
           isDefaultPrevented: function() { return false; }
         };
         var files = $.map(data, function(d) { return { name: d.file_name, size: d.file_size, id: d.url }; });
-
-        $.blueimp.fileupload.prototype.options.done.call(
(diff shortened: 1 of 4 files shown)


mjgiarlo:
@aeschylus When you get a few minutes, can you update the description of this issue to say what work is remaining? Is it just to get the specs green, or is there more to it?

aeschylus:
I closed this other issue related to this. https://github.com/projecthydra/sufia/issues/1884

This PR depends on a dependency (Browse Everything) being merged in in the correct form. Here's some more information from that closed (non)issue.

The goal with these changes is to allow the browse-everything feature to be initialised without placing javascript inside the template.

The change has to be made in the initialisation of the [BE component](https://github.com/projecthydra/browse-everything/tree/expose_as_class) (use a constructor); the initialisation boilerplate in views/curation_concerns/browse-everything in sufia, and an initialisation function needs to be added to the sufia app.js (shown in this PR). The reorganisation will work because the blacklight.onload function already binds to both the pageload (turbolinks) event and the domready events, so we don't need the initialisation code stuck in the template.

The outstanding Browse Everything work is in the [expose_as_class](https://github.com/projecthydra/browse-everything/tree/expose_as_class) branch

jcoyne:
Can you make a PR in BrowseEverything?

atz:
Branch could use a rebase, too.

mjgiarlo:
@jcoyne expressed some interest in this issue on 12/20/2016. :smile:

jcoyne:
I don't think we're ever getting back to this.

PR #1887, opened 2016-04-18
Add files button stays enabled even on small screens

### LINKED ISSUE #1873: Add files button becomes disabled
### Descriptive summary

On certain screen widths the "Add Files" button becomes disabled. I am not sure if the button becomes disabled or just un-clickable but the net result is that under certain screen-width values the user cannot add files. 

The two screenshots below show an example of a width that worked (779) and one that does not (724)

screen shot 2016-04-15 at 11 56 02 am

screen shot 2016-04-15 at 11 55 39 am

1. Click on Create New Work
1. Resize the screen width as shown in the screenshots above
1. Notice how the "Add Files" button cannot be clicked.

This happens in FireFox, I have not tested on other browsers.

atz:
If I used `sm` instead of `xs`, then it only punted the problem (i.e., button still disabled at a smaller size).  

Fixes #1873

I hit both templates to minimize interface drift.

### DIFF
--- a/app/jobs/batch_create_job.rb
+++ b/app/jobs/batch_create_job.rb
@@ -1,6 +1,5 @@
 class BatchCreateJob < ActiveJob::Base
   include Hydra::PermissionsQuery
-  include CurationConcerns::Messages
 
   queue_as :batch_create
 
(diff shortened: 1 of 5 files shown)


atz:
Build error is happening underneath me `class:BatchCreateJob: uninitialized constant CurationConcerns::Messages (NameError)`.  I think we can be certain that was not caused by a CSS styling change.

atz:
Also, coveralls is wack with its `75%` coverage drop.  lulz.

atz:
Now mergeable.

PR #1896, opened 2016-04-19
Adding FeaturedWorks migration for original Sufia 6 table

### LINKED ISSUE #1506: Featured Works - migration
Implement code to migrate "Featured Files" to "Featured Works"

Keep in mind that the SQL table `featured_works` was named `featured_works` (not `featured_generic_files`) since before Sufia 7. However, this table used to have a `generic_file_id` field that now has been replaced with `generic_work_id`

### LINKED ISSUE #1893: Need a migration for featured_works
### Descriptive summary

The featured_works table has generic_work_id for a column, but in Sufia 6 apps, it's generic_file_id. We'll need a migration to change that.

awead:
Fixes #1893 

### DIFF
--- a/lib/generators/sufia/templates/migrations/change_featured_work_generic_file_id_to_generic_work_id.rb
+++ b/lib/generators/sufia/templates/migrations/change_featured_work_generic_file_id_to_generic_work_id.rb
@@ -1,0 +1,6 @@
+class ChangeFeaturedWorkGenericFileIdToGenericWorkId < ActiveRecord::Migration
+  def change
+    return unless column_exists?(:featured_works, :generic_file_id)
+    rename_column :featured_works, :generic_file_id, :generic_work_id
+  end
+end
--- a/lib/generators/sufia/upgrade700_generator.rb
+++ b/lib/generators/sufia/upgrade700_generator.rb
@@ -27,7 +27,8 @@ This generator for upgrading sufia from 6.0.0 to 7.0 makes the following changes
       'change_trophy_generic_file_id_to_generic_work_id.rb',
       'change_proxy_deposit_generic_file_id_to_generic_work_id.rb',
       'change_audit_log_generic_file_id_to_file_set_id.rb',
-      'change_proxy_deposit_request_generic_file_id_to_generic_work_id.rb'
+      'change_proxy_deposit_request_generic_file_id_to_generic_work_id.rb',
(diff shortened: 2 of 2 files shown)


jcoyne:
Travis says: :no_good:

lib/generators/sufia/templates/migrations/change_featured_work_generic_file_id_to_generic_work_id.rb
```diff
@@ -0,0 +1,6 @@
+class ChangeFeaturedWorkGenericFileIdToGenericWorkId < ActiveRecord::Migration
+  def change
+    return unless column_exists?(:featured_works, :generic_file_id)
+    rename_column :featured_works, :generic_file_id, :generic_work_id
```
jcoyne: I don't see where you're finding the `generic_file_id` column.  It was created as `generic_work_id` here: https://github.com/projecthydra/sufia/blob/master/lib/generators/sufia/templates/migrations/create_featured_works.rb#L5

jcoyne: Ah! I see. Maybe part of this commit needs to be reverted: https://github.com/projecthydra/sufia/commit/2cb6ef129d1712fde5c2f6fc4d028b203f230805#diff-5bb87ab96f0bcf36eb184441813db0e7L5

awead: @jcoyne the above commit is ok because it works for a first-time installation. When someone is upgrading from 6, they'll need to rename that column because the table already exists. I've put it under the upgrade700 step, but since the engine cart is running this generator as well, we need to skip it since the table and its correct column already exists.

Another option is to make this its own generator step, separate from upgrade700, and then the engine_cart wouldn't call it.

mjgiarlo: @awead :speech_balloon: 
> ... I've put it under the upgrade700 step, but since the engine cart is running this generator as well, we need to skip it since the table and its correct column already exists. ...

FWIW, this wouldn't be the only migration that has a conditional in it, so this solution is OK. We have a backlog ticket to work on spiffying up our migration-having generators, so we'll fix that up at some point. For now, :shipit:

PR #1898, opened 2016-04-19
Bootstrap smallboxes

### LINKED ISSUE #1875: when screen is a large width, seach field gets small
screen shot 2016-04-15 at 11 15 36 am

It looks okay at a medium width.

atz:
fixes #1875 

Consolidates feature to use the same CC template.  Delivers more uniform bootstrap styling.

### DIFF
--- a/app/views/collections/_sort_and_per_page.html.erb
+++ b/app/views/collections/_sort_and_per_page.html.erb
@@ -3,12 +3,12 @@
     <%= render partial: 'collections/form_for_select_collection', locals: { user_collections: @user_collections }  %>
   </div>
 
-  <% if  params[:action] == "edit" %>
+  <% if params[:action] == "edit" && @response.response['numFound'] > 1 %>
   <div class="batch-toggle">
-     <% session[:batch_edit_state] = "on" %>
+    <% session[:batch_edit_state] = "on" %>
     <%= button_for_remove_selected_from_collection collection %>
-   </div>
-  <%end %>
+  </div>
+  <% end %>
    <div class="sort-toggle">
      <%# kind of hacky way to get this to work on catalog and folder controllers.  May be able to simple do {action: "index"} but I'm not sure -%>
      <% unless @response.response['numFound'] < 2 %>
--- a/app/views/collections/edit.html.erb
+++ b/app/views/collections/edit.html.erb
@@ -17,7 +17,7 @@
 <% end %>
 
 <h2>Manage Items in this Collection</h2>
-<%= render 'search_collection_dashboard_form'%>
+<%= render 'search_form', presenter: @collection %>
 <%= render 'my/did_you_mean' %>
 <%= render 'my/facet_selected' %>
 <%= render 'collections/sort_and_per_page', collection: @collection %>
--- a/app/views/collections/edit.html.erb
+++ b/app/views/collections/edit.html.erb
@@ -16,8 +16,14 @@
   </div>
 <% end %>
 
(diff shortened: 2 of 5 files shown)


app/views/collections/edit.html.erb
```diff
@@ -16,8 +16,14 @@
   </div>
 <% end %>
 
-<h2>Manage Items in this Collection</h2>
-<%= render 'search_collection_dashboard_form'%>
+<br/>
```
cbeer: `<br />`? Can we style things appropriately instead?

app/views/collections/edit.html.erb
```diff
@@ -16,8 +16,14 @@
...
+<br/>
+<div class="row">
+  <h2 class="col-xs-12 col-sm-6"><%= t('Manage Items in this Collection') %></h2>
```
cbeer: Does this `#t` actually work?

atz: Yep.  I don't have a populated translation, but it gives me back the parameter string at least.

atz:
Rebased.

mjgiarlo:
@atz Inclined to make the changes @cbeer suggests above?

app/views/collections/edit.html.erb
cbeer: I'd guess it also gives a `span.missing_translation`, and might be difficult to make work in the i18n file with all the spaces (and the lack of expected namespacing)?

atz:
There was no such warning delivered.  If it is important, there probably should be one.  

But I agree using the translation file is more proper.  Pushed modified branch.

atz:
As for `<br/>`, I don't care.  It's not worth holding anything up about.  I know we aren't supposed to mix bootstrap and inline styles, so I attempted to find the correct bootstrapism and felt like it was a wasted effort.  Vertical separator was removed in bootstrap 3.  _Maybe_ something will be included for [bootstrap 4](https://github.com/twbs/bootstrap/issues/4286#issuecomment-36331550).  

If you want to declare what "appropriate style" would be, I can transcribe it for you.  Or in the same amount of time, you could push to this topic branch.

cbeer:
What are we trying to do with the `<br />` anyway? Does it actually look bad without them?

A far preferable solution, in my mind, may involve adding a class or id to an element (the `h2`, or whatever element we're trying to create a margin from) and add explicit styling for it.

atz:
Yeah, it looks bad w/o some separation.  Adding an ID is unnecessary, since presumably the need for vertical separation is not unique to a single element.  What class makes sense?  `.edit_collection`?  That provides:
```scss
margin-top: 2 * $padding-large-vertical;
```

And also doesn't appear to be used anywhere in the application.

cbeer:
I was assuming a bespoke class, named appropriately for whatever we're trying to style.

app/views/collections/_sort_and_per_page.html.erb
```diff
@@ -3,12 +3,12 @@
     <%= render partial: 'collections/form_for_select_collection', locals: { user_collections: @user_collections }  %>
   </div>
 
-  <% if  params[:action] == "edit" %>
+  <% if params[:action] == "edit" && @response.response['numFound'] > 1 %>
```
mjgiarlo: Can you say more about this new condition?

atz: When there is nothing to display, don't attempt to display it.

cbeer: How about `any?` or `many?`?

PR #1902, opened 2016-04-19
Use hydra-batch-edit 2.x that removes its dependency on hydra-collections

cbeer:


jcoyne:
@cbeer I added a commit which I *think* finally cuts ties with hydra-collections

cbeer:
Ok. I'll get the test suite passing for `hydra-batch-edit` and get a new release out.

atz:
Seems like some substantive errors in CI.

mjgiarlo:
Yep, still busted because of missing helpers.

PR #1908, opened 2016-04-20
Don't default rights

### LINKED ISSUE #1862: Work show pages & forms -- no default rights and add text if no metadata
### Descriptive summary

- [ ] the default rights in our upload/edit forms for works and batches should be blank, rather than the first entry in the dropdown.
- [ ] if no metadata is present on the show page, display "No metadata is defined for this work." in the metadata section. (Impossible to test until rights change above is made.)

Pertains to Sufia master.

### Expected UI

https://wiki.duraspace.org/x/DAF6B

### Related work

Refs #1722, #1843

atz:
Allow non-selection of rights and collapse metadata display when empty.

Uses translation.

Fixes #1862

### DIFF
--- a/app/views/curation_concerns/base/_metadata.html.erb
+++ b/app/views/curation_concerns/base/_metadata.html.erb
@@ -1,12 +1,18 @@
 <h2><%= t('.header') %></h2>
-<table class="table table-striped <%= dom_class(presenter) %> attributes" <%= presenter.microdata_type_to_html %>>
-  <thead>
-    <tr><th>Attribute Name</th><th>Values</th></tr>
-  </thead>
-  <tbody>
-    <%= render 'attribute_rows', presenter: presenter %>
-    <%= presenter.attribute_to_html(:embargo_release_date) %>
-    <%= presenter.attribute_to_html(:lease_expiration_date) %>
-    <%= presenter.attribute_to_html(:rights) %>
-  </tbody>
-</table>
+<% attributes ||= [ :embargo_release_date, :lease_expiration_date, :rights] %>
+<% presentations = attributes.map { |attribute| presenter.attribute_to_html(attribute) }.reject { |x| x.empty? } %>
+<% if presentations.empty? %>
+  <p><%= t('.no_metadata') %></p>
+<% else %>
+  <table class="table table-striped <%= dom_class(presenter) %> attributes" <%= presenter.microdata_type_to_html %>>
+    <thead>
+      <tr><th><%= t('.attribute_name_label') %></th><th><%= t('.attribute_values_label') %></th></tr>
+    </thead>
+    <tbody>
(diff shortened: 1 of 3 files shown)


app/views/records/edit_fields/_rights.html.erb
```diff
@@ -1,4 +1,6 @@
+<% opts = RightsService.select_options %>
+<% opts.unshift ['', ''] %>
```
jcoyne: Can you pass `include_blank: true` to `#input` instead of manipulating the collection?

app/views/curation_concerns/base/_metadata.html.erb
```diff
@@ -1,12 +1,18 @@
...
-  </thead>
-  <tbody>
-    <%= render 'attribute_rows', presenter: presenter %>
-    <%= presenter.attribute_to_html(:embargo_release_date) %>
-    <%= presenter.attribute_to_html(:lease_expiration_date) %>
-    <%= presenter.attribute_to_html(:rights) %>
-  </tbody>
-</table>
+<% attributes ||= [ :embargo_release_date, :lease_expiration_date, :rights] %>
+<% presentations = attributes.map { |attribute| presenter.attribute_to_html(attribute) }.reject { |x| x.empty? } %>
```
jcoyne: I think it would be great to put this logic in the presenter.  So we can say `presenter.metadata_empty?`

app/views/records/edit_fields/_rights.html.erb
atz: Yeah, that makes sense.

app/views/curation_concerns/base/_metadata.html.erb
atz: Agree.  Will give it a shot.

atz: @jcoyne OK, practical questions.  The presenter does not necessarily fully know which of its attributes are metadata and which are just... regular data or something else.  We delegate these to the solr_document:
```
:based_near
:related_url
:depositor
:identifier
:resource_type
:tag
```

`CurationConcerns::WorkShowPresenter` delegates what a comment calls "# Metadata Methods":
```
:title
:date_created
:date_modified
:date_uploaded
:description
:creator
:contributor
:subject
:publisher
:language
:embargo_release_date
:lease_expiration_date
:rights
```

But that set may change underneath us (i.e., minor version can add to it) and there is no enumeration of that set for us to reference downstream.  Also, a check like `empty?` should be cheap, right?  It shouldn't have to interrogate 20 methods underneath it.  Lastly, right here we only care about 3 fields.  Some metadata fields could well be populated, but not the 3 we care about.  We are worried about collapsing the display of 3 possible fields as part of the display logic.  I'm fine keeping that in the template.

Issue #1832, opened 2016-04-12
Batch upload async error


atz:
I used batch upload to post 2 files.  I waited and shortly thereafter received a notification via [/notifications](http://localhost:3000/notifications) :
```
Failing batch create  -- The batch create for x@y.com failed
```

This ticket has two parts:
- [ ] batch create shouldn't fail
- [ ] when it does, the error message should be more informative

mjgiarlo:
If someone other than @atz could pick this up and try to reproduce it, that'd be a great start.

jcoyne:
It isn't failing for me. Is there anything in the rails log when this happens? Can you reproduce it dependably? 
<img width="958" alt="screen shot 2016-04-13 at 11 01 35 pm" src="https://cloud.githubusercontent.com/assets/92044/14516748/bc481b9e-01cb-11e6-8d92-ae5d8f277235.png">

atz:
Was dependable on Tuesday.  Will try it again tomorrow.

atz:
Still failing in the same way.

atz:
Hrm, on a subsequent attempt, passed.  Interesting.

atz:
Well, if we check the first box and say "it isn't failing right now!", we still need a more informative message for when it does fail.

atz:
We can close this, or at least backlog it.  Not really a release blocker anymore.

mjgiarlo:
Closing per @atz's recommendation. We can re-open if we repro it later.

Issue #1779, opened 2016-04-06
Add logs/status for batch and item level operations


jcoyne:
For BatchCreate there should be a log that shows the overall status ('processing', 'failed',  'success'). There should also be logs for each item (create work operations in this case). 

Perhaps something like this: https://github.com/curationexperts/tufts_models/blob/master/app/models/batch.rb (but do not serialize fields as it introduces race conditions).

awead:
Possibly related to https://github.com/psu-stewardship/scholarsphere/issues/373 where we need to know when a single work is finished processing.

PR #1917, opened 2016-04-22
Add better logging for batch update jobs

jcoyne:

screen shot 2016-04-25 at 11 07 49 am

screen shot 2016-04-25 at 11 08 00 am

Ref #1832 Ref #1779

### DIFF
--- a/app/jobs/batch_create_job.rb
+++ b/app/jobs/batch_create_job.rb
@@ -1,63 +1,33 @@
 class BatchCreateJob < ActiveJob::Base
-  include Hydra::PermissionsQuery
-
   queue_as :batch_create
 
-  attr_accessor :saved, :success
+  before_enqueue do |job|
+    log = job.arguments.last
+    log.pending_job(self)
+  end
 
   # This copies metadata from the passed in attribute to all of the works that
   # are members of the given upload set
-  def perform(user, titles, resource_types, uploaded_files, attributes)
-    @saved = []
-    @success = true
+  def perform(user, titles, resource_types, uploaded_files, attributes, log)
+    log.performing!
 
     titles ||= {}
     resource_types ||= {}
 
-    create(user, titles, resource_types, uploaded_files, attributes)
-    send_user_message(user)
+    create(user, titles, resource_types, uploaded_files, attributes, log)
   end
 
   private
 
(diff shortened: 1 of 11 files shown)


mjgiarlo:
@jcoyne Just to confirm: still WIP?

jcoyne:
Yes still WIP. TODO:
* Make a controller/view to show the status of a batch
* Decompose a BatchCreateJob to fan out the create jobs
* Roll up the status onto the BatchCreate
* Update notifications to direct user at the log.
* Logging for non-batch things.

mjgiarlo:
This was blocked by a CC PR that is now merged. Also must be rebased.

jcoyne:
@mjgiarlo Okay, ready for review.

app/jobs/batch_create_job.rb
```diff
@@ -1,63 +1,33 @@
...
-    @saved = []
-    @success = true
+  def perform(user, titles, resource_types, uploaded_files, attributes, log)
+    log.performing!
 
     titles ||= {}
     resource_types ||= {}
 
-    create(user, titles, resource_types, uploaded_files, attributes)
-    send_user_message(user)
```
mjgiarlo: Not worth sending the user a notification about this any longer?

mjgiarlo:
How do users get to the views in the screenshots above?

app/jobs/batch_create_job.rb
jcoyne: It's moved to the BatchCreateOperation, and it occurs when all of the parts are done.

jcoyne:
Presently you have to go directly to the URL.  I hope that we can integrate them somewhere better than that in the future.

app/jobs/batch_create_job.rb
mjgiarlo: Ah, I see now. :+1:

mjgiarlo:
@jcoyne :clap: 

I'll ticket the UI integration piece separately for 7.1.0.

ggeisler:
In the first screenshot, the "performing" label is using the same  `.label-success` class as the "success" label. I wonder if using the `.label-info` class would be better for the "performing" state? That would visually distinguish the two different states ("performing" would have a blue label) and "info" seems appropriate for a process that's still in progress and might not actually be successful.

PR #1920, opened 2016-04-22
Prevent submit if the form isn't valid. Fixes #1891

### LINKED ISSUE #1891: Pushing "return" on file upload page creates a work
Even if files haven't been uploaded.

The validation javascript should block form submission if no files are uploaded

jcoyne:

### DIFF
--- a/app/assets/javascripts/sufia/save_work/save_work_control.es6
+++ b/app/assets/javascripts/sufia/save_work/save_work_control.es6
@@ -29,6 +29,19 @@ export class SaveWorkControl {
   }
 
   /**
+   * Keep the form from submitting (if the return key is pressed)
+   * unless the form is valid.
+   *
+   * This seems to occur when focus is on one of the visibility buttons
+   */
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/sufia/save_work/save_work_control.es6
```diff
@@ -29,6 +29,19 @@ export class SaveWorkControl {
   }
 
   /**
+   * Keep the form from submitting (if the return key is pressed)
+   * unless the form is valid.
+   *
+   * This seems to occur when focus is on one of the visibility buttons
+   */
+  preventSubmitUnlessValid() {
+    this.form.on('submit', (evt) => {
```
cbeer: Maybe not now, but we should consider using HTML5 form validations to replace some of this custom code.

jcoyne: It is using html5 form validations on certain fields, but not the hidden ones (associated files).

Issue #1906, opened 2016-04-20
catch-all routing for 404 should be appended in generator, not in the engine


barmintor:
### Descriptive summary

Sufia 7 brings down from Sufia 6 this catch all route for 404s that requires the Sufia engine to be last in the application routes. This is annoying when you start running other generators in your app, and makes for surprising behavior changes because it is only added in non-dev, non-test environments. So your app will work until it's deployed, when suddenly things 404 mysteriously.

### Expected behavior

The generator appends the catch all route to the application routes instead of including them in the engine.

### Actual behavior

It includes them in the engine.

### Steps to reproduce the behavior

1. Generate a Sufia app, then localize it with some new routes.

### Related work

None so far.

cbeer:
I'd say, don't include a catch-all route at all. That's the application's job, not sufia's.

barmintor:
I would be happy to escalate to deletion! It just shouldn't be squirreled away in the engine where I can't get to it.

mjgiarlo:
Deleting works for me. I'm pretty sure that was extracted from ScholarSphere. We'll want to include a note about this in the 7.0.0 release notes when we make this change (included in #1751).

Any objections, @jcoyne @awead @cam156 @hectorcorrea?

awead:
Yep. The downstream app might want to handle it differently anyway. 👍

PR #1930, opened 2016-04-27
Mount the Sufia engine as the last route

### LINKED ISSUE #1928: Sufia7 routes not found when using staging environment
We're using Sufia 7.0.0.beta1 with an appropriate set of configuration files for a "staging" environment. We are running solr_wrapper and fcrepo_wrapper on unconventional ports (9983 and 9984) for our Sufia7 testing, both solr and fedora seem to be getting the proper data during file deposit and characterization.

File deposit works fine, but we noticed the thumbnail images were rendering as broken links. The thumbnail images are generated and placed on the filesystem, however, the `downloads/:id?file=thumbnail` was returning a HTTP404. We also noticed that `collections/new` was returning a HTTP404.. 

`rake routes` shows both of these routes in the list.

After an extended period of testing and troubleshooting, we gave up and switched our configurations to use the "development" environment and everything seemed to be working as expected.

jcoyne:

Otherwise, any routes declared after sufia is mounted will get a 404 in production
mode due to the error wildcard route. Fixes #1928

### DIFF
--- a/lib/generators/sufia/install_generator.rb
+++ b/lib/generators/sufia/install_generator.rb
@@ -172,8 +172,8 @@ module Sufia
         "  # This behavior seems to show up only in production mode.\n" \
         "  mount Sufia::Engine => '/'\n"
 
-      sentinel = /devise_for :users/
-      inject_into_file 'config/routes.rb', routing_code, after: sentinel, verbose: false
+      sentinel = /end\Z/
+      inject_into_file 'config/routes.rb', routing_code, before: sentinel, verbose: false
     end
 
     # Add behaviors to the SolrDocument model


cbeer:
Didn't we agree (somewhere) to drop the error routing?

mjgiarlo:
@cbeer Here that is: https://github.com/projecthydra/sufia/issues/1906

Should we shift that back to 7.0.0?

lib/generators/sufia/install_generator.rb
```diff
@@ -172,8 +172,8 @@ def inject_routes
         "  # This behavior seems to show up only in production mode.\n" \
         "  mount Sufia::Engine => '/'\n"
 
-      sentinel = /devise_for :users/
-      inject_into_file 'config/routes.rb', routing_code, after: sentinel, verbose: false
+      sentinel = /end\Z/
```
mjgiarlo: @jcoyne How does this ensure the Sufia engine is mounted on the last line? Won't this match the first `end` statement found? Or does `/end\Z/` match the first `end` statement with no whitespace on either side of it (which, presumably, will be on the last line of the file?

mjgiarlo: This approach will break if folks have a malformatted routes file, say if another gem sticks an `end` statement at the beginning of a line before the end of the file, but meh, I'm not sure there's a better approach.

jcoyne: The `\Z` matches the end of string (effectively EOF). So I'm inserting before the last `end` in the file.

mjgiarlo: :clap: Just verified this in `irb` too.

PR #1932, opened 2016-04-27
Consolidate upload templates

### LINKED ISSUE #1865: Remove file uploader template duplication
Consolidate templates between:
`app/views/sufia/batch_uploads/_form_files.html.erb` and   `app/views/curation_concerns/base/_form_files.html.erb`

Look for other duplication also.

atz:
Fixes #1865

Removes a handful of bugs that had been fixed in one location but not
the other. Beneficial results/side effects:

- Removes extra top level upload start/cancel/delete buttons.
- Adds fixed sticky.
- Adds browse everything.
- Moves javascript templates to common external file.
- Allows flexible selection/ordering of tabs.
- Is open to extension for other per-flow tab content (via `content_for`).
- Removes JS function declaration workaround from JS template context.


app/views/sufia/batch_uploads/_form.html.erb
```diff
@@ -1,31 +1,14 @@
...
+  <% content_for :files_tab do %>
+    <p class="instructions"><strong><%= t("sufia.batch_uploads.files.instructions") %></strong></p>
+    <p class="switch-upload-type">To create a single work for all the files, go to <%= link_to "New Work", new_curation_concerns_generic_work_path %></p>
+  <% end %>
+  <%= render 'curation_concerns/base/guts4form', f: f, tabs: %w[files metadata relationships share] do %>
+  <% end %>
 <% end %>
+
+<script type="text/javascript">
+  $(window).load(function() {
```
jcoyne: What about when turbolinks is used. Will this event would ever be triggered?

atz: I don't have enough understanding of turbolinks to contemplate why it wouldn't.  This is straight jQuery. If this doesn't work, I don't think any jQuery would work.

.travis.yml
```diff
@@ -15,7 +15,7 @@ before_install:
   - "if [ $(phantomjs --version) != '2.1.1' ]; then tar -xvf $PWD/travis_phantomjs/phantomjs-2.1.1-linux-x86_64.tar.bz2 -C $PWD/travis_phantomjs; fi"
   - "phantomjs --version"
 rvm:
-  - 2.2
```
atz: This breaks Travis for reasons unclear:
https://travis-ci.org/projecthydra/sufia/builds/126259182#L531

atz:
Test failure is some contamination from elsewhere in the test suite.  The test is `./spec/views/sufia/batch_uploads/_form.html.erb_spec.rb`, but here's the breakdown:
- passes: run individually
- fails: `be rspec ./spec/views`
- passes!: 
```bash
for x in `ls -ld1 ./spec/views/*`; do be rspec $x ./spec/views/sufia/batch_uploads/_form.html.erb_spec.rb; done`
```

So when running everything in `views`, something contaminates, but when you run every file/subdir in views individually paired with the test in question, no single partner is sufficient to induce failure.  wtf❓❓

atz:
Using `be rspec --bisect spec/views` yields:
```
The minimal reproduction command is:
  rspec ./spec/views/stats/file.html.erb_spec.rb[1:1:2:1] ./spec/views/sufia/batch_uploads/_form.html.erb_spec.rb[1:1]
```

atz:
Any idea why anything in this spec would cause side effects later?
https://github.com/projecthydra/sufia/blob/master/spec/views/stats/file.html.erb_spec.rb

Looks reasonably conventional to me.

atz:
Rebased.  The new test failures are seemingly introduced by dependency shift in mainline.

PR #1934, opened 2016-04-28
Deposit agreement for works with files

### LINKED ISSUE #1870: Editing a work and the Deposit Agreement
### Descriptive summary

When users edit a Work they shouldn't have to acknowledge the Deposit Agreement again unless they are adding a new file. In the current version, even changing the title of a Work or any of the metadata fields forces the user to check the "I have read and do agree to the Deposit Agreement" checkbox. 

### Expected behavior
When editing a Work, only force users to check the Deposit Agreement if files are added to the work.

### Actual behavior
Currently users must check the Deposit Agreement on any change to the work.

1. Click Edit on a Work
1. Change any field (e.g. the title)
1. Try to click Save, you'll see the button is disabled until you agree to the Deposit Agreement.

hectorcorrea:
Fixes #1870 

Defaults the deposit agreement checkbox to checked when editing a work that has already been saved.

The change in this PR is that the `work_form.rb` now initializes the value of the agreement checkbox to checked/unchecked depending on whether the Work already has already been saved (we assume that on a saved work the user has already accepted the deposit agreement)

~~There is also some temporary code (see the `.es6` files) to highlight where we could update the JavaScript of the page to have a more sophisticated logic and force the user to re-accept the deposit agreement if they add new Files to the Work (but don't bother them if they only edit the metadata of the work)~~ (I'll push the JavaScript enhancements in a separate PR to keep things cleaner)

This is how the Work Edit looks when editing a work that already has Files associated with it.

screen shot 2016-04-28 at 7 21 30 am

### DIFF
--- a/app/forms/sufia/forms/work_form.rb
+++ b/app/forms/sufia/forms/work_form.rb
@@ -2,7 +2,14 @@ module Sufia::Forms
   class WorkForm < CurationConcerns::Forms::WorkForm
     delegate :depositor, :on_behalf_of, :permissions, to: :model
 
+    attr_reader :agreement_accepted
+
     self.terms += [:collection_ids]
+
+    def initialize(model, current_ability)
+      @agreement_accepted = !model.new_record?
+      super
+    end
 
     def [](key)
       return [] if key == :collection_ids
--- a/app/views/curation_concerns/base/_form_progress.html.erb
+++ b/app/views/curation_concerns/base/_form_progress.html.erb
@@ -22,7 +22,7 @@
     <div class="panel-footer text-center">
       <% if Sufia::Engine.config.active_deposit_agreement_acceptance %>
           <label>
-            <%= check_box_tag 'agreement', 1, nil, required: true %>
+            <%= check_box_tag 'agreement', 1, f.object.agreement_accepted, required: true %>
             <%= t('sufia.active_consent_to_agreement') %><br>
             <%= link_to t('sufia.deposit_agreement'),
                         sufia.static_path('agreement'),
--- a/spec/views/curation_concerns/base/_form_progress.html.erb_spec.rb
+++ b/spec/views/curation_concerns/base/_form_progress.html.erb_spec.rb
@@ -10,7 +10,6 @@ describe 'curation_concerns/base/_form_progress.html.erb' do
   before do
     view.lookup_context.view_paths.push 'app/views/curation_concerns'
     allow(controller).to receive(:current_user).and_return(user)
-    assign(:form, form)
   end
 
   let(:page) do
@@ -21,6 +20,8 @@ describe 'curation_concerns/base/_form_progress.html.erb' do
   end
 
   context "for a new object" do
+    before { assign(:form, form) }
+
     let(:work) { GenericWork.new }
 
     context "with options for proxy" do
@@ -50,6 +51,7 @@ describe 'curation_concerns/base/_form_progress.html.erb' do
       it "shows accept text" do
         expect(page).to have_content 'I have read and agree to the'
         expect(page).to have_link 'Deposit Agreement', href: '/agreement'
+        expect(page).to_not have_selector("#agreement[checked]")
       end
     end
 
@@ -62,6 +64,19 @@ describe 'curation_concerns/base/_form_progress.html.erb' do
         expect(page).to have_content 'By saving this work I agree to the'
         expect(page).to have_link 'Deposit Agreement', href: '/agreement'
       end
+    end
+  end
+
+  context "when the work has been saved before" do
+    before do
+      allow(work).to receive(:new_record?).and_return(false)
+      assign(:form, form)
+    end
+
+    let(:work) { stub_model(GenericWork, id: '456') }
+
+    it "renders the deposit agreement already checked" do
+      expect(page).to have_selector("#agreement[checked]")
     end
   end
 end
--- a/spec/forms/curation_concerns/work_form_spec.rb
+++ b/spec/forms/curation_concerns/work_form_spec.rb
@@ -1,7 +1,8 @@
 require 'spec_helper'
 
 describe CurationConcerns::GenericWorkForm do
-  let(:form) { described_class.new(GenericWork.new, nil) }
+  let(:work) { GenericWork.new }
+  let(:form) { described_class.new(work, nil) }
 
   describe "#primary_terms" do
     subject { form.primary_terms }
@@ -76,5 +77,17 @@ describe CurationConcerns::GenericWorkForm do
   describe "on_behalf_of" do
     subject { form.on_behalf_of }
     it { is_expected.to be nil }
+  end
+
+  describe "#agreement_accepted" do
+    subject { form.agreement_accepted }
+    it { is_expected.to eq false }
+  end
+
+  context "on a work already saved" do
+    before { allow(work).to receive(:new_record?).and_return(false) }
+    it "defaults deposit agreement to true" do
+      expect(form.agreement_accepted).to eq(true)
+    end
   end
 end


app/assets/javascripts/sufia/save_work/uploaded_files.es6
```diff
@@ -2,11 +2,16 @@ export class UploadedFiles {
   // Monitors the form and runs the callback if any files are added
   constructor(form, callback) {
     this.form = form
+    this.startedChecked = this.form.find('input[name="agreement"]').val() === "1"
```
jcoyne: Why are these changes in uploaded_files.es6 instead of deposit_agreement.es6?

app/views/curation_concerns/base/_form.html.erb
```diff
@@ -34,7 +34,7 @@
       </div>
 
       <div id="savewidget" class="col-xs-12 col-sm-4 fixedsticky" role="complementary">
-        <%= render 'form_progress', f: f %>
+        <%= render 'form_progress', f: f, agreement_accepted: @form.agreement_accepted %>
```
jcoyne: `#agreement_accepted` is not a model property, so I'm surprised this isn't throwing an error. If it were a model property you wouldn't need to pass it as a local. You could access it via `f.object.agreement_accepted`

hectorcorrea: Do we want to make `#agreement_accepted` a model property? At this point I am calculating the value when the WorkForm is instantiated.

app/forms/sufia/forms/work_form.rb
```diff
@@ -4,6 +4,12 @@ class WorkForm < CurationConcerns::Forms::WorkForm
 
     self.terms += [:collection_ids]
 
+    def initialize(model, current_ability)
+      # Default agreement to accepted if users have already uploaded files to the work
+      @agreement_accepted = model.file_sets.count > 0
```
jcoyne: I assume you need an accessor for `agreement_accepted`. This seems like it might be a very expensive way (lots of fedora hits) to determine if the agreement is accepted. Might we just assume that if model is persisted the agreement was accepted?

app/assets/javascripts/sufia/save_work/save_work_control.es6
```diff
@@ -95,6 +95,19 @@ export class SaveWorkControl {
...
       this.requiredFiles.check()
+
+      if (this.uploads.hasFiles) {
+        // TODO: A flag to track that the user has agreed during this session should
+        //       be set when the user clicks on the deposit agreement checkbox.
+        let agreedRealtime = true
+        if (agreedRealtime) {
+          return true
+        }
+        if (this.uploads.agreedLasttime) {
```
jcoyne: This code will never be run, as agreedRealtime is always true.

app/forms/sufia/forms/work_form.rb
hectorcorrea: @jcoyne this is where I am setting the value of the `#agreement_accepted` property.

app/assets/javascripts/sufia/save_work/uploaded_files.es6
hectorcorrea: deposit_agreement.es6 would probably be a better place. Thanks!

app/assets/javascripts/sufia/save_work/save_work_control.es6
hectorcorrea: Correct, that's why I have the TODO in the comments. I've been manually setting this value as I run the code to test the logic.

app/forms/sufia/forms/work_form.rb
```diff
@@ -2,8 +2,15 @@ module Sufia::Forms
   class WorkForm < CurationConcerns::Forms::WorkForm
     delegate :depositor, :on_behalf_of, :permissions, to: :model
 
+    attr_reader :agreement_accepted
+
     self.terms += [:collection_ids]
 
+    def initialize(model, current_ability)
+      @agreement_accepted = !model.new_record?
```
hectorcorrea: @jcoyne as you suggested I changed the code to use `model.new_record?` (instead of model.file_sets.count > 1). This should be pretty fast, right?

app/views/curation_concerns/base/_form_progress.html.erb
```diff
@@ -22,7 +22,7 @@
     <div class="panel-footer text-center">
       <% if Sufia::Engine.config.active_deposit_agreement_acceptance %>
           <label>
-            <%= check_box_tag 'agreement', 1, nil, required: true %>
+            <%= check_box_tag 'agreement', 1, defined?(agreement_accepted) && agreement_accepted, required: true %>
```
hectorcorrea: The `defined?(agreement_accepted)` check is needed because the `_form_progress` is rendered from more than one place. Instead of this check I could pass the `agreement_accepted` from all places where the form is rendered.

app/forms/sufia/forms/work_form.rb
jcoyne: Yep. :+1:

app/views/curation_concerns/base/_form.html.erb
jcoyne: Now that you've made it an accessor on the form it shouldn't be a problem. However, there's still no need to pass it as a separate local.

hectorcorrea: Fixed, thanks!

spec/views/curation_concerns/base/_form_progress.html.erb_spec.rb
```diff
@@ -64,4 +65,16 @@
       end
     end
   end
+
+  context "when the work has been saved before" do
+    let(:work) do
+      gw = GenericWork.new(creator: ["somebody"], title: ["abc"])
```
jcoyne: Can we do this with stubbing `new_record?`. Saving to fedora is slow.

hectorcorrea: I tried using a `before { allow(work).to receive(:new_record?).and_return(false) }` like I did on the other test but that did not work. RSpec reported that it could not find a route when the id was null or something like that.

jcoyne: Have you tried using `stub_model`?

spec/views/curation_concerns/base/_form_progress.html.erb_spec.rb
```diff
@@ -64,4 +66,17 @@
...
     end
   end
+
+  context "when the work has been saved before" do
+    before do
+      allow(work).to receive(:new_record?).and_return(false)
+      assign(:form, form)
+    end
+
+    let(:work) { stub_model(GenericWork, id: '456') }
```
hectorcorrea: @jcoyne Updated to use `stub_model(GenericWork)`. This was a bit tricky since I had to also move the `assign(:form, form)` inside each different `context` for it to use the different works.

jcoyne: You shouldn't have had to move the assign.  It works from the inside out when it's resolving variables (`work` in this case).

hectorcorrea: That's what I thought. But it does not matter what I do, it seems to be resolving `work` out of wack and picking up the wrong setting for `form.agreement_accepted`. It's probably because the setting is in the `form` (which depends on the `work`) and we are stubbing the `work` itself.

jcoyne:
Good job @hectorcorrea. I think this is ready to merge when CI passes. :+1:

PR #1938, opened 2016-04-28
Events should fire for Works

### LINKED ISSUE #1437: Do Events go on FileSets or Works?
To close this ticket we should investigate what events are currently being created, where they are being created and then ticket the differences from the following specification. There are six content-related event jobs at the moment. Here's how they should break down between FileSets and Works:

1. ContentDepositEvent: Work & FileSet
1. ContentUpdateEvent: Work & FileSet
1. ContentDeleteEvent: Work & FileSet
1. ContentDepositorChangeEvent: Work
1. ContentNewVersionEvent: FileSet, since we're only versioning content at this point
1. ContentRestoredVersionEvent: FileSet, since we're only versioning content at this point

mjgiarlo:
Fixes #1437.

### DIFF
--- a/app/jobs/content_delete_event_job.rb
+++ b/app/jobs/content_delete_event_job.rb
@@ -1,4 +1,4 @@
-# A specific job to log a file deletion to a user's activity stream
+# Log object deletion to activity streams
 #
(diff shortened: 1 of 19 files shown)


jcoyne:
0.14.0.pre3 has been released

PR #1940, opened 2016-04-29
Bump curation concerns to 0.14.0.pre3

atz:


atz:
More bogus CI failures...

PR #1941, opened 2016-04-29
Forces users to agree to deposit agreement again only if users add files

### LINKED ISSUE #1870: Editing a work and the Deposit Agreement

hectorcorrea:
Fixes #1870 (part II)

When a user edits a Work that already has files the Deposit Agreement checkbox is now checked initially (see PR #1934) since users must have agreed when they first submitted the Work. This PR forces the users to agree again when they add a new file to the Work, but not if they only change the Work metadata.

### DIFF
--- a/app/assets/javascripts/sufia/save_work/save_work_control.es6
+++ b/app/assets/javascripts/sufia/save_work/save_work_control.es6
@@ -78,7 +78,8 @@ export class SaveWorkControl {
     // avoid short circuit evaluation. The checkboxes should be independent.
     let metadataValid = this.validateMetadata()
     let filesValid = this.validateFiles()
-    return metadataValid && filesValid && this.depositAgreement.isAccepted
+    let agreementValid = this.validateAgreement(filesValid)
+    return metadataValid && filesValid && agreementValid
   }
 
   // sets the metadata indicator to complete/incomplete
@@ -99,6 +100,15 @@ export class SaveWorkControl {
     }
     this.requiredFiles.uncheck()
     return false
+  }
+
+  validateAgreement(filesValid) {
+    if (filesValid && this.uploads.hasFiles && this.depositAgreement.mustAgreeAgain) {
+      // Force the user to agree again
+      this.depositAgreement.setNotAccepted()
+      return false
+    }
+    return this.depositAgreement.isAccepted
   }
 }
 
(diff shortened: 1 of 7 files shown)


hectorcorrea:
@jcoyne I still need to add tests, but this PR is what I am thinking for the JavaScript part of the deposit agreement validation. I left the code in the `save_work_control.es6 `since it needs to interact with a couple of components on the page (the upload control and the checkbox)

app/assets/javascripts/sufia/save_work/save_work_control.es6
```diff
@@ -95,6 +96,13 @@ export class SaveWorkControl {
   validateFiles() {
     if (!this.isNew || this.uploads.hasFiles) {
       this.requiredFiles.check()
+
+      if (this.uploads.hasFiles && this.mustAgreeAgain) {
```
jcoyne: I don't think this code belongs in this method because the code you added is not validating that they have files on the page.

jcoyne: Can you move it to `isValid()` instead?

app/assets/javascripts/sufia/save_work/save_work_control.es6
```diff
@@ -61,6 +61,7 @@ export class SaveWorkControl {
 
     this.saveButton = this.element.find(':submit')
 
+    this.mustAgreeAgain = this.form.find('input[name="agreement"]').val() === "1"
```
jcoyne: Is this the same as `this.depositAgreement.isAccepted()`?

app/assets/javascripts/sufia/save_work/save_work_control.es6
```diff
@@ -95,6 +96,13 @@ export class SaveWorkControl {
   validateFiles() {
     if (!this.isNew || this.uploads.hasFiles) {
       this.requiredFiles.check()
+
+      if (this.uploads.hasFiles && this.mustAgreeAgain) {
+        // Force the user to agree before submitting
+        this.form.find('input[name="agreement"]').prop("checked", false)
```
jcoyne: Can you deal with this at a higher level? `this.depositAgreement.setNotAccepted` perhaps?

jcoyne:
In light of https://github.com/projecthydra/sufia/issues/1943, keep in mind that the form having files is not the same as the form having new files added.  Let's work to create that distinction here.

hectorcorrea:
I will submit this PR again with `master` as the target branch.

Issue #1947, opened 2016-04-29
Creator displays in brackets on index page


jcoyne:
<img width="801" alt="screen shot 2016-04-29 at 3 39 21 pm" src="https://cloud.githubusercontent.com/assets/92044/14928954/9b3d4864-0e20-11e6-9135-9e3c2cd2b54d.png">

PR #1950, opened 2016-04-30
Prevent title from displaying twice

### LINKED ISSUE #1948: Title should not display twice in index view
screen shot 2016-04-29 at 3 39 21 pm

Since it's the header, no need to display it in the table too.

atz:
fixes #1948

### DIFF
--- a/app/views/shared/_attributes.html.erb
+++ b/app/views/shared/_attributes.html.erb
@@ -1,8 +1,11 @@
-      <table class="table">
+<% include_title ||= false # title is part of the exterior template in common cases %>
+    <table class="table">
+      <% if include_title %>
         <tr>
           <th><span class="attribute-label h4">Title:</span></th>
           <td><%= work.title %></td>
         </tr>
+      <% end %>
         <tr>
           <th><span class="attribute-label h4">Owner:</span></th>
           <td><%= link_to_profile work.depositor("no depositor value") %></td>
@@ -33,5 +36,4 @@
           <th><span class="attribute-label h4">Date Uploaded:</span></th>
           <td><%= work.date_uploaded %></td>
         </tr>
-      </table>
-
+    </table>


app/views/shared/_attributes.html.erb
```diff
@@ -1,8 +1,11 @@
-      <table class="table">
+<% include_title ||= false # title is part of the exterior template in common cases %>
```
jcoyne: Do we need this conditional? This basically leaves us with a block of uncovered code.

atz:
Not sure how to handle the Gallery template since it comes straight from blacklight:
`.../gems/blacklight-6.1.0/app/views/catalog/_index_default.html.erb`

Options are:
- Use "title" field from that BL template (override `/gems/blacklight-gallery-0.5.0/app/views/catalog/_index_gallery.html.erb` to exclude `_index_header_default`)
- Alter or locally revise the returns from `index_fields(document)`
- Override `should_render_index_field?(document, field)` 
- ???

app/views/shared/_attributes.html.erb
atz: Strictly speaking, no.  But if we (or our consumers) want to reuse the template in a display that includes the title as just the first of the set of attributes, then this conditional saves them from having to fork and override the whole template.  In general we need to move towards more configurable displays, ideally with the fields to be displayed enumerated outside of the templates themselves (similar to the way blacklight itself does it).

It doesn't seem to be the type of coverage that coveralls cares about, but you are correct that nothing is testing whether this conditional is honored.  Willing to add a test.  Also willing to defer.

jcoyne: My philosophy is not to build software that nobody is asking for. All code has bugs and thus the less code we build the higher quality product we have. Doubly so for untested code.

atz: Looks like this is superceded in the (ostensibly unrelated) PR #1947.  I support the approach there since it does what I projected was necessary.

atz: I feel like the ticketing set us up here.  These should have been the same ticket.

mjgiarlo:
@atz Do you think we can close this PR, or should we first verify that it has the effect re: titles that we expect?

atz:
Confirmed gallery view also follows display logic that Justin implemented.

PR #1953, opened 2016-05-02
Admin users should be editors

awead:
Found a bug where admin users are not considered editors. This fixes the problem by making `editor?` return true is the user is an admin.

### DIFF
--- a/app/models/concerns/sufia/ability.rb
+++ b/app/models/concerns/sufia/ability.rb
@@ -52,6 +52,7 @@ module Sufia
 
       can :create, TinymceAsset
       can [:create, :update], ContentBlock
+      can :edit, ::SolrDocument
     end
 
     def stats_abilities
--- a/spec/presenters/sufia/work_show_presenter_spec.rb
+++ b/spec/presenters/sufia/work_show_presenter_spec.rb
@@ -32,28 +32,32 @@ describe Sufia::WorkShowPresenter do
     end
   end
 
-  describe 'Featured Works for admin users' do
-    let(:user) { create(:user) }
-    before { allow(user).to receive_messages(groups: ['admin', 'registered']) }
+  describe 'admin users' do
+    let(:user)    { create(:user) }
     let(:ability) { Ability.new(user) }
-    let!(:work) { build(:public_generic_work) }
+    let!(:work)   { build(:public_generic_work) }
 
-    context 'on a brand new public work' do
-      it 'allows user to feature the work' do
+    before { allow(user).to receive_messages(groups: ['admin', 'registered']) }
+
+    context 'with a new public work' do
+      it 'can feature the work' do
         allow(user).to receive(:can?).with(:create, FeaturedWork).and_return(true)
         expect(presenter.display_feature_link?).to be true
         expect(presenter.display_unfeature_link?).to be false
       end
     end
 
-    context 'on an already featured work' do
-      before do
-        FeaturedWork.create(generic_work_id: work.id)
-      end
-      it 'allows user to unfeature the work' do
+    context 'with a featured work' do
+      before { FeaturedWork.create(generic_work_id: work.id) }
+      it 'can unfeature the work' do
         expect(presenter.display_feature_link?).to be false
         expect(presenter.display_unfeature_link?).to be true
       end
+    end
+
+    describe "#editor?" do
+      subject { presenter.editor? }
+      it { is_expected.to be true }
     end
   end
 end


mjgiarlo:
Hmm, why not modify the `Ability` class here? If admins lack access to editorial functions in the work presenter, it stands to reason that they may also lack abilities elsewhere in the app. Perhaps something like:

```ruby
def admin_abilities
  return unless admin?
  can :manage, :all
end
```

awead:
@mjgiarlo I'm guessing `:all` really does mean everything? even `solr_document?

awead:
@mjgiarlo and this ought to go under CurationConcerns.

mjgiarlo:
@awead I do believe that `:all` is everything. Agree that this makes sense to go in CC! Good call.

cbeer:
👎  I finally removed `manage :all` from curation_concerns. An engine should only grant abilities provided by that engine.

jcoyne:
I'm not even sure why we need `editor?`.  Some users can edit some of the things, but not all the things.  That's why it makes sense to do `can? :edit, thing`

mjgiarlo:
@jcoyne The `editor?` method in this presenter is a check to show/hide delete and edit functionality in the work show page. Perhaps instead of directly altering the presenter, given @cbeer's point above, we should add a new ability for admins to edit solr documents?

jcoyne:
@mjgiarlo thanks for the explanation, I was thinking that was in ability, not in a presenter.

awead:
@mjgiarlo something odd is going on. When I add: `can :edit, SolrDocument` to the mix:
``` ruby
(byebug) current_ability.admin?
true
(byebug) current_ability.can?(:edit, solr_document)
false
(byebug) current_ability.can?(:edit, SolrDocument)
true
(byebug) solr_document.class
SolrDocument
```
I don't know what CanCanCan is doing under the hood to get from `solr_document`.

mjgiarlo:
@awead could hydra-access-controls or blacklight-access-controls be making the determination there?

jcoyne:
@awead https://github.com/projecthydra/hydra-head/blob/master/hydra-access-controls/lib/hydra/ability.rb#L49-L52

app/models/concerns/sufia/ability.rb
```diff
@@ -52,6 +52,7 @@ def editor_abilities
 
       can :create, TinymceAsset
       can [:create, :update], ContentBlock
+      can :edit, ::SolrDocument
```
jcoyne: I'm still unsure why we need this at all.  If we want the object to be editable by admins, shouldn't we add the admin group to the object's rights metadata?

awead: @jcoyne yeah, I've used that approach before, however, I think I prefer this. You don't have to maintain another group, i.e. visibility's "registered" or "public", and that's one less Fedora resource created per object. Plus, if the object looses that admin group somehow, you wouldn't be able to edit it.

OTOH, using rights metadata would also push the admin ability to the Fedora level, and enable it there once the Fedora ACL/Hydra ACL alignment is complete... (if/when that happens)

awead: Personally, I'd say remove lines 53-55 and just have `can :manage, :all` and be done with it. The implementor can override the method in their own `ability.rb`.

jcoyne: I think we're trying to get away from having to do overrides.  I'm strongly advocating putting this in the rights metadata. Someone should be able to understand who has access even if Hydra goes away

awead: @jcoyne so, a group named "admin" would have to go into the `edit_groups` of every object created in Hydra?

jcoyne: If the admins should be able to edit it, then yes.  I think it is odd that we have a single admin group. Many places have multiple levels of administrators or admin sets, with different administrators per set.

cbeer: I'm still 👎  having an engine provide `can :manage, :all`; it's obnoxious for an application to undo, and requires overriders to closely track implicit ability changes. I'm fine with `:manage, [list of models we provide/care about]`, though.

Having an explicit admin role in the rights metadata seems appropriate, although I wonder if it'll get us in trouble elsewhere (e.g. needing to disallow other managers from removing the admin group). Or, maybe there's something more subtle going on (e.g. "application admins should be able to manage any object created through the application").

awead: @jcoyne ok. I see where you're going, but I thinking we're creeping out of scope. I agree that we will need a finer-grained idea of "administrator" but I'm interpreting the present scenario as a local root user. In other words, this is a user specific to the application, not to Fedora, that just needs to do anything. Even with multiple levels of administrators, aren't you still going to need a super user whose rights trump all?

atz: The argument against putting admin rights in metadata is imagining you have multiple (N) hydra heads on the same repository.  Distinct domains.  Each head would have its own admins.  If each head is not aware of all others permissions expectations (a maintenance nightmare), then you end up having to have all heads listen for all potentially relevant events and serially inject their own rights strings, versioning your object N-1 additional times.  Massive performance cost and logical complexity. 

An app tracking and constraining its own admin rights is a favor to the rest of the ecosystem.  The rights MD should represent object rights on a timescale beyond the lifespan of an application.

mjgiarlo: After reading the thread, I'm OK with this `can :edit, ::SolrDocument` solution for now. Because we're using `::SolrDocument` and not `:all`, this steers clear of what @cbeer was warning us against. Inclined to merge this work as soon as @awead can get Travis :green_apple:.

jcoyne: can :edit, SolrDocument is only slightly better than :manage :all. It gives the admin full access to any AF::Base object

PR #1954, opened 2016-05-02
Fields displayed on search results come from config

### LINKED ISSUE #1947: Creator displays in brackets on index page

jcoyne:

Instead of requiring implementers to override a view template to change
the values that display on the search results page, use the Blacklight
configuration as it was intended. This will pave the way for dynamic
configuration features in the future.

There are also a few changes here that make the views more agnostic
about the model type by moving generic_work partials default partials

Fixes #1947


app/views/shared/_attributes.html.erb
```diff
@@ -1,37 +0,0 @@
...
-          <td><%= work.title %></td>
-        </tr>
-        <tr>
-          <th><span class="attribute-label h4">Owner:</span></th>
-          <td><%= link_to_profile work.depositor("no depositor value") %></td>
-        </tr>
-        <tr>
-          <th><span class="attribute-label h4">Depositor:</span></th>
-          <td>
-            <% if work.proxy_depositor  %>
```
mjgiarlo: Where is this logic in the new stuff?

jcoyne: Presently it's just displaying proxy_depositor as Depositor if it exists.  If it doesn't exist, it only displays the Owner field. I think that makes sense. Why do we want to see the depositor if it's the same as the owner?

mjgiarlo: For consistency, but I'm not that attached to this behavior.

PR #1957, opened 2016-05-02
Title, Keyword, Creator and Rights are required fields

### LINKED ISSUE #1942: Metadata does not display on Work show page
### Descriptive summary

Sufia 7.0.0.beta1; Ruby 2.2.0

Setting metadata fields for a New Work does not display metadata on work show page after saving.

### Expected behavior

Setting values for metadata fields of a new work should display those fields and their values on the work show page in the Metadata section.

### Actual behavior

When setting values for metadata fields of a new work and saving, the work show page has "No metadata is defined for this work." in the Metadata section.

1. Works -> New Work
1. Metadata Tab:  Set Title
1. Metadata Tab: View additional fields and set Language = English
1. File Tab:  Add one file
1. Save
1. NOTICE: Title is correct
1. NOTICE: Metadata section should show Language: English, but actually shows "No metadata is defined for this work."  (I reproduced this with Tag, Publisher, and several other fields)
1. Edit
1. Metadata Tab: View additional fields and NOTICE Language does equal English

### Related work

#1943, #1944

jcoyne:

Fixes #1942

### DIFF
--- a/app/assets/javascripts/sufia/save_work/required_fields.es6
+++ b/app/assets/javascripts/sufia/save_work/required_fields.es6
@@ -2,7 +2,8 @@ export class RequiredFields {
   // Monitors the form and runs the callback if any of the required fields change
   constructor(form, callback) {
     this.form = form
-    this.requiredFields = this.form.find('input[required]')
+    // ":input" matches all input, select or textarea fields.
+    this.requiredFields = this.form.find(':input[required]')
     this.requiredFields.change(callback)
   }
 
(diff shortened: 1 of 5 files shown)


awead:
@jcoyne there are also validations on the model for title. Can we put all the validations in one place?

jcoyne:
@awead I don't think validations necessarily belong on the model. They are a property of the form. You may want one path, say a bulk upload that only validates that a title is present, but on the self deposit form you want to ensure they have a bit more than that.

awead:
@jcoyne ok. then we should remove this line from GenericWork:
`validates :title, presence: { message: 'Your work must have a title.' }`

jcoyne:
@awead I don't think so. The modeling folks said many times that all objects must have a label or title.

Issue #1949, opened 2016-04-29
Verify we're not broken with turbolinks turned on


mjgiarlo:
See: https://github.com/projecthydra/sufia/pull/1932#discussion_r61469566

Test that in particular to see if turbolinks causes problems. Also, verify whether this step is still accurate and needed:

https://github.com/projecthydra/sufia/wiki/Sufia-Management-Guide#user-interface

jcoyne:
Why is this tagged "needs feedback"?  Who are we waiting on to provide it?

mjgiarlo:
Removed that label.

jcoyne:
I have confirmed that the issue I raised on #1932 is in fact broken with turbolinks enabled.  You can see in this screenshot that an incorrect template for batch upload has been used (no field for label or resource type):

<img width="1248" alt="screen shot 2016-05-02 at 7 24 35 am" src="https://cloud.githubusercontent.com/assets/92044/14965418/f3203dee-1072-11e6-98fa-a39e4434e2ac.png">

jcoyne:
TinyMCE (for the admin), is also broken with Turbolinks.

jcoyne:
@mjgiarlo close this?

PR #1959, opened 2016-05-02
Always set the correct upload template on page load

jcoyne:

Ref #1949


atz:
That's the default, so setting it is unnecessary.

jcoyne:
@atz but if you visit the batch page first, it gets changed. Then when you return to the default page, it uses the batch template.

PR #1961, opened 2016-05-02
Update to Curation_concerns 0.14.0.pre4

jcoyne:

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -79,8 +79,6 @@ else
   end
 end
 
-Resque.inline = Rails.env.test?
-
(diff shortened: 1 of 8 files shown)


spec/spec_helper.rb
```diff
@@ -79,8 +79,6 @@ def loaddb
   end
 end
 
-Resque.inline = Rails.env.test?
```
atz: Was this just leftover cruft?

jcoyne: It causes an error if you don't have Resque (which is no longer required by anything else in the stack)

PR #1965, opened 2016-05-03
Add ability to upload folders

mjgiarlo:
Only works in Google Chrome, so don't show the button in other browsers.

This was supported in Sufia 1-6, so we should support it in Sufia 7 as well.

### DIFF
--- a/app/views/curation_concerns/base/_form_files.html.erb
+++ b/app/views/curation_concerns/base/_form_files.html.erb
@@ -6,12 +6,20 @@
             <div class="col-xs-7">
                 <!-- The fileinput-button span is used to style the file input field as button -->
                 <span class="btn btn-success fileinput-button">
-                    <i class="glyphicon glyphicon-plus"></i>
+                    <span class="glyphicon glyphicon-plus"></span>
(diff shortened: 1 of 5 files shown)


jcoyne:
Should this have a test?

mjgiarlo:
@jcoyne Given the intermittent test failures we were seeing w/ the feature spec where it'd make sense to test this, what do you think is a reasonable test for this change?

app/views/curation_concerns/base/_form_files.html.erb
```diff
@@ -10,6 +10,14 @@
                     <span>Add files...</span>
                     <input type="file" name="files[]" multiple>
                 </span>
+                <% if !!(request.env['HTTP_USER_AGENT'] =~ /Chrome/) %>
```
cbeer: Could we extract this condition to a helper? Either as e.g. `chrome?`, or maybe `supports_folder_upload?`, or some javascript-side progressive enhancement would be appropriate?

mjgiarlo:
@hectorcorrea++

PR #1966, opened 2016-05-04
Bring consistency to the Dashboard

mjgiarlo:
* Display number of works instead of files, since works are now the center of attention.
* Allow `#link_to_field` helper to take options, and rename its args to be less redundant.
* Align display of followers/following with collections & works created blocks above it.
* Link to works created should show only works, and not collections.
* Collections created should link to the catalog like works created does.
* Add `human_readable_type` as a facetable field.

### DIFF
--- a/app/views/users/_following_modal.html.erb
+++ b/app/views/users/_following_modal.html.erb
@@ -1,9 +1,8 @@
+<!-- Text to trigger modal -->
+<span class="badge"><%= following.count %></span>
+<span class="glyphicon glyphicon-hand-right"></span> <a data-toggle="modal" href="#following">Following</a>
+
 <div class="modal-div">
-
-  <!-- Text to trigger modal -->
-  <i class="glyphicon glyphicon-hand-right"></i> <a data-toggle="modal" href="#following">Following:</a> <span class="badge"><%= following.count %></span>
-
-  <!-- Modal -->
   <!-- Modal -->
   <div class="modal fade" id="following" tabindex="-1" role="dialog"  aria-hidden="true">
     <div class="modal-dialog">
--- a/app/views/users/_vitals.html.erb
+++ b/app/views/users/_vitals.html.erb
@@ -1,22 +1,21 @@
 <div class="list-group-item">
-  <i class="glyphicon glyphicon-time"></i> Joined on <%= user.created_at.to_date.strftime("%b %d, %Y") %>
+  <span class="glyphicon glyphicon-time"></span> Joined on <%= user.created_at.to_date.strftime("%b %d, %Y") %>
 </div>
 
 <div class="list-group-item">
-  <span class="badge"><%= number_of_collections %></span>
-  <i class="glyphicon glyphicon-folder-open"></i> <%= t("sufia.dashboard.stats.collections") %>
+  <span class="badge"><%= number_of_collections(user) %></span>
+  <span class="glyphicon glyphicon-folder-open"></span> <%= link_to_field('depositor', user.to_s, t("sufia.dashboard.stats.collections"), human_readable_type: "Collection") %>
 </div>
-  
-<div class="list-group-item">
-  <span class="badge"><%= number_of_files(user) %></span>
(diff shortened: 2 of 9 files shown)


mjgiarlo:
Here's how the dashboard used to look:

dashboard_vitals_old

Here's how it looks with this PR:

dashboard_vitals_new

mjgiarlo:
Old catalog view (linked from "Works created" which used to be "Deposited Files"):

files_created_old

New catalog view (linked from "Works created"):

works_created_new

app/views/users/_following_modal.html.erb
```diff
@@ -1,9 +1,8 @@
-<div class="modal-div">
-
-  <!-- Text to trigger modal -->
-  <i class="glyphicon glyphicon-hand-right"></i> <a data-toggle="modal" href="#following">Following:</a> <span class="badge"><%= following.count %></span>
+<!-- Text to trigger modal -->
+<span class="badge"><%= following.count %></span>
+<i class="glyphicon glyphicon-hand-right"></i> <a data-toggle="modal" href="#following">Following</a>
```
mtribone: Bootstrap 3 uses the `<span>` element instead of the `<i>` element. Small, picky thing.

app/views/users/_vitals.html.erb
```diff
@@ -4,21 +4,20 @@
...
   <span class="badge"><%= number_of_collections %></span>
-  <i class="glyphicon glyphicon-folder-open"></i> <%= t("sufia.dashboard.stats.collections") %>
+  <i class="glyphicon glyphicon-folder-open"></i> <%= link_to_field('depositor', user.to_s, t("sufia.dashboard.stats.collections"), 'f[human_readable_type_sim][]': "Collection") %>
 </div>
-  
+
 <div class="list-group-item">
-  <span class="badge"><%= number_of_files(user) %></span>
-    <i class="glyphicon glyphicon-upload"></i> <%= link_to_field('depositor', user.to_s, 'Deposited Files') %>
+  <span class="badge"><%= number_of_works %></span>
```
jcoyne: I don't think you want to works of the current user. You want the number of works for the user you are looking at on this page.

app/views/users/_vitals.html.erb
```diff
@@ -3,22 +3,21 @@
 </div>
 
 <div class="list-group-item">
```
mtribone: Not 100% sure, but I believe that list-group-items need to be part of a list group.

mjgiarlo:
Tagging @projecthydra/sufia-ui-ux-advisors re: :arrow_heading_up:

app/views/users/_vitals.html.erb
mjgiarlo: There is a containing `list-group` div in the containing partial, I believe.

newmanld:
+1 to changes to dashboard.  Works instead of files, and better alignment, indeed.  Thanks!

On this last suggested bullet point ' Add human_readable_type as a facetable field.' - is this what is put into effect on the New Catalog view shown above, under 'Type of Work'? - The example just lists 'Works' as the facet - what will determine the values in  'human_readable_type'?

simosacchi:
+1 on the changes: they bring consistency in terminology, functions, and presentation.

mjgiarlo:
@newmanld :speech_balloon: 
> On this last suggested bullet point ' Add human_readable_type as a facetable field.' - is this what is put into effect on the New Catalog view shown above, under 'Type of Work'? 

Yep.

> The example just lists 'Works' as the facet - what will determine the values in  'human_readable_type'?

`human_readable_type` is a property of all CurationConcern-based models. So the default value is the name of the model capitalized. You can override it, like we do with GenericWorks in Sufia, which you can see here:

https://github.com/projecthydra/sufia/blob/7e65c9f2f03461bb3ab2c16ea30d47e64811bc9a/lib/generators/sufia/install_generator.rb#L90

newmanld:
Thanks for the answer @mjgiarlo.  +1 to that change as well!

PR #1975, opened 2016-05-04
add visibility to collections in Sufia

elrayle:
Fixes #1974

Adds new feature visibility for collections.  Follows the same pattern used in Curation Concerns.

Effects:
* UI New/Edit/Show page for collections

@projecthydra/sufia-code-reviewers

PR #1976, opened 2016-05-04
Tidying up rubocop

jeremyf:
## Extracting methods for readability

@51884765704d946ec6ec0795aa836e79792e5e7c

The `Sufia::UserStatImporter#import` had a long method (as per a
Rubocop check). Extracting the `#process_files` and `#process_works`
helps remove a violation and improve the immediate readability of the
method.

## Extracting method to reduce method length

@03c5a2f3f78c524dff44609c2f0c961f9e2dab9d

The `CitationsBehaviors::Formatters::OpenUrlFormatter#format` method
was overly long due to the field_map being defined inline. By
extracting the method, we improve the legibility and separate the
different levels of abstraction.

There is also a change in which I switched the `next` guard condition.
There was an assumption that we had a single or an array object, but
then wrap in an array. By moving the logic into an `Array.wrap` method
we tidy up the guard conditions.

Also, note that `Array(Time.now)` explodes the Time object into an
array of its constituent parts:

```ruby
Array(Time.now)
=> [1, 41, 16, 4, 5, 2016, 3, 125, true, "EDT"]
```

So prefer `Array.wrap` when we are using ActiveSupport.

## Tightening the Rubocop parameters

@b0b8c3e03a9eca7773089b209bfd99f376d48366

By adding explicit exclusions and lowering the threshold, we can help
keep future code changes tidier.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -21,10 +21,19 @@ Metrics/LineLength:
   Max: 400
 
 Metrics/AbcSize:
-  Max: 50
+  Max: 30
+  Exclude:
+    - 'app/controllers/concerns/sufia/admin/depositor_stats.rb'
+    - 'app/controllers/concerns/sufia/my_controller_behavior.rb'
+    - 'app/services/sufia/user_stat_importer.rb'
+    - 'lib/sufia/arkivo/metadata_munger.rb'
 
 Metrics/MethodLength:
-  Max: 30
+  Max: 20
+  Exclude:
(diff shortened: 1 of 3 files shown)


jeremyf:
Please hold on this until https://github.com/projecthydra/sufia/pull/1975 is merged.

mjgiarlo:
@jeremyf Looks like your changes to the `UserStatImporter` may have had some unexpected consequences. Happy to merge this once Travis is :smiley:

PR #1992, opened 2016-05-06
Do include on the class instead of the generator

jcoyne:

The application depends on this module being included. It's not something that
is optional or configurable.


mjgiarlo:
Failure above is unrelated to the work in this PR.

PR #2011, opened 2016-05-11
Refactor Admin Stats

jcoyne:


app/services/sufia/admin/depositor_stats.rb
```diff
@@ -0,0 +1,46 @@
...
+  # @param [String] end_date optionally specify the end date to gather the stats from
+  def initialize(start_date, end_date)
+    @start_dt = Time.zone.parse(start_date) unless start_date.blank?
+    @end_dt = Time.zone.parse(end_date).end_of_day unless end_date.blank?
+  end
+
+  attr_accessor :start_dt, :end_dt
+
+  def depositors
+    [].tap do |depositors|
```
cbeer: I know this is just copying from above.. but can't this be simplified? `results.each_slice(2).map`, if nothing else?

app/services/sufia/admin/depositor_stats.rb
```diff
@@ -0,0 +1,46 @@
...
+
+  private
+
+    delegate :blacklight_config, to: CatalogController
+
+    # results come back in an array where the first item is the user and the second item is the count
+    # [ abc123, 55, ccczzz, 205 ]
+    def results
+      facet_results = repository.search(query)
+      facet_results["facet_counts"]["facet_fields"]["depositor_ssim"]
```
cbeer: I wonder why this doesn't use rsolr's accessors?

cam156:
@jcoyne 

I'm a little bit concerned that you are changing the result format from here: https://github.com/projecthydra/sufia/pull/2011/files#diff-148e071fb370f5fa79a702869f724c15L32 to here:
https://github.com/projecthydra/sufia/pull/2011/files#diff-f6d5241feedcd0f1d99159499b23c081R32

I do not see anywhere in the code where you change the view that is using the output.

jcoyne:
@cam156 can you elaborate. My changes should not be affecting the format of that output at all.

app/services/sufia/admin/depositor_stats.rb
jcoyne: Yes, good suggestions.

cam156:
I would expect the original code to output 
[ {key:'abc123',deposits: 55, user: User(abc123)}, {key:'ccczzz',deposits: 205, user: User(ccczzz)}]

I think the new code is going to return [ abc123, 55, ccczzz, 205 ]

jcoyne:
@cam156 it maps to the expected output here: https://github.com/projecthydra/sufia/pull/2011#discussion-diff-62856419R19

cam156:
👍  Missed it!  Thanks!

PR #2012, opened 2016-05-11
Depositor statistics should return user_keys. Fixes #2010

### LINKED ISSUE #2010: Admin stats controller is returning nonsense
It's returning `example.com` as an active user, even though there are only users `user1@example.com` and `user2@example.com`. I think this is because it's using `depositor_tesim` (a parsed field).

I think we can simply switch to `depositor_ssim`

jcoyne:
Previously it was using a parsed text field and splitting user keys that looked like emails on the `@`.


cam156:
@jcoyne Looks like a good idea.  Unfortunately some of the tests refer to the un namespaced class.

mjgiarlo:
:clap: Rad.

PR #2018, opened 2016-05-11
Remove old upgrade generators. Fixes #1967

### LINKED ISSUE #1967: Get rid of upgrade 6.0.0 generator
https://github.com/projecthydra/sufia/blob/master/lib/generators/sufia/upgrade600_generator.rb

Can we merge that back into the install generator?  I'm not sure we want to support a Sufia 5 -> Sufia 7 upgrade. Having an extra generator slows the install process, and adds unnecessary indirection that makes it complicated to understand how it works.

jcoyne:


lib/generators/sufia/templates/migrations/create_proxy_deposit_requests.rb
```diff
@@ -1,7 +1,7 @@
 class CreateProxyDepositRequests < ActiveRecord::Migration
   def change
     create_table :proxy_deposit_requests do |t|
-      t.string :pid, null: false
+      t.string :generic_file_id, null: false
```
mjgiarlo: Shouldn't this be `:generic_work_id`?

mjgiarlo:
Wipe out `update_content_blocks_generator.rb` too?

lib/generators/sufia/templates/migrations/create_proxy_deposit_requests.rb
jcoyne: No, there is a file in the upgrade_700 generator that will do that.

mjgiarlo: :ok_hand:

PR #2022, opened 2016-05-13
Collapse Jobs to common queues

cbeer:


mjgiarlo:
Should we wait to see how https://github.com/projecthydra-labs/curation_concerns/pull/788 and https://github.com/projecthydra-labs/curation_concerns/pull/786 play out before merging this? We'll want Sufia to use the same patterns used in CC.

PR #2025, opened 2016-05-13
Move away from Generic terminology

mjgiarlo:
Waiting on https://github.com/projecthydra/sufia/pull/2019, and will then need rebasing.

### DIFF
--- a/app/assets/javascripts/sufia/app.js
+++ b/app/assets/javascripts/sufia/app.js
@@ -31,7 +31,7 @@ Sufia = {
 
     permissions: function () {
         var perm = require('sufia/permissions/control');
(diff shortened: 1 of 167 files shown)


jcoyne:
I think merging this now is going to make it much harder to add multiple work types at a later date simply because search/replace gets harder.  I'm not sure how I feel about this.

mjgiarlo:
I'm not decided on whether it's going to be worth the effort of rebasing this branch.

mjgiarlo:
We won't merge this branch. Rather, parts of it are being decomposed to prepare for Sufia 7.x to support the full range of curation concern types (rather than assuming `GenericWorks`).

PR #2026, opened 2016-05-13
[WIP] Add list of previously uploaded files to the edit files tab.

paulrosen:
Lists previously uploaded files on the edit page.

The edit page has a tab marked "Files" and it previously only had buttons to upload new files. Users who had previously uploaded files and returned to the edit page couldn't tell if their files were still there or whether they had successfully uploaded them.

(This is also a precursor to the request to modify the displayed label of a file in the "Files" tab. That's why I'm displaying both the file name and the title here.)

### DIFF
--- a/app/forms/sufia/forms/work_form.rb
+++ b/app/forms/sufia/forms/work_form.rb
@@ -4,6 +4,7 @@ module Sufia::Forms
     include HydraEditor::Form::Permissions
 
     attr_reader :agreement_accepted
+    attr_accessor :file_sets
 
     self.terms += [:collection_ids]
     self.required_fields = [:title, :creator, :tag, :rights]
--- a/app/views/curation_concerns/base/_form_files.html.erb
+++ b/app/views/curation_concerns/base/_form_files.html.erb
@@ -8,7 +8,7 @@
                          <th>Filename</th>
                          <th>Display Label</th>
                      </tr>
-                     <% f.object.file_sets.each { |file| %>
+                     <% f.object.file_sets.each do |file| %>
                          <tr>
                              <td>
                                  <span><%= file.solr_document.label %></span>
@@ -17,7 +17,7 @@
                                  <span><%= file.title %></span>
                              </td>
                          </tr>
-                     <% } %>
+                     <% end %>
                      </tbody>
                  </table>
              </fieldset>
--- a/app/views/curation_concerns/base/_form_files.html.erb
+++ b/app/views/curation_concerns/base/_form_files.html.erb
@@ -1,4 +1,27 @@
      <div id="fileupload">
+         <% if f.object.file_sets.present? > 0 %>
+             <fieldset id="uploaded-files">
+                 <legend>Previously Uploaded Files</legend>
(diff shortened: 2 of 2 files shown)


app/views/curation_concerns/base/_form_files.html.erb
```diff
@@ -1,4 +1,27 @@
      <div id="fileupload">
+         <% if f.object.file_sets.present? > 0 %>
+             <fieldset id="uploaded-files">
+                 <legend>Previously Uploaded Files</legend>
+                 <table role="presentation" class="table table-striped uploaded-files">
+                     <tbody>
+                     <tr>
+                         <th>Filename</th>
+                         <th>Display Label</th>
+                     </tr>
```
jcoyne: can you please switch this to `do ... end` instead of `{ }`. Our convention is to use the former for multi-line blocks.

app/forms/sufia/forms/work_form.rb
```diff
@@ -4,6 +4,7 @@ class WorkForm < CurationConcerns::Forms::WorkForm
     include HydraEditor::Form::Permissions
 
     attr_reader :agreement_accepted
+    attr_accessor :file_sets
```
jcoyne: What sets this value?

app/views/curation_concerns/base/_form_files.html.erb
paulrosen: Ok, done.

app/forms/sufia/forms/work_form.rb
paulrosen: Do you mean, "why attr_accessor instead of attr_reader"? It's because of the next pull request that I'll make where I allow the user to change the file title.

jcoyne: No, I mean, that I don't see anything writing anything into this attribute.  That makes me thing we don't have a complete feature here.  Can you please complete the feature and add tests?

paulrosen: The attribute is already in the object by some magic. I don't know what writes it. I just needed to change it from private to public so that I could access it in the view.

I've got a demo to get ready for tomorrow. I'll get back to this Wednesday. I haven't been able to get my environment set up to run the tests yet.

app/views/curation_concerns/base/_form_files.html.erb
```diff
@@ -1,4 +1,27 @@
      <div id="fileupload">
+         <% if f.object.file_sets.present? > 0 %>
```
jcoyne: `present?` returns a boolean. It doesn't make sense to compare it to an integer.

app/forms/sufia/forms/work_form.rb
jcoyne: This method is already public:
```ruby
Sufia::Forms::WorkForm.new(GenericWork.new, nil).files
=> [""]
```

mjgiarlo:
@paulrosen is now a licensed Hydra contributor.

mjgiarlo:
Moving this to Hyrax.

PR #2027, opened 2016-05-16
Extract query methods to a service

jcoyne:

This positions us to remove GenericWork as the service can be modified
to query for more than one class in the future.


app/services/sufia/query_service.rb
```diff
@@ -0,0 +1,47 @@
...
+      start_date_str =  start_datetime.utc.strftime(date_format)
+      end_date_str = if end_datetime.blank?
+                       "*"
+                     else
+                       end_datetime.utc.strftime(date_format)
+                     end
+      "system_create_dtsi:[#{start_date_str} TO #{end_date_str}]"
+    end
+
+    delegate :count, to: :klass
```
jeremyf: Should this class then initialize with `#klass` as an arg then?

jcoyne: My intention is to just keep the existing behavior for now, but to position someone to refactor this class down the road such that it will query every class in `CurationConcerns.config.curation_concerns`

app/services/sufia/system_stats.rb
```diff
@@ -87,13 +87,17 @@ def validate_limit(count)
 
     def document_by_date_by_permission
       files_count = {}
-      files_count[:total] = ::GenericWork.find_by_date_created(start_date, end_date).count
-      files_count[:public] = ::GenericWork.find_by_date_created(start_date, end_date).merge(::GenericWork.where_public).count
-      files_count[:registered] = ::GenericWork.find_by_date_created(start_date, end_date).merge(::GenericWork.where_registered).count
+      files_count[:total] = query_service.find_by_date_created(start_date, end_date).count
+      files_count[:public] = query_service.find_by_date_created(start_date, end_date).merge(query_service.where_public).count
```
jeremyf: Can you add a todo about consolidating the dual query_service call into another method inside query service.

```ruby
query_service.find_public_by_date_created(start_date, end_date).count
```

app/services/sufia/system_stats.rb
```diff
@@ -87,13 +87,17 @@ def validate_limit(count)
...
+      files_count[:registered] = query_service.find_by_date_created(start_date, end_date).merge(query_service.where_registered).count
```
jeremyf: Can you add a todo about consolidating the dual query_service call into another method inside query service.

```ruby
query_service.find_registered_by_date_created(start_date, end_date).count
```

app/services/sufia/system_stats.rb
jcoyne: Damn the TODOs. I'll just create the methods.

mjgiarlo:
Refs #2025.

PR #2028, opened 2016-05-16
Remove GenericWork from the views

jcoyne:
This positions us to offer more work types besides `GenericWork`

### DIFF
--- a/app/views/sufia/homepage/_home_header.html.erb
+++ b/app/views/sufia/homepage/_home_header.html.erb
@@ -1,5 +1,5 @@
 <div class="home_call_action col-xs-12 col-sm-4 pull-right">
-  <% if can? :view_share_work, GenericWork %>
+  <% if @presenter.display_share_button? %>
     <div class="home_share_work">
       <%= link_to sufia.new_curation_concerns_generic_work_path, class: "btn btn-primary btn-lg btn-block", id: "contribute_link" do %>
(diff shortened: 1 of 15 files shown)


mjgiarlo:
Refs #2025.

app/views/sufia/homepage/_home_header.html.erb
```diff
@@ -1,13 +1,13 @@
 <div class="home_call_action col-xs-12 col-sm-4 pull-right">
-  <% if can? :view_share_work, GenericWork %>
+  <% if Sufia.config.always_display_share_button || can_ever_create_works? %>
```
jeremyf: Not a fan of this logic; Where is `can_ever_create_works?` defined? It makes sense to me to have a method that reflects this particular bit of logic?

jcoyne: The idea here is that sufia by default has a "Share your work" button on the homepage. Some institiutions (Oregon Digital) don't want this turned on unless the user actually has permission. `can_ever_create_works?` is from CurationConcerns (originally from Curate I think).

jeremyf: I was wondering about consolidating the following into a single method:

```ruby```
Sufia.config.always_display_share_button || can_ever_create_works?
```

There are two layers of abstraction; A method call and a configuration option co-mingled in the view.

jcoyne: 👌

PR #2029, opened 2016-05-16
Rename generic_work_id columns to work_id

jcoyne:

This will position us to be able to use more work types (besides
`GenericWork`) in the future

### DIFF
--- a/app/models/proxy_deposit_request.rb
+++ b/app/models/proxy_deposit_request.rb
@@ -5,10 +5,10 @@ class ProxyDepositRequest < ActiveRecord::Base
   belongs_to :receiving_user, class_name: 'User'
   belongs_to :sending_user, class_name: 'User'
 
-  # attribute generic_work_id exists as result of renaming in db migrations.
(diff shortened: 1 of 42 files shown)


mjgiarlo:
Refs #2025.

app/models/proxy_deposit_request.rb
```diff
@@ -5,10 +5,10 @@ class ProxyDepositRequest < ActiveRecord::Base
   belongs_to :receiving_user, class_name: 'User'
   belongs_to :sending_user, class_name: 'User'
 
-  # attribute generic_work_id exists as result of renaming in db migrations.
+  # attribute work_id exists as result of renaming in db migrations.
```
jeremyf: Would it make sense to add an `alias_attribute` for `#generic_work_id` as well as a deprectate notice on `#generic_work_id` and `#generic_work_id=`?

jcoyne: We've never released any code that used generic_work_id, so I don't think this is necessary.

PR #2035, opened 2016-05-17
Use the existing Solr connection to query the TermsComponent

mjgiarlo:

### DIFF
--- a/app/services/sufia/system_stats.rb
+++ b/app/services/sufia/system_stats.rb
@@ -98,17 +98,23 @@ class Sufia::SystemStats
       @query_service ||= Sufia::QueryService.new
     end
 
+    def solr_connection
+      ActiveFedora::SolrService.instance.conn
+    end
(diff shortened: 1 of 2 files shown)


app/services/sufia/system_stats.rb
```diff
@@ -98,16 +98,23 @@ def query_service
...
       unless json
-        Rails.logger.error "Unable to reach #{query_url}. Is the terms component enabled in your solr config?"
+        Rails.logger.error "Unable to reach TermsComponent via Solr connection. Is it enabled in your solr config?"
         return []
       end
 
-      tuples = JSON.parse(json)['terms'][key]
+      tuples = json['terms'][key]
       # Change to hash where keys = logins and values = counts
       Hash[*tuples]
```
cbeer: I think we can omit this array -> hash step too if we ask solr to do it for us.

PR #2040, opened 2016-05-18
Rename batchuser and audituser to batch_user and audit_user

jcoyne:


cam156:
@jcoyne I realize the user methods were not previously tested, but how about some tests?  That would make coveralls happier.  Otherwise looks great!

jcoyne:
You want test for deprecated methods that call their non deprecated counterpart?

jcoyne:
Check this out. I increased the coverage without writing any tests.  This is why it's important not to put too much faith in the coverage number. 😉

mjgiarlo:
@cam156 what do you think? is this ready to go, or do you think it needs more work?

PR #2043, opened 2016-05-18
Remove group_list column

jcoyne:

This was part of Scholarsphere's caching of ldap groups


mjgiarlo:
What say ye, Penn Staters? @cam156 @awead @hectorcorrea @mtribone

cam156:
@mjgiarlo I'm good with it.  We have our own user that includes our groups from ldap.  I do not believe this gives us anything beyond not needing the migration for storing groups, which we have overridden anyway.

PR #2046, opened 2016-05-18
Removing group and login from profile update

cam156:
Only permit bulk update of attributes on the user form


jcoyne:
I hope that this PR will fix the build error: https://github.com/projecthydra/sufia/pull/2047

cam156:
@jcoyne & @mjgiarlo: This is looking good after I rebased with jcoyne's fix.

PR #2059, opened 2016-05-19
Delegate collection identifier to solr document

awead:
Collections cannot be viewed if they have an identifier because that property isn't delegated to the solr document.

### DIFF
--- a/spec/presenters/sufia/collection_presenter_spec.rb
+++ b/spec/presenters/sufia/collection_presenter_spec.rb
@@ -58,4 +58,9 @@ describe Sufia::CollectionPresenter do
     subject { presenter.total_items }
     it { is_expected.to eq 0 }
   end
+
+  describe "#identifier" do
+    subject { presenter.identifier }
+    it { is_expected.to be_nil }
+  end
 end
--- a/app/presenters/sufia/collection_presenter.rb
+++ b/app/presenters/sufia/collection_presenter.rb
@@ -1,6 +1,6 @@
 module Sufia
   class CollectionPresenter < CurationConcerns::CollectionPresenter
-    delegate :resource_type, :based_near, :related_url, to: :solr_document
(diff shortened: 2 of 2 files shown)


spec/presenters/sufia/collection_presenter_spec.rb
```diff
@@ -58,4 +58,9 @@
     subject { presenter.total_items }
     it { is_expected.to eq 0 }
   end
+
+  describe "#identifier" do
+    subject { presenter.identifier }
```
jeremyf: Will this always be nil? Or is it a function of the delegation?

https://github.com/thoughtbot/shoulda-matchers/blob/master/lib/shoulda/matchers/independent/delegate_method_matcher.rb

` it { is_expected.to delegate_method(:identifier).to(:solr_document) }`

awead: @jeremyf it's nil because the sample collection's identifier it nil. This is what it looks like if you just create a collection without filling in anything for the identifier.

awead: incidentally, based_near and related_url are also delegated and are being tested the same way.

jeremyf: @awead Would adding shoulda-matchers help convey intention? The `{ is_expected.to be_nil }` is misleading for the specification of behavior; You are expecting delegation.

awead: I'm just following the test convention that already there. Although, yes, using a syntax expecting delegation would be more "unit-y" and less "integrate-y" but we'd need to make that change throughout the application. Do you want to ticket that for a refactor?

jeremyf: @awead With #2060 submitted, I defer my concern to that issue.

Issue #2060, opened 2016-05-19
Add shoulda-matchers for rspec


jeremyf:
At present, there are tests that read something analogous to the following:

```ruby
class MyClass
  delegate :example_of_delegate_method, to: :target
  attr_accessor :target
end

RSpec.describe MyClass do
  context '#example_of_delegate_method' do
    subject { described_class.new.example_of_delegate_method }
    it { is_expected.to be_nil }
  end
end
```

The documentation mode of `rspec` (e.g. `rspec -fd`) renders as such:

```console
MyClass
  #example_of_delegate
    should be nil
```

From a specification and documentation stand point of behavior, we are not clearly stating the intention of the test. We are instead stating a side-effect of delegation.

A solution is to leverage [Shoulda::Matchers delegate_method syntax](https://github.com/thoughtbot/shoulda-matchers/blob/master/lib/shoulda/matchers/independent/delegate_method_matcher.rb). So the above example could instead be represented as:

```ruby
class MyClass
  delegate :example_of_delegate_method, to: :target
  attr_accessor :target
end

RSpec.describe MyClass do
  subject { described_class.new }
  it { is_expected.to delegate_method(:example_of_delegate_method).to(:target) }
end
```

The documentation mode of `rspec` (e.g. `rspec -fd`) renders as such:

```console
MyClass
  should delegate :example_of_delegate to :target
```

I believe this helps clarify the specification.

PR #2065, opened 2016-05-20
2060 add shoulda matcher to rspec

jeremyf:
## Adding should-matchers for delegate_method specs

RSpec is a specification tool. By leveraging existing macros we can
better convey our expected behavior via the documentation. In this
particular instance, the original expected to be nil is confusing as
it is asserting the side-effect of the implementation detail.

Closes to #2060


jcoyne:
It appears that you accidentally committed your fcrepo directory. Can you remove that?

jeremyf:
@jcoyne Good catch. I added that to my global ignore.

jcoyne:
This looks great. Thanks!.

PR #2067, opened 2016-05-20
Removing mocking of `.new` method

### LINKED ISSUE #1939: views/stats/file.html.erb_spec.rb side effects other tests
See #1932 for discovery/discussion.  Obviously, one test file should not compromise the execution of a second one, meaning any manipulation of global config should be undone before leaving scope.

jeremyf:
Removing mocking of `.new` method

In my experience, mocking `.new` method of
objects creates weird behavior. By replacing the
`stub_model` with a `double` I can avoid the mock
of `.new`.

Closes #1939

As a related question, is there a reason we aren't using random spec order?

### DIFF
--- a/spec/views/stats_file.html.erb_spec.rb
+++ b/spec/views/stats_file.html.erb_spec.rb
@@ -8,23 +8,22 @@ describe 'stats/file.html.erb', type: :view do
   describe 'usage statistics' do
     before :each do
       allow_message_expectations_on_nil
-      expect(FileUsage).to receive(:new)
     end
 
     let(:no_stats) {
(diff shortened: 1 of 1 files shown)


jcoyne:
> As a related question, is there a reason we aren't using random spec order?

Probably just because sufia predates when rspec started that by default.

mjgiarlo:
A year or two, I put some effort into making Sufia's suite run in random order and it wasn't pretty. Maybe our suite is in better shape for that now.

atz:
You left the comment in the file about its former filename (being intentional).  Thanks for fixing this oddity though.

Issue #2066, opened 2016-05-20
Passing collection_ids into GenericWorkActor


awead:
### Descriptive summary

When uploading a file via the UI, a `collections_ids` property is passed via the params hash. This does not cause a problem if you're uploading or creating a work in the web application. However, passing this parameter in a unit test of GenericWorkActor results in `ActiveFedora::ObjectNotFoundError`. Ex:

``` ruby
let(:user)  { create(:user) }
let(:work)  { GenericWork.new }
let(:actor) { CurationConcerns::CurationConcern.actor(work, user) }

let(:parameters) do
  {
    "title"=>["Sample Title"],
    "creator"=>["test"],
    "keyword"=>["test"],
    "rights"=>["http://creativecommons.org/licenses/by-nc-sa/3.0/us/"],
    "collection_ids"=>[""]
  }
end

actor.create(parameters)
=> ActiveFedora::ObjectNotFoundError
```

The cause of the problem is:
https://github.com/projecthydra/curation_concerns/blob/master/app/actors/curation_concerns/actors/add_to_collection_actor.rb#L31

Because an empty string is getting passed. However, this problem is not present in the UI. Is there some aspect of the form or view that's omitting arrays of these empty strings?

awead:
I can confirm this behavior in the UI as well if you alter `self.terms` in the GenericWorkForm. You must inherit all terms from `Sufia::Forms::WorkForm`.

jcoyne:
In the form, those empty fields are typically removed by https://github.com/projecthydra/hydra-editor/blob/master/app/forms/hydra_editor/form.rb#L58-L64

awead:
@jcoyne yeah, I think the issue is that if `collection_ids` isn't in `self.terms` it'll never get run through that method to clean it out.

mjgiarlo:
@awead can you tag this issue with some labels? It's not clear to me whether this is a bug or what.

awead:
@mjgiarlo I don't think it's a bug, exactly, but it pertains to how one would override and add additional properties. For example, you'll have to inherit self.terms whether you use the terms or not. So, it should be captured in documentation, perhaps a refactor of GenericWorkForm that makes it more explicit as to which terms you *have* to have in your form versus which are just optional metadata.

jcoyne:
@awead, @mjgiarlo what milestone for this ticket?

awead:
[...]

PR #2070, opened 2016-05-20
Removing nil or empty string identifiers from CreateWithFilesActor

### LINKED ISSUE #2051: If file upload errors, then work creation fails.
Because the file upload sets a hidden field on the page with no value.  Then when you submit a work it looks for an uploaded file with the ID of nil:
```
ActiveRecord::RecordNotFound (Couldn't find all Sufia::UploadedFiles with 'id': (, 1) (found 1 results, but was looking for 2)):
  activerecord (4.2.6) lib/active_record/relation/finder_methods.rb:324:in `raise_record_not_found_exception!'
  activerecord (4.2.6) lib/active_record/relation/finder_methods.rb:467:in `find_some'
  activerecord (4.2.6) lib/active_record/relation/finder_methods.rb:426:in `find_with_ids'
  activerecord (4.2.6) lib/active_record/relation/finder_methods.rb:71:in `find'
  activerecord (4.2.6) lib/active_record/querying.rb:3:in `find'
  activerecord (4.2.6) lib/active_record/core.rb:131:in `find'
  /opt/rubies/ruby-2.3.0/lib/ruby/gems/2.3.0/bundler/gems/sufia-9c1afba2bb55/app/actors/sufia/create_with_files_actor.rb:38:in `uploaded_files'
  /opt/rubies/ruby-2.3.0/lib/ruby/gems/2.3.0/bundler/gems/sufia-9c1afba2bb55/app/actors/sufia/create_with_files_actor.rb:19:in `validate_files'
  /opt/rubies/ruby-2.3.0/lib/ruby/gems/2.3.0/bundler/gems/sufia-9c1afba2bb55/app/actors/sufia/create_with_files_actor.rb:6:in `create'
  curation_concerns (1.0.0.beta3) app/actors/curation_concerns/actors/actor_stack.rb:21:in `create'
  /opt/rubies/ruby-2.3.0/lib/ruby/gems/2.3.0/bundler/gems/sufia-9c1afba2bb55/app/actors/sufia/create_with_remote_files_actor.rb:6:in `create'
  curation_concerns (1.0.0.beta3) app/actors/curation_concerns/actors/actor_stack.rb:21:in `create'
  curation_concerns (1.0.0.beta3) app/controllers/concerns/curation_concerns/curation_concern_controller.rb:43:in `create'
```

We should probably remove any nil file upload ids in `create_with_files_actor.rb`

(BTW, the initial file upload failed because it was too big and exceeded the limit set in nginx)

jeremyf:
Removing nil or empty string identifiers

Following @jcoyne's suggestion of removing any nil file upload ids
as it relates to #2051 (e.g. "If file upload errors, then
work creation fails.")

Closes #2051

### DIFF
--- a/app/actors/sufia/create_with_files_actor.rb
+++ b/app/actors/sufia/create_with_files_actor.rb
@@ -2,16 +2,21 @@ module Sufia
   # Creates a work and attaches files to the work
   class CreateWithFilesActor < CurationConcerns::Actors::AbstractActor
     def create(attributes)
-      @uploaded_file_ids = attributes.delete(:uploaded_files)
+      self.uploaded_file_ids = attributes.delete(:uploaded_files)
       validate_files && next_actor.create(attributes) && attach_files
     end
 
     def update(attributes)
-      @uploaded_file_ids = attributes.delete(:uploaded_files)
+      self.uploaded_file_ids = attributes.delete(:uploaded_files)
       validate_files && next_actor.update(attributes) && attach_files
     end
 
     protected
+
+      attr_reader :uploaded_file_ids
+      def uploaded_file_ids=(input)
+        @uploaded_file_ids = Array.wrap(input).select(&:present?)
+      end
 
       # ensure that the files we are given are owned by the depositor of the work
       def validate_files
@@ -34,8 +39,8 @@ module Sufia
 
       # Fetch uploaded_files from the database
       def uploaded_files
-        return [] unless @uploaded_file_ids
-        @uploaded_files ||= UploadedFile.find(@uploaded_file_ids)
+        return [] if uploaded_file_ids.empty?
+        @uploaded_files ||= UploadedFile.find(uploaded_file_ids)
       end
   end
 end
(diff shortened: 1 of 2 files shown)


awead:
@jeremyf does this also relate to #2066 ? It's a similar situation where `[""]` is getting passed in the parameters hash.

jcoyne:
Can you check to see if the files attribute is or is not going through here? https://github.com/projecthydra/hydra-editor/blob/master/app/forms/hydra_editor/form.rb#L58-L64  I wonder if it makes sense to update that method?

jeremyf:
@awead At first I had used `Array.wrap(uploaded_file_ids).compact` but assumed that I might get `[nil, 1, ""]` so I switched to `Array.wrap(input).select(&:present?)` which should catch that scenario.

jeremyf:
@jcoyne I won't be able to check that in my allotted time today; And will be on a new spring at ND starting Monday. So could you add the issue to HydraEditor as a follow-up?

jeremyf:
Bringing @projecthydra/sufia-code-reviewers into the conversation regarding the particular bug.

PR #2071, opened 2016-05-20
Splitting apart Rake tasks

jeremyf:
Splitting apart Rake tasks

Because sometimes, I don't want to generate from engine cart, but
instead prefer to spin up the test servers (i.e. Fedora and SOLR) and
then run the tests against my existing engine cart provided
application.

### DIFF
--- a/tasks/sufia-dev.rake
+++ b/tasks/sufia-dev.rake
@@ -16,16 +16,18 @@ task :spec do
   RSpec::Core::RakeTask.new(:spec)
 end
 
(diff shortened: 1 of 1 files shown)


awead:
@jeremyf why not just keep your solr and fedora running and `bundle exec rspec`? The test suite is going to clean them out before each run.

jeremyf:
@awead That is another option; However I don't recall all of the server commands to do this. So instead, I have a task that I can use.

PR #2074, opened 2016-05-20
Refactored the test to speed it up.

hectorcorrea:
The new way of calling "save" also exposed a bug in the JavaScript code when queuing the JavaScript requests which we have fixed.

Refactoring session with @cam156 @awead @mtribone

### DIFF
--- a/app/assets/javascripts/sufia/batch_edit.js
+++ b/app/assets/javascripts/sufia/batch_edit.js
@@ -79,6 +79,7 @@ function batch_edit_init () {
             },
             combine_requests: function (requests) {
                 var data = deserialize(requests[0].data.replace(/\+/g, " "));
+                var adata;
                 form = [requests[0].form]
                 for (var i = requests.length - 1; i > 0; i--) {
                     req = requests.pop();
@@ -86,7 +87,7 @@ function batch_edit_init () {
 
                     for (key in  Object.keys(adata)) {
                         curKey = Object.keys(adata)[key];
-                        if (curKey.slice(0, 12) == "file_set") {
+                        if (curKey.slice(0, 12) == "generic_work") {
                             data[curKey] = adata[curKey];
                             form.push(req.form);
                         }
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/sufia/batch_edit.js
```diff
@@ -79,14 +79,15 @@ function batch_edit_init () {
             },
             combine_requests: function (requests) {
                 var data = deserialize(requests[0].data.replace(/\+/g, " "));
+                var adata;
```
mjgiarlo: What's this variable about? Cruft, or?

app/assets/javascripts/sufia/batch_edit.js
```diff
@@ -79,14 +79,15 @@ function batch_edit_init () {
...
                 form = [requests[0].form]
                 for (var i = requests.length - 1; i > 0; i--) {
                     req = requests.pop();
                     adata = deserialize(req.data.replace(/\+/g, " "));
 
                     for (key in  Object.keys(adata)) {
                         curKey = Object.keys(adata)[key];
-                        if (curKey.slice(0, 12) == "file_set") {
+                        if (curKey.slice(0, 12) == "generic_work") {
```
mjgiarlo: Good catch!

app/assets/javascripts/sufia/batch_edit.js
hectorcorrea: That variable is being used in line 86 but it wasn't being declared. This PR declares it so that is local to the function rather than going to the global scope.

app/assets/javascripts/sufia/batch_edit.js
atz: Hah, if 12-char string matches 8-char string...

mjgiarlo: I guess it works?

mjgiarlo: @hectorcorrea ?

Issue #2080, opened 2016-05-24
Work does not show any help text by default


cam156:
### Descriptive summary

In Sufia 7.0

### Expected behavior

Work help tags would show the help

![screen shot 2016-05-24 at 11 01 57 am](https://cloud.githubusercontent.com/assets/1599081/15508809/f752db3c-219e-11e6-9c7b-e76bff16b530.png)
### Actual behavior

![screen shot 2016-05-24 at 10 43 02 am](https://cloud.githubusercontent.com/assets/1599081/15508764/c831e730-219e-11e6-988b-0cfead1a4130.png)
![screen shot 2016-05-24 at 10 41 35 am](https://cloud.githubusercontent.com/assets/1599081/15508772/ca7f2070-219e-11e6-9403-364a7590757c.png)

### Steps to reproduce the behavior

1. Got to Add New Work
1. And then click on the help

### Related work

#2068
#2003

cam156:
UNTESTED!!! https://github.com/projecthydra/sufia/blob/master/config/locales/sufia.en.yml#L239 should be `work:` instead of `file_set:`

mjgiarlo:
@cam156 or `generic_work:` depending on how we're mapping models to the structure of the locale file.

This may be included in the work @jonearley or @fultonis are doing. Is that true, @mtribone? Or am I getting this confused with other work?

mtribone:
@mjgiarlo Yes addressing the help text is part of their work.

PR #2088, opened 2016-05-24
Cleanup i18n values

### LINKED ISSUE #2077: locales file values should avoid double encoding
The locales files like `config/locales/sufia.en.yaml` should avoid double encoding values to accomplish HTML-injection.  This *should* be a default rails behavior as described here:
    http://guides.rubyonrails.org/i18n.html#using-safe-html-translations

But we are apparently working around or disabling that default.  Example:
```yaml
based_near_html: "A place name related to the file, such as its site of publication, or the city, state, or country the file's contents are about. Calls upon the GeoNames web service (&lt;a href=&quot;http://www.geonames.org&quot;&gt;http://www.geonames.org&lt;/a&gt;)."
```

In a `html` field value we should be able to use regular `<a>` values (as in the `simple_form.en.yml` example).  Instead we get strings that are html_decoded in order to inject them into HTML.  o_0

atz:
- No HTML in non-`_html` keys
- Remove "This is a required field" statements: required fields are
  dynamic. Phrase should be translated once and invoked by simple_form.
- Make link to geonames use text description instead of just URL.

Fixes #2077

### DIFF
--- a/config/locales/sufia.en.yml
+++ b/config/locales/sufia.en.yml
@@ -238,21 +238,21 @@ en:
     metadata_help:
       file_set:
         resource_type_html: "Pre-defined categories to describe the type of file content being uploaded, such as &quot;article&quot; or &quot;dataset.&quot;  More than one type may be selected."
-        title: "A name for the file to aid in identifying it. Defaults to the file name, though a more descriptive title is encouraged. &lt;em&gt;This is a required field&lt;/em&gt;."
-        keyword: "Words or phrases you select to describe what the file is about. These are used to search for content. &lt;em&gt;This is a required field&lt;/em&gt;."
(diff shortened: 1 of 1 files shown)


justcolin:
@atz: is this ticket ready for review?

mjgiarlo:
I believe it is, @justcolin

justcolin:
@atz: Before merging this can you also address #2080 in this PR? It should just involve changing the `file_set` key. I had a fix for it, but your PR will get pulled before mine and I can't validate these change in the UI without that fix.

Thanks!

atz:
I don't see the need to address separate tickets here.  I don't mind picking it up, but won't block this for it.

mjgiarlo:
@justcolin I'm inclined to merge this as is, and I'll put in a follow-up PR for #2080.

PR #2092, opened 2016-05-25
Renames locale keys to light up helps for works

### LINKED ISSUE #2080: Work does not show any help text by default

mjgiarlo:
Fixes #2080

This PR also:
* replaces 'file' in the helps with 'item'
* removes usage guidelines that were from ScholarSphere
* applies work metadata helps to collection metadata helps, since those helps were also broken
* lets collection labels use work labels rather than duplicating them

### DIFF
--- a/config/locales/sufia.en.yml
+++ b/config/locales/sufia.en.yml
@@ -217,42 +217,50 @@ en:
 
   simple_form:
     labels:
-      file_set:
+      generic_work: &work_labels
         based_near:       "Location"
         description:      "Abstract or Summary"
         keyword:          "Keyword"
         date_created:     "Date Created"
         related_url:      "Related URL"
         files:            "Upload a new version of this file from your computer"
+        collection_ids:   "Add as member of collection"
       collection:
-        based_near:       "Location"
-        description:      "Abstract or Summary"
-        keyword:          "Keyword"
-        date_created:     "Date Created"
-        related_url:      "Related URL"
         total_items:      "Total Items"
         size:             "Size"
-      generic_work:
-        collection_ids:   "Add as member of collection"
+        <<: *work_labels
 
     metadata_help:
-      file_set:
-        resource_type_html: "Pre-defined categories to describe the type of file content being uploaded, such as &quot;article&quot; or &quot;dataset.&quot;  More than one type may be selected."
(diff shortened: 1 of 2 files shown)


justcolin:
@projecthydra/sufia-ui-ux-advisors: How do you feel about having help text for the various metadata fields referring to the thing you are adding metadata to as an "object" instead of a "file"?

An example is the publisher field which would read as "The person or group making the **object** available. Generally this is the institution." instead of "The person or group making the **file** available. Generally this is the institution"

justcolin:
@heidisb? (See above)

heidisb:
Ehhh. Not super into object. I get the resistance with file since it maybe
insinuates document. What about "item"?

Heidi

---
Heidi Steiner Burkhardt | User Experience Specialist | University of
Michigan Library
heidisb@umich.edu | @heidi_sb

mjgiarlo:
@heidisb I'm cool with "item." Will make the change in my next go-round (and apparently I have failures to clean up...).

newmanld:
Item is better than object.  But I think 'work' is better than both.  I think that PCDM moved from the term 'work' to 'object', but I still think in terms of works (with attached files).  

In at least one case we don't need the word at all - 'describe the type of file content being uploaded' could be simply 'describe the type of content being uploaded'.  But in other cases we do need a description for the result of the input process.  I lean toward 'work' but can live with 'item'.

mjgiarlo:
@newmanld I opted for item in this case since works and collections are currently sharing help messages. They don't *need* to share help messages, though, and some code duplication is defensible if it means our helps are more contextually accurate. I'm not sure "item" works equally well for works and collections. I'm now inclined to call works "works" and collections "collections".

newmanld:
@mjgiarlo - I would agree with calling works 'works' and collections 'collections' and not trying to avoid duplication of help messages by calling both 'items.'  Avoid being abstract where we can?

mjgiarlo:
@newmanld Thanks! Pushed that change.

Issue #2073, opened 2016-05-20
Investigate why Form#model_attributes isn't preventing issue #2051


jcoyne:
Is it because it only handles empty string an not nil?  If we update our assumptions, are there any changes that should be made to the changes in https://github.com/projecthydra/sufia/pull/2070 ?

atz:
ref: #2051

atz:
In Sufia, we have:
- `Sufia::Forms::WorkForm < CurationConcerns::Forms::WorkForm`
- `Sufia::Forms::BatchEditForm < CurationConcerns::Forms::WorkForm`
- `Sufia::BatchUploadForm < Sufia::Forms::WorkForm`

Should `BatchUploadForm` be in the `Forms` module?  More importantly, if batch *upload* is able to inherit from `Sufia::Forms::WorkForm`, what is the reason batch *edit* doesn't as well?  Just trying to get some context here.

jcoyne:
👍 to consistency.

mjgiarlo:
:+1: to `BatchUploadForm` being in the `forms` module.

mjgiarlo:
As to the inheritance hierarchy, I'm not sure why the `BatchEditForm` inherits from CC's work form rather than Sufia's form. I'd be happy to see those made consistent as well.

atz:
@jcoyne how do I test to reproduce #2051?

jcoyne:
@atz you need to have one of your file uploads fail and one succeed. I'm not sure how you can create that condition. Perhaps turn off fedora as you are uploading the first file?

atz:
@jcoyne Can you describe the necessary timing and location of the failure?  My interpretation is that this is during non-batch work creation, drag-drop a couple files, while they auto-upload, first succeeds, second fails.  Is the failure specifically between the jQuery file-uploader and the webserver? To the browser, does that look like a 500 error or a dropped connection (killed socket)?

atz:
Is it appropriate to test this at the controller level?

jcoyne:
Yes, this could be a controller test

atz:
On one level, the problem is about filtering the empty uploaded file "line" from the post to create the work.  But on another, it might also be about the jQuery file-uploader reacting appropriately to failure on the server side to avoid adding the empty line (or removing it if already added).  The approaches are not exclusive.

jcoyne:
[...]

PR #2100, opened 2016-05-26
Make batch_upload_form live in under Sufia::Forms

atz:
This makes it consistent with batch_edit_form and the others.


atz:
ref: #2073

PR #2101, opened 2016-05-27
Cleanup test

atz:
secondary_terms test wasn't even testing secondary terms


spec/forms/sufia/batch_upload_form_spec.rb
```diff
@@ -7,24 +7,20 @@
   describe "#primary_terms" do
     subject { form.primary_terms }
     it { is_expected.to eq [:creator, :keyword, :rights] }
+    it { is_expected.not_to include(:title) }
```
jcoyne: To me, it is clear that title is not in the three element list on line 9, so this test becomes superfluous. It's not an issue to hold up the PR over though.

PR #2108, opened 2016-05-31
Upload failure filter

atz:
Relates to #2073

### DIFF
--- a/spec/actors/create_with_files_actor_spec.rb
+++ b/spec/actors/create_with_files_actor_spec.rb
@@ -22,7 +22,7 @@ describe Sufia::CreateWithFilesActor do
       context "when uploaded_file_ids include nil" do
         let(:uploaded_file_ids) { [nil, uploaded_file1.id, nil] }
         it "will discard those nil values when attempting to find the associated UploadedFile" do
-          allow(AttachFilesToWorkJob).to receive(:perform_later)
+          expect(AttachFilesToWorkJob).to receive(:perform_later)
           expect(Sufia::UploadedFile).to receive(:find).with([uploaded_file1.id]).and_return([uploaded_file1])
           actor.public_send(mode, attributes)
         end
--- a/spec/controllers/generic_works_controller_spec.rb
+++ b/spec/controllers/generic_works_controller_spec.rb
@@ -94,9 +94,7 @@ describe CurationConcerns::GenericWorksController do
     let(:actor) { double('An actor') }
     let(:work) { create(:work) }
     before do
-      allow(controller).to receive(:actor)
-        .and_return(actor)
-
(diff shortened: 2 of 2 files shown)


spec/actors/create_with_files_actor_spec.rb
```diff
@@ -22,7 +22,7 @@
       context "when uploaded_file_ids include nil" do
         let(:uploaded_file_ids) { [nil, uploaded_file1.id, nil] }
         it "will discard those nil values when attempting to find the associated UploadedFile" do
-          allow(AttachFilesToWorkJob).to receive(:perform_later)
+          expect(AttachFilesToWorkJob).to receive(:perform_later)
```
jcoyne: I don't think the idea was to test for this here, meerly to prevent it from occurring.

spec/controllers/generic_works_controller_spec.rb
```diff
@@ -136,12 +134,8 @@
                                  file_set: {}
           }.to change(FileSet, :count).by(2)
           created_files = FileSet.all
-          ["https://dl.dropbox.com/fake/blah-blah.Getting%20Started.pdf", "https://dl.dropbox.com/fake/blah-blah.filepicker-demo.txt.txt"].each do |url|
-            expect(created_files.map(&:import_url)).to include(url)
-          end
-          ["filepicker-demo.txt.txt", "Getting+Started.pdf"].each do |filename|
-            expect(created_files.map(&:label)).to include(filename)
-          end
+          expect(created_files.map(&:import_url)).to include(url1, url2)
```
jcoyne: I think this is harder to read. Now I have to figure out what rspec evaluates created_files to be before I can figure out why this test isn't working.  This is more difficult for a human to understand.

atz: You **already** have to understand created_files to understand this test file and these tests.  This is more DRY and utilizes the expressivity of `include`.   Loops over `expect` are anti-pattern in rspec, especially when include can already check for multiple values.

spec/actors/create_with_files_actor_spec.rb
atz: Right, I'm saying this test *should* break if `perform_later` never gets called.  This is how you avoid cargo cult stubbing.  If you need to stub it, it should be because you expect it to be called.

mjgiarlo:
I'm good with these changes. I'll give @jcoyne a chance to respond to @atz's replies.

spec/controllers/generic_works_controller_spec.rb
jcoyne: I agree that loops should be removed. I don't think that a DRY test is always the easiest to understand.

PR #2110, opened 2016-05-31
Produce statistics for collections

jcoyne:


atz:
CI is hung up on rubocopu right now.

jcoyne:
Yeah. This is WIP. Pushed because I had to do meetings for the remainder of the day

app/services/sufia/statistics/collections.rb
```diff
@@ -0,0 +1,31 @@
...
+      # @param [Fixnum] delta_x change in x (in days)
+      # @param [Date] x_min minimum date
+      # @param [Date] x_max max date
+      def initialize(delta_x: 7, x_min: 1.month.ago.beginning_of_day, x_max: Date.today.end_of_day)
+        @delta_x = delta_x
+        @x_min = x_min
+        @x_max = x_max
+      end
+
+      def points
```
cbeer: Would it be useful to make this an enumerable? At the very least, it'd read better to me.

```ruby
def points
  return to_enum(:points) unless block_given?

  x = @x_min
  while (x <= @x_max)
    x += @delta_x.days
    yield [ x.to_i * 1000, y: collections_in_range(@x_min, x) ]
  end 
```

app/services/sufia/statistics/collections.rb
```diff
@@ -0,0 +1,31 @@
+module Sufia
+  module Statistics
+    class Collections
+      # @param [Fixnum] delta_x change in x (in days)
+      # @param [Date] x_min minimum date
+      # @param [Date] x_max max date
+      def initialize(delta_x: 7, x_min: 1.month.ago.beginning_of_day, x_max: Date.today.end_of_day)
```
cbeer: I see this is a pattern from the other stats classes, and there's probably value in keeping it. I'm surprised it isn't a `Range` type :shrug:.

app/services/sufia/statistics/collections.rb
jcoyne: :hear_no_evil: I can't hear you. Must be because you're on vacation. :wink:

app/services/sufia/statistics/collections.rb
jcoyne: Using ranges of DateTime with arbitrary step size is hard.

PR #2140, opened 2016-06-06
Add a method to fetch number of works by resource type

jcoyne:

Also a substantial refactor of the existing statistics classes.
* Move date parsing out of services and into the presenter (maybe it should be moved to the controller?)
* Moved the statistics classes into the `Statistics` module
* Decomposed the `SystemStats` class into `Works::ByDepositor`, `FileSets::ByFormat`, `Works::Count`, and `TermQuery`
* Added `Works::ByResourceType` and index the resource type in a non-parsed way so that it is useful to the Solr terms query handler.


awead:
I like the approach. I give it a cursory 👍 but others should take a look.

PR #2177, opened 2016-06-08
Fix 2175 - Use plural labels for relationships

### LINKED ISSUE #2175: Use Plural Language for Works -> Batch Create
### Descriptive summary

Version: Sufia 7.x
UI Change

Relationships tab uses singular language (i.e. This Work) instead of plural (i.e. These New Works)

### Expected behavior

Relationships tab uses plural: These New Works in Collections

See:  [UIWG wireframe](https://wiki.duraspace.org/x/5oZcB)

### Actual behavior

Relationships tab uses singular language:  This Work in Collections

1. Top menu -> Works -> [Batch Create](http://localhost:3000/batch_uploads/new)
1. Click Relationships tab
1. Verify in Collections section appears as defined in [UIWG wireframe](https://wiki.duraspace.org/x/5oZcB)

### Related work

#2174  Works -> Batch Create: Move Related URL from Metadata tab to Relationships tab

elrayle:
Fixes #2175 ; refs #issuenumber

Adds a check to determine if the current page is batch create works and sets the label for 'in Collections' accordingly.

### DIFF
--- a/app/views/curation_concerns/base/_form_relationships.html.erb
+++ b/app/views/curation_concerns/base/_form_relationships.html.erb
@@ -1,4 +1,4 @@
-<h3>This Work in Collections</h3>
+<h3><%= t("sufia.#{controller_name}.#{action_name}.in_collections") %></h3>
 <div id="collection-widget">
   <%= f.input :collection_ids, as: :select,
       collection: available_collections(nil),
--- a/config/locales/sufia.en.yml
+++ b/config/locales/sufia.en.yml
@@ -160,8 +160,10 @@ en:
     generic_works:
       new:
         header: Add New Work
+        in_collections: This Work in Collections
       edit:
         header: Edit Work
+        in_collections: This Work in Collections
       progress:
         header: Save Work
       show:
@@ -169,6 +171,7 @@ en:
     batch_uploads:
       new:
         header: Batch Create New Works
+        in_collections: These Works in Collections
       progress:
         header: Save Works
       files:
--- a/spec/views/curation_concerns/base/_form.html.erb_spec.rb
+++ b/spec/views/curation_concerns/base/_form.html.erb_spec.rb
@@ -16,6 +16,8 @@ describe 'curation_concerns/base/_form.html.erb' do
     allow(view).to receive(:curation_concern).and_return(work)
     allow(controller).to receive(:current_user).and_return(stub_model(User))
     assign(:form, form)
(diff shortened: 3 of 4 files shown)


app/views/curation_concerns/base/_form_relationships.html.erb
```diff
@@ -1,4 +1,9 @@
-<h3>This Work in Collections</h3>
+<%
+  batch = current_page?(sufia.new_batch_upload_path)
+  in_collection_label = "This Work in Collections"
+  in_collection_label = "These New Works in Collections" if batch
+%>
+<h3><%= in_collection_label %></h3>
```
jcoyne: Is there a reason we have to assign a variable?  Can you do:
```
<% if current_page?(...) %>
This Work in Collections
<% else %>
Other thing
<% end %>
```

mjgiarlo: Or can we handle this via i18n (vs. having logic in the view) like here? https://github.com/projecthydra/sufia/blob/master/app/views/curation_concerns/base/new.html.erb#L3

I'd also recommend re-wording a bit. Why would one label use the word "new" and the other not? Suggest striking it from both. Thoughts, @projecthydra/sufia-ui-ux-advisors?

jcoyne: I18n would be even better.

elrayle:
I totally think this should be handled via I18n.  The same conditional will apply.  I was putting in the simplest fix, which is what I had time to do.  I may have to come back to this later if you want the I18n fix in before merging.  OR we could open a second issue for all the I18n issues for New Work and Batch Create works.

mjgiarlo:
@elrayle See the link I posted -- that's a way to handle this without any conditional in the view. Thanks!

elrayle:
I was being too quick with my response.  I took a look at @mjgiarlo's suggested modification and it was a very simple change.  I put the requested change in.

NOTE: This still doesn't cover other places where labels are hardcoded in the view.

elrayle:
@mjgiarlo There are tests failing related to a test controller.  Is there a fix for this when using the controller and action to drive the label?

```
translation missing: en.sufia.test.in_collections
```

jcoyne:
@elrayle you can stub out the controller_name

elrayle:
@jcoyne Thanks for pointing me to stubbing the controller_name.  That worked.

PR #2179, opened 2016-06-09
Adjusting test to match updated blacklight-access-controls

awead:
The solr query changed in blacklight-access-controls 0.5.1. This changes the test expectation to match.

### DIFF
--- a/spec/search_builder/sufia/my_shares_search_builder_spec.rb
+++ b/spec/search_builder/sufia/my_shares_search_builder_spec.rb
@@ -8,11 +8,13 @@ describe Sufia::MySharesSearchBuilder do
                        current_user: me) }
   let(:builder) { described_class.new(scope) }
 
+  before { allow(builder).to receive(:gated_discovery_filters).and_return(["access_filter1", "access_filter2"]) }
+
   subject { builder.to_hash['fq'] }
 
   it "filters things we have access to in which we are not the depositor" do
-    expect(subject).to eq ["edit_access_group_ssim:public OR edit_access_group_ssim:registered OR edit_access_person_ssim:#{me.user_key}",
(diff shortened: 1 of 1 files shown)


cbeer:
Do we need to pin the blacklight-access-controls dependency too?

spec/search_builder/sufia/my_shares_search_builder_spec.rb
```diff
@@ -11,8 +11,8 @@
   subject { builder.to_hash['fq'] }
 
   it "filters things we have access to in which we are not the depositor" do
-    expect(subject).to eq ["edit_access_group_ssim:public OR edit_access_group_ssim:registered OR edit_access_person_ssim:#{me.user_key}",
+    expect(subject).to eq ["({!terms f=edit_access_group_ssim}public,registered) OR edit_access_person_ssim:user1@example.com",
                            "{!terms f=has_model_ssim}GenericWork,Collection",
-                           "-_query_:\"{!field f=depositor_ssim}#{me.user_key}\""]
+                           "-_query_:\"{!field f=depositor_ssim}user1@example.com\""]
```
jcoyne: I don't think you want to hard code the username. Continue to use `#{me.user_key}`

jcoyne:
If we want to pin, I prefer it be done at the hydra-access-controls level https://github.com/projecthydra/hydra-head/blob/master/hydra-access-controls/hydra-access-controls.gemspec#L26

jcoyne:
we could write the test in a was such that it stubs the return value from apply_access_controls and then it wouldn't matter.

spec/search_builder/sufia/my_shares_search_builder_spec.rb
awead: doh! copy pasta...

PR #2180, opened 2016-06-09
Add I18n settings to toolbar

awead:
Add I18n terms to the toolbar.

### DIFF
--- a/app/views/_toolbar.html.erb
+++ b/app/views/_toolbar.html.erb
@@ -1,55 +1,39 @@
 <% if user_signed_in? %>
   <ul class="nav navbar-nav">
-    <% if can? :manage, :all %>
-      <li class="dropdown">
-        <%= link_to sufia.admin_stats_path, role: 'button', data: { toggle: 'dropdown' }, aria: { haspopup: true, expanded: false } do %>
-          <span class="fa fa-gear"></span> Admin <span class="caret"></span>
-        <% end %>
-        <ul class="dropdown-menu">
-          <% if respond_to? :admin_stats_path %>
-            <li><%= link_to 'Repository Statistics', sufia.admin_stats_path %></li>
-          <% end %>
-
-          <% if respond_to? :admin_queues_path %>
-            <li><%= link_to 'Message Queues', sufia.admin_queues_path %></li>
-          <% end %>
-        </ul>
-      </li>
-    <% end %>
-
+    <%= render '/admin/menu' if can? :manage, :all %>
     <li class="dropdown">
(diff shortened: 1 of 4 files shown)


app/views/_toolbar.html.erb
```diff
@@ -3,65 +3,65 @@
...
-          <span class="fa fa-gear"></span> Admin <span class="caret"></span>
+          <span class="fa fa-gear"></span> <%= t("sufia.toolbar.admin.menu") %> <span class="caret"></span>
         <% end %>
         <ul class="dropdown-menu">
           <% if respond_to? :admin_stats_path %>
-            <li><%= link_to 'Repository Statistics', sufia.admin_stats_path %></li>
+            <li><%= link_to t("sufia.toolbar.admin.statistics"), sufia.admin_stats_path %></li>
           <% end %>
 
           <% if respond_to? :admin_queues_path %>
```
jcoyne: Can we just eliminate this one? it doesn't respond to this anymore.

awead: Alternatively, could we provide a blank partial that uses could override with links to whatever queuing service they're using?

jcoyne: I'm not a fan of that. I wouldn't mind breaking the admin dropdown partial out. Then you'd only have a few lines to duplicate if you choose to add additional items.

awead: That's sounds even better.

config/locales/sufia.en.yml
```diff
@@ -73,6 +73,26 @@ en:
         details:            'Details'
         tab_label:          'Recently Uploaded'
         no_public:          'No public work has been contributed.'
+    toolbar:
+      admin:
+        menu: "Admin"
+        statistics: "Repository Statistics"
+        queues: "Message Queues"
```
jcoyne: Remove this.

app/views/admin/_menu.html.erb
```diff
@@ -0,0 +1,10 @@
+<li class="dropdown">
+  <%= link_to sufia.admin_stats_path, role: 'button', data: { toggle: 'dropdown' }, aria: { haspopup: true, expanded: false } do %>
+    <span class="fa fa-gear"></span> <%= t("sufia.toolbar.admin.menu") %> <span class="caret"></span>
+  <% end %>
+  <ul class="dropdown-menu">
+    <% if respond_to? :admin_stats_path %>
```
jcoyne: Can we remove this conditional? It should always be true.

mjgiarlo: Yes, this is now true. Related: we should remove this conditional as well: https://github.com/projecthydra/sufia/blob/master/config/routes.rb#L153

I'd be fine with seeing that snuck into this PR.

awead: See #2183

PR #2184, opened 2016-06-09
Give media_display a presenter

jcoyne:

Media display only takes a presenter in the latest version of
curation_concerns. Previously we were sending a FileSet.

### DIFF
--- a/app/models/concerns/sufia/file_set_behavior.rb
+++ b/app/models/concerns/sufia/file_set_behavior.rb
@@ -2,5 +2,10 @@ module Sufia
   module FileSetBehavior
     extend ActiveSupport::Concern
     include Sufia::WithEvents
+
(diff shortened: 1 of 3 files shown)


app/models/concerns/sufia/file_set_behavior.rb
```diff
@@ -2,5 +2,10 @@ module Sufia
   module FileSetBehavior
     extend ActiveSupport::Concern
     include Sufia::WithEvents
+
+    # Cast to a SolrDocument by querying from Solr
+    def to_presenter
+      CatalogController.new.fetch(id).last
```
awead: This is clever. Never seen it before... Is this a Blacklight thing or a Rails thing?

jcoyne: it's a blacklight thing, and relatively new (6.x)

PR #2189, opened 2016-06-09
Link on collections show page should be "Add new work"

mjgiarlo:

### DIFF
--- a/app/views/collections/show.html.erb
+++ b/app/views/collections/show.html.erb
@@ -1,35 +1,32 @@
 <% provide :page_title, construct_page_title(@presenter.title) %>
-  <div itemscope itemtype="http://schema.org/CollectionPage" class="row">
-    <div class="col-sm-10 pull-right">
-      <header>
-        <%= render 'collection_title', presenter: @presenter %>
-      </header>
-      <%= render 'collection_description', presenter: @presenter %>
+<div itemscope itemtype="http://schema.org/CollectionPage" class="row">
+  <div class="col-sm-10 pull-right">
+    <header>
+      <%= render 'collection_title', presenter: @presenter %>
+    </header>
+    <%= render 'collection_description', presenter: @presenter %>
 
-      <% unless has_collection_search_parameters? %>
+    <% unless has_collection_search_parameters? %>
       <%= render 'collections/show_descriptions' %>
-      <% end %>
-    </div>
-    <div class="col-sm-2">
-      <%= render 'collections/media_display' %>
-      <% unless has_collection_search_parameters? %>
-      <%= render partial: 'collections/show_actions' %>
-      <% end %>
-    </div>
+    <% end %>
   </div>
-
-<% if has_collection_search_parameters? %>
-    <% header_title = "Search Results within this Collection" %>
-<% else %>
-    <% header_title = "Items in this Collection" %>
-<% end %>
-
-<div style="float: right">
-<%= link_to t('sufia.collection.select_works'), sufia.dashboard_works_path %> | <%= link_to t('sufia.collection.add_work'), new_curation_concerns_generic_work_path %>
+  <div class="col-sm-2">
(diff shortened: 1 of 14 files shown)


spec/views/collections/show.html.erb_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+    stub_template 'collections/show_descriptions' => 'descriptions'
+    stub_template 'collections/media_display' => 'media display'
+    stub_template 'search_form' => 'search form'
+    stub_template 'sort_and_per_page' => 'sort and per page'
+    stub_template 'document_list' => 'document list'
+    stub_template 'paginate' => 'paginate'
+    render
+  end
+
+  it 'displays an actions header' do
```
jcoyne: You know that it's rendering separately for each `it`  block, right? That can slow down the test suite. Can we combine the checks into `it 'draws the stuff on the page'`

spec/views/collections/show.html.erb_spec.rb
```diff
@@ -0,0 +1,40 @@
+describe 'collections/show.html.erb' do
+  let(:document) { SolrDocument.new({ id: 'xyz123z4', 'title_tesim' => ['Make Collections Great Again'] }) }
```
jcoyne: Rubocop doesn't like this line.

spec/views/collections/show.html.erb_spec.rb
mjgiarlo: Yeah, but right now this is all failing and I've got to abandon this branch possibly 'til after OR.

awead:
There's a Travis error: `rspec ./spec/features/ownership_transfer_spec.rb:75 # Transferring work ownership: When someone requests a work transfer to me I should receive a notification` which is intermittent.

mjgiarlo:
I believe we've seen that error before in Travis and it's unrelated to this PR. It's annoying, I'll grant you. I've never been able to repro it locally.

app/views/collections/show.html.erb
```diff
@@ -1,35 +1,32 @@
...
   </div>
-
-<% if has_collection_search_parameters? %>
-    <% header_title = "Search Results within this Collection" %>
-<% else %>
-    <% header_title = "Items in this Collection" %>
-<% end %>
-
-<div style="float: right">
-<%= link_to t('sufia.collection.select_works'), sufia.dashboard_works_path %> | <%= link_to t('sufia.collection.add_work'), new_curation_concerns_generic_work_path %>
```
mjgiarlo: We're dropping these links because we already have buttons elsewhere for selecting works and adding new works?

config/locales/sufia.en.yml
```diff
@@ -196,10 +196,25 @@ en:
...
-      add_work:        "Add Work"
-      select_works:    "Select Works"
+      browse_view: "Browse View"
+      actions:
+        header: "Actions"
+        edit:
+          label: "Edit"
+          desc: "Edit this collection"
+        add_works:
+          label: "Add works"
```
mjgiarlo: The ticket asks for this to be worded "Add new work".

config/locales/sufia.en.yml
```diff
@@ -196,10 +196,25 @@ en:
...
+          desc: "Add works to this Collection"
```
mjgiarlo: Same as above re: "Add new work to..."

app/views/collections/show.html.erb
jcoyne: They were added in Dec, but we don't have them in Sufia 6 and we already have those actions in the `show_actions` partial.

mjgiarlo: OK.

config/locales/sufia.en.yml
jcoyne: I think we should reconsider that, it's not a new work, it's existing works

PR #2203, opened 2016-06-10
Break locales out into their own files.

mjgiarlo:
Makes it easier to find and read i18n keys.

### DIFF
--- a/config/locales/blacklight.en.yml
+++ b/config/locales/blacklight.en.yml
@@ -1,0 +1,6 @@
+en:
+  blacklight:
+    search:
+      fields:
+        show:
+          keyword: Keyword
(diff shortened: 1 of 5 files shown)


jcoyne:
This is not easier for me.

mjgiarlo:
@jcoyne Really? It makes sense to me to have one top-level namespace in the locale file and for that to correspond to the name of the file. That way I can know where I am generally without having to scroll up much and scan for whitespace to see how indented a key is.

jcoyne:
Yes, my editor can just collapse the YAML blocks, so it's all nicely organized in on place. Now it's spread out over N files.

jcoyne:
http://smartic.us/2009/04/06/code-folding-in-vim/

jcoyne:
screen shot 2016-06-10 at 2 22 27 pm

mjgiarlo:
I've got folding too, but am still offended by the length of that locale file. ;)

No biggie.

PR #2206, opened 2016-06-10
Add Works link on collection show page should be a button

mjgiarlo:
Fixes a recently introduced UI bug.


mjgiarlo:
Failure unrelated to PR.

PR #2209, opened 2016-06-13
Merge Sufia configuration with the CurationConcerns

### LINKED ISSUE #2139: Generator should remove config/initializer/curation_concerns.rb
But we must call
```ruby
CurationConcerns.configure.register_curation_concern :generic_work
```
in some way.  Perhaps Sufia config can delegate that method to CurationConcerns?

jcoyne:

This reduces confusion over which configuration someone should be using.
Fixes #2139

### DIFF
--- a/lib/generators/sufia/templates/config/sufia.rb
+++ b/lib/generators/sufia/templates/config/sufia.rb
@@ -1,16 +1,11 @@
 Sufia.config do |config|
-  config.max_days_between_audits = 7
-  config.max_notifications_for_dashboard = 5
+  config.register_curation_concern :generic_work
 
-  config.permission_levels = {
-    "Choose Access" => "none",
-    "View/Download" => "read",
-    "Edit" => "edit"
-  }
+  # How many notifications should be displayed on the dashboard
+  # config.max_notifications_for_dashboard = 5
 
-  config.owner_permission_levels = {
-    "Edit" => "edit"
-  }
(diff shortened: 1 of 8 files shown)


lib/generators/sufia/templates/config/sufia.rb
```diff
@@ -1,16 +1,11 @@
 Sufia.config do |config|
-  config.max_days_between_audits = 7
-  config.max_notifications_for_dashboard = 5
+  config.register_curation_concern :generic_work
 
-  config.permission_levels = {
-    "Choose Access" => "none",
-    "View/Download" => "read",
-    "Edit" => "edit"
-  }
```
awead: do these just not do anything anymore?

awead:
@jcoyne looks good. Do you want me to merge https://github.com/projecthydra/curation_concerns/pull/848 first so you can go ahead and remove that other stuff in this PR?

lib/generators/sufia/templates/config/sufia.rb
jcoyne: I just moved them into Configuration like all the other defaults.

jcoyne:
Naw, we can merge this now and defer removing those methods.

awead:
Works for me. Travis is complaining, but that error appears to be intermittent. I wasn't able to reproduce it locally.

PR #2212, opened 2016-06-13
Use the standard Rails way of doing migrations

jcoyne:


lib/generators/sufia/install_generator.rb
```diff
@@ -44,22 +37,7 @@ def run_curation_concerns_work_generator
...
-        'add_linkedin_to_users.rb',
-        'create_tinymce_assets.rb',
-        'create_content_blocks.rb',
-        'create_featured_works.rb',
-        'add_external_key_to_content_blocks.rb',
-        'create_uploaded_files.rb'
-      ].each do |file|
-        better_migration_template file
-      end
+      rake 'sufia:install:migrations'
```
cam156: @jcoyne at one point we were putting the names in so that we could guarantee an order.  Is that no longer necessary?

cam156:
@jcoyne I like the idea of doing this.  Will this change work well for existing users of Sufia 6 or Sufia 7?  We have modified migrations in the past and that has been a big pain for us to figure out what migrations got modified and what migrations we need to add to adapt to the new ones...

lib/generators/sufia/install_generator.rb
jcoyne: The timestamps guarantee an order.

jcoyne:
> Will this change work well for existing users of Sufia 6 or Sufia 7

I can't speak to 6. I don't see it being any better or worse than before. For Sufia 7 users, there are none. It hasn't been released yet.

lib/generators/sufia/install_generator.rb
cam156: @jcoyne But our classes are not timestamped before the migrations are generated.  The order is guaranteed once the app has been generated but what order do the migrations get created in?

jcoyne: They are now timestamped in db/migrations

PR #2214, opened 2016-06-13
Turn off javascript on intermittently failing test.

jcoyne:

Recently this test started failing frequently. However, turning off javascript (and slightly modifying the test) seems to prevent it from failing.


jcoyne:
I suspect we're hitting problems with form validation and PhantomJS, since the title of the collection was just marked as required.  See https://github.com/ariya/phantomjs/issues/11809

PR #2217, opened 2016-06-14
Expose an API for reading data

jcoyne:
This allows you to fetch RDF by setting the `Accept` request headers to an RDF type on an object show page.

@tpendragon should we push this down to curation_concerns?

### DIFF
--- a/app/connections/clean_connection.rb
+++ b/app/connections/clean_connection.rb
@@ -1,0 +1,23 @@
+# This stands in for ActiveFedora::CleanConnection. It behaves the same way,
+# but it doesn't clear the has_model assertion
+class CleanConnection < SimpleDelegator
+  def get(*args)
+    result = __getobj__.get(*args) do |req|
+      prefer_headers = Ldp::PreferHeaders.new(req.headers["Prefer"])
+      prefer_headers.omit = prefer_headers.omit | omit_uris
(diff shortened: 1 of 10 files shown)


app/connections/clean_connection.rb
```diff
@@ -0,0 +1,23 @@
+# This stands in for ActiveFedora::CleanConnection. It behaves the same way,
```
cbeer: Why does `ActiveFedora::CleanConnection` need to clear `hasModel`? Looking at https://github.com/projecthydra/active_fedora/commit/dffef733b0d9e952972353edf9f69375ad1bc57e, it seems like it was added to support this feature, so it's a little funny we're actively opting out.

jcoyne: I think @tpendragon wrote the original in ActiveFedora so that a resource could be displayed in the catalog controller.  In that case he didn't want to expose has_model and he didn't need it to figure out the route (all route to `/catalog/{id}`)

mjgiarlo:
Waiting to hear back from @tpendragon before merging.

tpendragon:
I'd be :+1: to having this in CC, but this isn't totally useful because it doesn't embed the ordered statements in the list_source contained resource. I don't have a clean solution for that though, so...

jcoyne:
Moved to https://github.com/projecthydra/curation_concerns/pull/850

PR #2220, opened 2016-06-14
Add ResourceSync resource list capability

jcoyne:

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,6 +1,7 @@
 inherit_from: .rubocop_todo.yml
 
 AllCops:
+  TargetRubyVersion: 2.1
   DisplayCopNames: true
   Include:
     - '**/Rakefile'
--- a/app/controllers/sufia/resource_sync_controller.rb
+++ b/app/controllers/sufia/resource_sync_controller.rb
@@ -1,0 +1,41 @@
+class Sufia::ResourceSyncController < ApplicationController
+  def source_description
+    # Caching based on host, for multitenancy support
+    body = Rails.cache.fetch("source_description_#{request.host}", expires_in: 1.week) do
+      build_source_description
+    end
+    render body: body, content_type: 'application/xml'
+  end
+
+  def capability_list
+    # Caching based on host, for multitenancy support
+    body = Rails.cache.fetch("source_description_#{request.host}", expires_in: 1.week) do
+      build_capability_list
+    end
+    render body: body, content_type: 'application/xml'
+  end
+
+  def resource_list
+    # Caching based on host, for multitenancy support
+    body = Rails.cache.fetch("source_description_#{request.host}", expires_in: 1.week) do
+      build_resource_list
+    end
+    render body: body, content_type: 'application/xml'
+  end
+
+  private
+
+    def build_resource_list
+      Sufia::ResourceSync::ResourceListWriter.new(capability_list_url: sufia.capability_list_url,
+                                                  resource_host: request.host).write
+    end
+
+    def build_capability_list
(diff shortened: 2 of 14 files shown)


.rubocop.yml
```diff
@@ -1,6 +1,7 @@
 inherit_from: .rubocop_todo.yml
 
 AllCops:
+  TargetRubyVersion: 2.1
```
awead: Why target 2.1 when we're testing on 2.2?

lib/sufia/resource_sync/resource_list_writer.rb
```diff
@@ -0,0 +1,71 @@
...
+            build_resource(xml, doc)
+          end
+        end
+
+        def build_resource(xml, doc)
+          xml.url do
+            key = doc.fetch('has_model_ssim', []).first.constantize.model_name.singular_route_key
+            xml.loc routes.send(key + "_url", doc['id'], host: resource_host)
+            xml.lastmod doc['system_modified_dtsi']
+            # xml['rs'].md hash: '', length: '', type: ''
```
awead: remove?

app/controllers/sufia/resource_sync_controller.rb
```diff
@@ -0,0 +1,44 @@
+class Sufia::ResourceSyncController < ApplicationController
+  def source_description
+    host = request.host
+    # Caching based on host, for multitenancy support
+    body = Rails.cache.fetch("source_description_#{host}", expires_in: 1.week) do
```
awead: Does using the local `host` variable do anything versus just calling `request.host` directly, or is that related to the cache?

.rubocop.yml
jcoyne: It's a language spec, not an implementation spec. I just bumped it from 2.0

app/controllers/sufia/resource_sync_controller.rb
jcoyne: no, nothing directly.

.rubocop.yml
awead: Then why not bump to 2.2 to match Travis?

app/controllers/sufia/resource_sync_controller.rb
awead: then you could remove the `host` variable and use `"source_description_#{request.host}"`

.rubocop.yml
jcoyne: Travis is testing an implemented version of Ruby.  This is the spec for the language spec. They aren't the same thing. We aren't using any language features of the 2.2+ ruby language spec. This calls that out.

app/controllers/sufia/resource_sync_controller.rb
jcoyne: Yes I could. What's the benefit?

awead: fewer lines of code to read and execute.

jcoyne: There's a reason we don't write all the code on one line (we could), it's so that it's more readable.

awead: Yes, that's true, but I don't think assigning the extra variable makes it more readable. It think it's the opposite, in fact, because now I have to mentally keep track of `host`. Whereas using the provided `request` object seems to be the Rails way of using what's provided to us, and doesn't extend the line to any extraordinary length beyond what it already is.

You also do this three times in the class, once for each public method, which to me just yells "refactor."

awead:
@jcoyne thanks for the edit. Feel free to bust my chops on my next PR.

jcoyne:
https://media.giphy.com/media/10rsLtGrOcCR0s/giphy.gif

PR #2228, opened 2016-06-15
Update contact form and tests, fixes #2225

### LINKED ISSUE #2225: Contact form fails to send
### Descriptive summary

Contact form fails to send because these config items are missing:

https://github.com/projecthydra/sufia/blob/master/app/models/contact_form.rb#L24-25

Should we add dummy values into the generator or provide some kind of documentation?

awead:
Fixes #2225 

* Remove enable_contact_form_delivery in config
* fail better in ContactFormControllerBehavior
* use more accurate controller tests instead of features

### DIFF
--- a/app/controllers/concerns/sufia/contact_form_controller_behavior.rb
+++ b/app/controllers/concerns/sufia/contact_form_controller_behavior.rb
@@ -17,13 +17,15 @@ module Sufia
         flash.now[:error] << @contact_form.errors.full_messages.map(&:to_s).join(",")
       end
       render :new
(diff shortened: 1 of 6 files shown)


jcoyne:
👍 for removing all the variations in the feature test. But shouldn't we have at least 1 feature test for "contact form"?

awead:
@jcoyne Yeah, we could keep it simple... just one complete submission, since the controller test covers all the variations.

awead:
@jcoyne I've added the feature test.

PR #2233, opened 2016-06-16
Remove the unused DirectoryController

jcoyne:


awead:
This is a Scholarsphere-only thing, I think. We just override the whole controller, so removing it here should be fine. @cam156 thoughts?

cam156:
@awead Looks correct to me.

PR #2234, opened 2016-06-16
Remove has_access? method

jcoyne:
This was not doing anything.  If this functionality is needed it should
be handled by CanCan

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -38,7 +38,6 @@ Style/LineEndConcatenation:
 Style/PredicateName:
   Exclude:
     - 'app/helpers/sufia/sufia_helper_behavior.rb'
-    - 'app/controllers/concerns/sufia/controller.rb'
 
 Rails/Output:
   Exclude:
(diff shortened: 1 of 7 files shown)


awead:
This is doing something for Scholarsphere:

https://github.com/psu-stewardship/scholarsphere/blob/develop/app/controllers/application_controller.rb#L85

But it's more about negotiating access via Devise. We can remove it here, but those upgrading from Sufia 6 will need to be aware.

@cam156 thoughts?

jcoyne:
@awead is there some reason you can't negotiate access using CanCan?

cam156:
@awead Can you add a ticket to ScholarSphere to add the has_access into the overrides and reference this PR.  That way it will not get lost.

jcoyne:
For Scholarsphere it may be better to override the Devise sessions controller and put an `after_filter :has_access?, only: :create`

awead:
We're using it to check if the authenticated user exists in LDAP, so it ought to be a `before_filter`. So I guess we can do that with CanCan: `can :has_ldap`? But what I'm wondering is why this kind of check isn't earlier on, in the Devise::Behaviors level. Seems like we should immediately deny any authenticated user that's not in LDAP.

awead:
See https://github.com/psu-stewardship/scholarsphere/issues/315

PR #2235, opened 2016-06-16
Use a Logger that writes to STDOUT rather than puts

jcoyne:

### DIFF
--- a/app/services/sufia/user_stat_importer.rb
+++ b/app/services/sufia/user_stat_importer.rb
@@ -5,7 +5,11 @@ module Sufia
     UserRecord = Struct.new("UserRecord", :id, :user_key, :last_stats_update)
 
     def initialize(options = {})
-      @verbose = options[:verbose]
+      if options[:verbose]
+        stdout_logger = Logger.new(Logger.new(STDOUT))
+        stdout_logger.level = Logger::INFO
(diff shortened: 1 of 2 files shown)


app/services/sufia/user_stat_importer.rb
```diff
@@ -5,7 +5,11 @@ class UserStatImporter
     UserRecord = Struct.new("UserRecord", :id, :user_key, :last_stats_update)
 
     def initialize(options = {})
-      @verbose = options[:verbose]
+      if options[:verbose]
```
cbeer: This doesn't strike me as appropriate for `UserStatImporter` (especially extending `Rails.logger` for everyone..). I'd feel better if it were pushed to the rake task, but maybe there's an even better way (say, `UserStatImporter#logger` that can switch between a stdout logger and the ordinary rails logger?)

jcoyne: This is only if they specify `verbose`.  I'm using the pattern that is in Rails itself:  https://github.com/rails/rails/blob/a26a3a075637215c9028308436ca89cba8da2ed5/railties/lib/rails/commands/server.rb#L123-L133

mjgiarlo: I'm :ok: with this.

Issue #2082, opened 2016-05-24
Support nested works


mjgiarlo:
### Descriptive summary

This is an epic ticket wrapping up all tickets related to created arbitrarily nested works in Sufia. Please link other tickets related to this feature below.
- [ ] https://github.com/projecthydra/sufia/issues/1709
- [ ] https://github.com/projecthydra/sufia/issues/1563
- [ ] https://github.com/projecthydra/sufia/issues/1710
- [x] https://github.com/projecthydra/sufia/issues/1861
- [x] https://github.com/projecthydra/sufia/issues/1970
- [x] https://github.com/projecthydra/sufia/issues/2111
- [x] #2197
- [x] #2243

mjgiarlo:
@eliotjordan is interested in this work and how it might connect to existing work in GeoConcerns.

@tpendragon has done some related work in Plum that IIRC will be extracted to CurationConcerns. (We should link that in here once it's ready.)

@vantuyls ping.

vantuyls:
pinging @zhang4952

mjgiarlo:
Closing this epic.

Issue #1710, opened 2016-03-30
Add contained works (members)


jcoyne:
### Descriptive summary

Add works that are members of the work currently being edited.
- [in context of New Work](https://wiki.duraspace.org/display/hydra/New+Work+-+tab+Relationships+-+Phase+1+-+UI+Wireframes) - Relationships Tab - contents of tab
- [in context of Edit Work](https://wiki.duraspace.org/display/hydra/Edit+Work+-+Relationships+tab+-+Phase+1+-+UI+Wireframes) - Relationships Tab - contents of tab

Closely related to #1709 
Ref #1577

jcoyne:
@mjgiarlo backlog?

jcoyne:
@dlpierce can you give us an update on the status of this work?

dlpierce:
I've been working on it when able, but it's not ready yet. I'll try to have a branch pushed this week.

mjgiarlo:
Moved to Hyrax.

Issue #1709, opened 2016-03-30
Add work to another work (in_works)


jcoyne:
### Descriptive summary

Add the work currently being edited as a member of other works.
- [in context of New Work](https://wiki.duraspace.org/display/hydra/New+Work+-+tab+Relationships+-+Phase+1+-+UI+Wireframes) - Relationships Tab - contents of tab
- [in context of Edit Work](https://wiki.duraspace.org/display/hydra/Edit+Work+-+Relationships+tab+-+Phase+1+-+UI+Wireframes) - Relationships Tab - contents of tab

Closely related to #1710 
Ref #1577

jcoyne:
@mjgiarlo backlog?

jcoyne:
@dlpierce can you give us an update on the status of this work?

mjgiarlo:
@dlpierce We're 1-2 blockers away from cutting a beta/RC for 7.3.0. Do you think you might have time to wrap this and #1710 in the coming week or two?

dlpierce:
@mjgiarlo Yes, I'll make it a priority to get this done for 7.3.0.

mjgiarlo:
Moved to Hyrax.

Issue #1563, opened 2016-02-09
UI Create _find_work_widget partial


elrayle:
### Wireframes:
- [in context of New Work](https://wiki.duraspace.org/display/hydra/New+Work+-+tab+Relationships+-+Phase+1+-+UI+Wireframes#mockup_p1-work-new-tab_relationships) - Relationship Tab - sections This Work in Other Works, Other Works in This Work, Related Works
- [in context of Edit Work](https://wiki.duraspace.org/display/hydra/Edit+Work+-+Relationships+Tab+-+Phase+1+-+UI+Wireframes) - Relationship Tab - sections This Work in Other Works, Other Works in This Work, Related Works
- [in context of Batch Upload](https://wiki.duraspace.org/display/hydra/Batch+New+Works+-+tab+Relationships+-+Phase+1+-+UI+Wireframes) - Relationship Tab - sections These New Works in Other Works, Other Works in These New Works, Related Works
### Usage:

Used to find and select an existing work.  Options for finding a work:
- user begins typing the name of the work and auto-complete suggests works
- user chooses to search for works and can browse a list of all works with pagination
### Implementation:
- File: /app/views/generic_works/_find_work_widget.html.erb
- Create a partial for a view that includes the find_work_widget.
  - need to be able to name the field associated with the widget
  - need to be able to return a value for the field based on the user's selection
- Add class/methods that support auto-complete
- Add class/methods that support browse and select

jcoyne:
@mjgiarlo backlog this?

andjsmit:
For the autocomplete: Would the right/desired approach be to create new local authority which returns the available works? Something like `Qa::Authorities::Local::FindWorks`. This class could then do a Solr query to return available works.

jcoyne:
Sounds like a good idea to me.

andjsmit:
After hitting some road blocks, looks like QA is not the way to go after all.  In search of new direction.

jcoyne:
@andjsmit why would you not use QA for this?

tap87:
What is the status of work on this issue? Our team is interested in this feature as well. @andjsmit @dlpierce

andjsmit:
There is working code, but I have so far not been able (lack of experience) to write required tests to push code through to a merge. Hope to revamp code to latest master branch and then look for assistance it creating proper test.

tap87:
Let me know if you need a hand with that, I can put in some time this week

andjsmit:
[...]

PR #2243, opened 2016-06-16
Enables Works within Works and related UI and functionality

### LINKED ISSUE #1861: Work show page -- re-label text if no relationships exist
### Descriptive summary

Change "no relationships" text to "No relationships to collections, works, or external resources." in the relationships section

Pertains to Sufia master.

### Expected UI

https://wiki.duraspace.org/x/DAF6B

### Related work

Refs #1722, #1843

### LINKED ISSUE #2111: Verify contextual nested works don't break the UI
Now that CurationConcerns has nested works, we should see how this affects the Sufia UI and make sure it doesn't cause any breakages or unexpected behavior.

### LINKED ISSUE #2197: Remove/rework view overrides to support contextual nested works from CC
Now that CurationConcerns has contextual nested works (https://github.com/projecthydra/curation_concerns/pull/829), we should change any view overrides that prevent Sufia from making use of this code:

- [ ] Include the cc/base/form_in_works partial in our cc/base/form override
- [ ] Merge changes made to CC/base/show_actions partial in our override
- [ ] Merge changes made to CC/base/show template in our override
- [ ] More :question:

revgum:
Fixes #1861, #2197, #2111 
Refs #1710, #1709, #1563, #2082 

A handful of folks at OSU (Greg, Brandon, Steve, Josh, Mike, Ryan) have been working on integrating curation concerns nested works into sufia. We intend on following this PR up with another (couple?) to address more functionality and UI described by the wireframes linked in tickets 1710, 1709, and 1563. The goal for this PR is to address some of the baseline functionality of adding new child/parent works in the edit form and showing links to existing child/parent works in the show views.

We would like to have the full team at OSU to join and discuss in greater detail on the Hydra Tech call on Wednesday June 22nd if possible. In the meantime, we're looking forward to your feedback!

Thanks,
OSULP

### DIFF
--- a/app/assets/javascripts/sufia/app.js
+++ b/app/assets/javascripts/sufia/app.js
@@ -9,6 +9,7 @@ Sufia = {
         this.permissions();
         this.notifications();
         this.transfers();
+        this.relationships_table();
     },
 
     autocomplete: function () {
@@ -58,6 +59,13 @@ Sufia = {
 
     transfers: function () {
         $("#proxy_deposit_request_transfer_to").userSearch();
+    },
+
+    relationships_table: function () {
+        var rel = require('sufia/relationships/table');
+        $('table.relationships-ajax-enabled').each(function () {
+            new rel.RelationshipsTable($(this));
+        });
     }
 };
 
--- a/app/assets/javascripts/sufia/relationships/table.es6
+++ b/app/assets/javascripts/sufia/relationships/table.es6
@@ -1,0 +1,206 @@
+import { RelationshipsTableRow } from './table_row'
+
+export class RelationshipsTable {
+
+  /**
+   * Initializes the class in the context of an individual table element
+   * @param {jQuery} element the table element that this class represents
+   */
+  constructor(element) {
+    this.$element = element;
+    this.form_action = this.$element.parents("form").attr("action");
+    this.query_url = this.$element.data('query-url');
+    this.existing_related_works_values = this.$element.find("input.related_works_ids:not(.new-form-control)").map(function(i, e){ return e.value; });
+
+    // TODO: Fall back to just cloning rows and removing rows for form posting
+    if (!this.query_url)
+      return;
+
+    this.bindAddButton();
+    this.bindRemoveButton();
+    this.bindKeyEvents();
+  }
+
+  /**
+   * Handle click events by the "Add" button in the table, setting a warning
+   * message if the input is empty or calling the server to handle the request
+   */
+  bindAddButton() {
+    let $this = this;
+
+    $this.$element.on("click", ".btn-add-row", function(event) {
+      let $row = $(this).parents("tr:first");
+      let $input = $row.find("input.new-form-control");
+
+      // Display an error when the input field is empty, or if the work ID is already related,
+      // otherwise clone the row and set appropriate styles
+      if ($input.val() == "") {
+        $this.setWarningMessage($row, "ID cannot be empty.");
+      } else if ($.inArray($input.val(), $this.existing_related_works_values) > -1) {
+        $this.setWarningMessage($row, "Work is already related.");
+      } else {
+
+        let query_url = $this.query_url.replace('$id', $input.val());
+        $this.hideWarningMessage($row);
+        $this.callAjax({
+          row: $row,
+          table: $this.$element,
+          input: $input,
+          url: $this.form_action,
+          query_url: query_url,
+          on_error: $this.handleError,
+          on_success: $this.handleAddRowSuccess
+        });
+      }
+    });
+  }
+
+  /**
+   * Find and fire off the click event for the "Add" button
+   * @param {jQuery} $row the row containing the add button to click
+   */
+  clickAddbutton($row) {
+    $row.find(".btn-add-row").click();
+  }
+
+  /**
+   * Handle click events by the "Remove" buttons in the table, and calling the
+   * server to handle the request
+   */
+  bindRemoveButton() {
+    let $this = this;
+
+    $this.$element.on("click", ".btn-remove-row", function(event) {
+      let $row = $(this).parents("tr:first");
+      let $input = $row.find("input.related_works_ids:first");
+
+      // Track which input is attemping to be removed, to provide an easy way
+      // for the ajax call to exclude it when making the call to the server
+      $input.addClass("removing");
+      $this.callAjax({
+        row: $row,
+        table: $this.$element,
+        input: $input,
+        url: $this.form_action,
+        on_error: $this.handleError,
+        on_success: $this.handleRemoveRowSuccess
+      });
+    });
+  }
+
+  /**
+   * Handle keyup and keypress events at the form level to prevent the ENTER key
+   * from submitting the form. ENTER key within a relationships table should
+   * click the "Add" button instead. ESC key should clear the input and hide the
+   * error message.
+   */
+  bindKeyEvents() {
+    let $this = this;
+    let $form = this.$element.parents("form");
+
+    $form.on("keyup keypress", "input.related_works_ids", function(event) {
+      let $row = $(this).parents("tr:first");
+      let key_code = event.keyCode || event.which;
+
+      // ENTER key was pressed, wait for keyup to click the Add button
+      if (key_code === 13) {
+        if (event.type == "keyup") {
+          $this.clickAddbutton($row);
+        }
+        event.preventDefault();
+        return false;
+      }
+
+      // ESC key was pressed, clear the input field and hide the error
+      if (key_code === 27 && event.type == "keyup") {
+        $(this).val("");
+        $this.hideWarningMessage($row);
+      }
+    });
+  }
+
+  /**
+   * Set the warning message related to the appropriate row in the table
+   * @param {jQuery} $row the row containing the warning message to display
+   * @param {String} message the warning message text to set
+   */
+  setWarningMessage($row, message) {
+    $row.find(".message.has-warning").text(message).removeClass("hidden");
+  }
+
+  /**
+   * Hide the warning message on the appropriate row
+   * @param {jQuery} $row the row containing the warning message to hide
+   */
+  hideWarningMessage($row){
+    $row.find(".message").addClass("hidden");
+  }
+
+  /**
+   * Call the server, then call the appropriate callbacks to handle success and errors
+   * @param {Object} args the table, row, input, url, and callbacks
+   */
+  callAjax(args) {
+    let $this = this;
+    // Send only the IDs in this table that aren't in the midst of being "removed"
+    let data = args.table.find("input.related_works_ids:not('.removing')").serialize();
+    $.ajax({
+        type: 'patch',
+        url: args.url,
+        dataType: 'json',
+        data: data
+      })
+      .done(function(json) {
+        args.on_success($this, args, json);
+      })
+      .fail(function(jqxhr, status, err) {
+        args.on_error($this, args, jqxhr, status, err);
+      });
+  }
+
+  /**
+   * Set a warning message to alert the user on an error
+   * @param {jQuery} $this the RelationshipsTable class instance
+   * @param {Object} args the table, row, input, url, and callbacks
+   * @param {Object} jqxhr the jQuery XHR response object
+   * @param {String} status the HTTP error status
+   * @param {String} err the HTTP error
+   */
+  handleError($this, args, jqxhr, status, err) {
+    args.row.find('input.removing').removeClass('removing');
+    let message = jqxhr.statusText;
+    if(jqxhr.responseJSON){
+      message = jqxhr.responseJSON.description;
+    }
+    $this.setWarningMessage(args.row, message);
+  }
+
+  /**
+   * Remove the row when the API returns this type of success
+   * @param {jQuery} $this the RelationshipsTable class instance
+   * @param {Object} args the table, row, input, url, and callbacks
+   * @param {String} json the returned JSON string
+   */
+  handleRemoveRowSuccess($this, args, json) {
+    args.row.remove();
+  }
+
+  /**
+   * Add a new row to the table, query the server for details about the work to
+   * set the title and link for the new work that was added. Hide the input
+   * field and display the title and edit button
+   * @param {jQuery} $this the RelationshipsTable class instance
+   * @param {Object} args the table, row, input, url, and callbacks
+   * @param {String} json the returned JSON string
+   */
+  handleAddRowSuccess($this, args, json) {
+    let new_row = new RelationshipsTableRow(args.table);
+    new_row.clone(args.row);
+    new_row.callAjaxQuery(args.query_url);
+
+    // finally, empty the "add" row input value
+    args.row.find("input.new-form-control").val("");
+    // synch the related_works_values to include the new work relationship
+    $this.existing_related_works_values = $this.$element.find("input.related_works_ids:not(.new-form-control)").map(function(i, e){ return e.value; });
+  }
+}
--- a/app/presenters/sufia/work_show_presenter.rb
+++ b/app/presenters/sufia/work_show_presenter.rb
@@ -17,6 +17,17 @@ module Sufia
       end
     end
 
+    def presenter_types
+      CurationConcerns.config.registered_curation_concern_types.map(&:underscore) + ["collection"]
+    end
+
+    def grouped_presenters(filtered_by: nil, except: nil)
+      grouped = collection_presenters.group_by(&:model_name).transform_keys { |key| key.to_s.underscore }
+      grouped.select! { |obj| obj.downcase == filtered_by } unless filtered_by.nil?
+      grouped.except!(*except) unless except.nil?
+      grouped
+    end
+
     def display_feature_link?
       user_can_feature_works? && solr_document.public? && FeaturedWork.can_create_another? && !featured?
     end
--- a/app/views/curation_concerns/base/_form_child_work_relationships.html.erb
+++ b/app/views/curation_concerns/base/_form_child_work_relationships.html.erb
@@ -1,0 +1,65 @@
+<%# Form UI behavior code and details;
+Code:
+  app/assets/javascripts/sufia/relationships
+CSS:
+  table.relationships-ajax-enabled : Used by sufia JS app as selector to wire up the UI functionality
+  tr.new-row : The basic template row for cloning when user clicks "Add"
+  .btn-remove-row : Button to remove its parent TR from the table
+  .btn-add-row : Button to clone its parent TR and inject a new row into the table
+  input.new-form-control : Input field for client client side validation and additional features
+  .message.has-warning : Used to display UI errors related to input values and server errors
+HTML Properties:
+  table:
+    data-query-url : URL base to append newly added work ID to for querying details (title, href)
+%>
+<div class="form-group multi_value optional managed">
+  <%= link_to "Attach New Work", polymorphic_path([main_app, :new, :curation_concerns, :parent, curation_concern.model_name.singular], parent_id: params[:id]), target: "_blank", class: 'btn btn-primary' %>
+
+  <table class="table table-striped related-files relationships-ajax-enabled"
+         data-query-url="<%= polymorphic_path([main_app, :curation_concerns, curation_concern.model_name.singular], id: '$id') %>">
+    <thead>
+    <tr>
+      <th>Child Work</th>
+      <th>Actions</th>
+    </tr>
+    </thead>
+    <tbody>
+    <tr class="hidden">
+      <td>
+        <% f.object.ordered_fileset_members.each do |fileset| %>
+          <input class="related_works_ids work_filesets_ids" name="<%= f.object.model_name.param_key %>[ordered_member_ids][]" type="hidden" value="<%= fileset.id %>">
+        <% end %>
+      </td>
+      <td></td>
+    </tr>
+    <% f.object.ordered_work_members.each do |member| %>
+        <tr>
+          <td>
+            <%= link_to member.title.first, [main_app, member] %>
+            <input class="string multi_value optional form-control related_works_ids work_child_members_ids form-control multi-text-field hidden" value="<%= member.id %>" id="work_child_members_ids" aria-labelledby="work_child_members_ids_label" name="<%= f.object.model_name.param_key %>[ordered_member_ids][]" type="text">
+          </td>
+          <td>
+            <div class="child-actions">
+              <%= link_to "Edit", [main_app, :edit, member], target: "_blank", class: 'btn btn-default' %>
+              <a class="btn btn-danger btn-remove-row">Remove</a>
+            </div>
+          </td>
+        </tr>
+    <% end %>
+    <tr class="new-row">
+      <td>
+        <a href="" class="title hidden"></a>
+        <input class="new-form-control string multi_value optional related_works_ids work_child_members_ids form-control multi-text-field" value="" id="work_child_members_ids" aria-labelledby="work_child_members_ids_label" name="<%= f.object.model_name.param_key %>[ordered_member_ids][]" type="text">
+        <div class="message has-warning hidden"></div>
+      </td>
+      <td>
+        <div class="child-actions">
+          <a href="" class="edit hidden btn btn-default" target="_blank">Edit</a>
+          <a class="btn btn-danger btn-remove-row hidden">Remove</a>
+          <a class="btn btn-primary btn-add-row">Add</a>
+        </div>
+      </td>
+    </tr>
+    </tbody>
+  </table>
+</div>
--- a/app/views/curation_concerns/base/_form_relationships.html.erb
+++ b/app/views/curation_concerns/base/_form_relationships.html.erb
@@ -1,6 +1,15 @@
+<%= render 'form_in_works', f: f %>
 <h2><%= t("sufia.works.#{action_name}.in_collections") %></h2>
 <div id="collection-widget">
   <%= f.input :collection_ids, as: :select,
-      collection: available_collections(nil),
-      input_html: { class: 'form-control', multiple: true } %>
+              collection: available_collections(nil),
+              input_html: { class: 'form-control', multiple: true } %>
 </div>
+
+<% if params[:id] %>
+  <h3><%= t("sufia.works.#{action_name}.in_this_work") %></h3>
+  <%= render 'form_child_work_relationships', f: f %>
+
+  <h3><%= t("sufia.works.#{action_name}.in_other_works") %></h3>
+  <%= render_edit_field_partial(:in_works_ids, f: f) %>
+<% end %>
--- a/app/views/curation_concerns/base/_relationships.html.erb
+++ b/app/views/curation_concerns/base/_relationships.html.erb
@@ -1,22 +1,7 @@
-    <h2><%= t('.header') %></h2>
-    <% collection_presenters = presenter.collection_presenters %>
-    <% if collection_presenters.blank? %>
-        <p><%= t('.empty', type: presenter.human_readable_type) %></p>
-    <% else %>
-      <table class="table table-striped relationships">
-        <tbody>
-        <tr>
-          <th><%= t('.collections', type: presenter.human_readable_type) %></th>
-          <td>
-            <% collection_presenters.each do |collection| %>
-              <ul class="tabular">
-                <li class='attribute title'>
-                  <%= link_to collection.to_s, main_app.collection_path(collection) %>
-                </li>
-              </ul>
-            <% end %>
-          </td>
-        </tr>
-        </tbody>
-      </table>
-    <% end %>
+<h2><%= t('.header') %></h2>
+<table class="table table-striped relationships">
+  <tbody>
+    <%= render 'relationships_parent_rows', presenter: presenter %>
+    <%= render 'relationships_member_rows', presenter: presenter %>
+  </tbody>
+</table>
--- a/app/views/curation_concerns/base/_relationships_member_rows.html.erb
+++ b/app/views/curation_concerns/base/_relationships_member_rows.html.erb
@@ -1,0 +1,24 @@
+<% member_presenters = presenter.member_presenters.group_by(&:model_name).transform_keys{|key| key.to_s.singularize.underscore}.select{ |k,v| k.to_sym != :file_set } %>
+<% if member_presenters.blank? %>
+  <tr>
+    <th><%= t(".label", type: presenter.human_readable_type) %></th>
+    <td>
+      <p><%= t('.empty', type: presenter.human_readable_type) %></p>
+    </td>
+  </tr>
+<% else %>
+  <% member_presenters.each_pair do |model_name, members| %>
+    <tr>
+      <th><%= t(".label", type: presenter.human_readable_type) %></th>
+      <td>
+        <ul class="tabular">
+          <% members.each do |member| %>
+            <li class='attribute title'>
+              <%= link_to member.title.first, [main_app, member] %>
+            </li>
+          <% end %>
+        </ul>
+      </td>
+    </tr>
+  <% end %>
+<% end %>
--- a/app/views/records/edit_fields/_in_works_ids.html.erb
+++ b/app/views/records/edit_fields/_in_works_ids.html.erb
@@ -1,0 +1,59 @@
+<%# Form UI behavior code and details;
+Code:
+  app/assets/javascripts/sufia/relationships
+CSS:
+  table.relationships-ajax-enabled : Used by sufia JS app as selector to wire up the UI functionality
+  tr.new-row : The basic template row for cloning when user clicks "Add"
+  .btn-remove-row : Button to remove its parent TR from the table
+  .btn-add-row : Button to clone its parent TR and inject a new row into the table
+  input.new-form-control : Input field for client client side validation and additional features
+  .message.has-warning : Used to display UI errors related to input values and server errors
+HTML Properties:
+  table:
+    data-query-url : URL base to append newly added work ID to for querying details (title, href)
+%>
+<% if f.object.class.multiple? key %>
+  <div class="form-group multi_value optional work_in_works_ids managed">
+    <table class="table table-striped related-files relationships-ajax-enabled"
+           data-query-url="<%=polymorphic_path([main_app, :curation_concerns, curation_concern.model_name.singular], id: '$id') %>">
+      <thead>
+        <tr>
+          <th>Parent Work</th>
+          <th>Actions</th>
+        </tr>
+      </thead>
+      <tbody>
+        <% curation_concern.in_works.each do |parent| %>
+          <tr>
+            <td>
+              <%= link_to "#{parent.title.first}", polymorphic_path([main_app, :curation_concerns, curation_concern.model_name.singular], id: parent.id) %>
+              <input class="string multi_value optional form-control related_works_ids form-control multi-text-field hidden" name="<%= curation_concern.model_name.param_key %>[in_works_ids][]" value="<%= parent.id %>" id="work_in_works_ids" aria-labelledby="work_in_works_ids_label" type="text">
+            </td>
+            <td>
+              <div class="child-actions">
+                <%= link_to "Edit", edit_polymorphic_path([main_app, :curation_concerns, curation_concern.model_name.singular], id: parent.id), target: "_blank", class: 'btn btn-default' %>
+                <a class="btn btn-danger btn-remove-row">Remove</a>
+              </div>
+            </td>
+          </tr>
+        <% end %>
+        <tr class="new-row">
+          <td>
+            <a href="" class="title hidden"></a>
+            <input class="new-form-control string multi_value optional related_works_ids form-control multi-text-field" name="<%= curation_concern.model_name.param_key %>[in_works_ids][]" value="" aria-labelledby="work_in_works_ids_label" type="text">
+            <div class="message has-warning hidden"></div>
+          </td>
+          <td>
+            <div class="child-actions">
+              <a href="" class="edit hidden btn btn-default" target="_blank">Edit</a>
+              <a class="btn btn-danger btn-remove-row hidden">Remove</a>
+              <a class="btn btn-primary btn-add-row">Add</a>
+            </div>
+          </td>
+        </tr>
+      </tbody>
+    </table>
+  </div>
+<% else %>
+  <%= f.input key, required: f.object.required?(key) %>
+<% end %>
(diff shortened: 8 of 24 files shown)


app/assets/javascripts/sufia/form_add_remove_table_rows.js
```diff
@@ -0,0 +1,200 @@
+$(document).ready(function() {
```
jcoyne: This is not going to work with Turbolinks. Use Blacklight.onLoad instead.

app/assets/javascripts/sufia/form_add_remove_table_rows.js
```diff
@@ -0,0 +1,200 @@
...
+
+  // Add an empty new row to the table to take input for a new ID, when the form is posted
+  // to the server a new relationship will be established.
+  $(document).on("click", ".btn-add-row", function(event) {
```
jcoyne: I'd really like to see this whole js file made into something like a jquery plugin (scoped to a particular widget on the page) rather than a bunch of global button handlers.

app/assets/javascripts/sufia/form_add_remove_table_rows.js
```diff
@@ -0,0 +1,200 @@
...
+      event.preventDefault();
+      return false;
+    }
+  });
+
+  // If a new row has been added, and the input field is emptied,
+  // then display the warning div. This isn't strictly necessary since the
+  // add/remove buttons perform API calls now.
+  /*
+  $(document).on("keyup", "input.related_works_ids", function(event) {
```
jcoyne: Should this be removed?

app/assets/javascripts/sufia/form_add_remove_table_rows.js
```diff
@@ -0,0 +1,200 @@
...
+}
+
+// Adding parent failed, display the proper alert message
+var apiAddParentError = function(row, jqxhr) {
+  setWarningMessage($(row), jqxhr.responseText);
+  console.log("apiAddParentError:", jqxhr);
+}
+
+// Removing parent succeeded, remove the row from the table
+var apiRemoveParentSuccess = function(row, json) {
```
jcoyne: Can we avoid adding functions to the root namespace?

app/controllers/api/generic_works_controller.rb
```diff
@@ -0,0 +1,92 @@
+module API
+  # Adds an endpoint that customizes the update action on nested works
+  class GenericWorksController < ApplicationController
+    before_action :my_load_and_authorize_resource, only: [:update, :destroy, :show, :add_parent, :remove_parent]
```
jcoyne: why are you bothering with the `only: ...` it looks like you've listed everything.

spec/javascripts/form_add_remove_table_row_spec.js.coffee
```diff
@@ -0,0 +1,72 @@
+describe 'form add remove table rows', ->
```
jcoyne: Why did you write your test in coffeescript, but the code was written in ES4?

app/assets/javascripts/sufia/form_add_remove_table_rows.js
revgum: Good call, I'll do some rework.

app/assets/javascripts/sufia/form_add_remove_table_rows.js
revgum: Thanks, wasn't aware of that!

app/assets/javascripts/sufia/form_add_remove_table_rows.js
revgum: Indeed. Lost sight of this..

app/controllers/api/generic_works_controller.rb
straleyb: So the only action that isn't in there is #create. Should we change it to 
```
before_action :my_load_and_authorize_resource, except: [:create]
```
as a better alternative?

.rubocop_todo.yml
```diff
@@ -89,6 +89,7 @@ RSpec/VerifiedDoubles:
 
 RSpec/AnyInstance:
   Exclude:
+    - 'spec/controllers/api/generic_works_controller_spec.rb'
```
jcoyne: Please fix the problem rather than ignore it.

config/routes.rb
```diff
@@ -155,12 +155,8 @@
     get 'generic_works/:id/remove_parent/:parent_id' => 'generic_works#remove_parent'
   end
 
-  if defined?(Sufia::StatsAdmin)
```
jcoyne: If you rebase your work, there should be no need for this commit.  It's already done in master.

app/assets/javascripts/sufia/app.js
```diff
@@ -46,6 +47,16 @@ Sufia = {
 
     transfers: function () {
         $("#proxy_deposit_request_transfer_to").userSearch();
+    },
+
+    relationships_table: function () {
+        var rel = require('sufia/relationships_table');
+        $('table.relationships-api-enabled').each(function () {
+            var add_url = $(this).data('api-add-url');
```
jcoyne: It appears that these 3 variables are assigned to but never accessed.

app/controllers/api/generic_works_controller.rb
jcoyne: Why wouldn't you want to authorize creating?

app/controllers/api/generic_works_controller.rb
```diff
@@ -0,0 +1,98 @@
+module API
+  # Adds an endpoint that customizes the update action on nested works
+  class GenericWorksController < ApplicationController
```
jcoyne: Why are we adding a GenericWorksController?  The future direction for sufia is to not use GenericWorks and use whatever work types the user wants to generate (like CurationConcerns).

.rubocop_todo.yml
revgum: Any hints on how to go about fixing it, I wasn't sure why this one was failing but the items_controller_spec and zotero_controller_spec were passing (as this controller was based on).. we found these were being ignored in the rubocop_todo as well. 

Are we missing a simple fix that we could implement into the item_controller_spec and zotero_controller_spec as well?

app/controllers/api/generic_works_controller.rb
straleyb: So I guess it was assumed that the functionality from another API controller was to be retained in this one, but that assumption could easily incorrect. 

https://github.com/projecthydra/sufia/blob/master/app/controllers/api/items_controller.rb#L9

we'll make the change.

app/assets/javascripts/sufia/app.js
revgum: They referenced from $this.add_url, $this.remove_url, and $this.works_context used in the ajaxy part of the functionality.

config/routes.rb
```diff
@@ -149,6 +149,12 @@
     end
   end
 
+  namespace :api do
+    get 'generic_works/:id' => 'generic_works#show'
+    get 'generic_works/:id/add_parent/:parent_id' => 'generic_works#add_parent'
```
tpendragon: I'm not sure custom routes just for adding a single parent are the best route. CC already has support for just sending a PATCH to the resource's controller with in_work_ids as a parameter - why not use that?

app/views/records/edit_fields/_in_works_ids.html.erb
```diff
@@ -0,0 +1,78 @@
...
+           data-works-context="parent">
+      <thead>
+        <tr>
+          <th>Parent Work</th>
+          <th>Actions</th>
+        </tr>
+      </thead>
+      <tbody>
+        <% curation_concern.in_works_ids.each do |parent_id| %>
+          <% parent = ::ActiveFedora::Base.find(parent_id) %>
```
tpendragon: This is going to be really expensive and there's no great way to abstract it out later if the logic's in the view. Maybe `f.object.in_works.each`?

app/controllers/api/generic_works_controller.rb
revgum: Our approach was to expose a simple add_parent and remove_parent action that responded with JSON, and it seemed like the API was the place to do it. What's a better/appropriate approach to do this sort of thing?

.rubocop_todo.yml
jcoyne: You can allow `allow(MyClass).to receive(:find).and_return(double)` or `allow(MyClass).to receive(:new).and_return(double)` and then call your expectation on that double.. You could do that with all the grandfathered exceptions if you wanted to.

app/controllers/api/generic_works_controller.rb
luisgreg99: Having the api for GenericWorksController also enabled us to implement the ajax requests, and the ability to update the relationships (adding children or parents) asynchronously within the edit form.

jcoyne: No problem with having an API. My issue is that it's specific to `GenericWork`s.  What if I add an `Image` type of work?

config/routes.rb
revgum: I think if we implement the front end UI to send a PATCH http request to the resources controller, then we don't need to think about implementing the API controller at all. Is that right? @jcoyne @tpendragon

app/controllers/api/generic_works_controller.rb
straleyb: Would a name change do? Something like WorkRelationshipsController, or are you describing a deeper change? Maybe generating this controller based on some rake task that generates `Image` types of works and the supporting files?

config/routes.rb
jcoyne: Yes, That sounds correct.

app/controllers/api/generic_works_controller.rb
jcoyne: Either of those are better than `GenericWork*`, although I wonder if we can't just put this in the resource controller.

straleyb: `ResourceSyncController` ? 

https://github.com/projecthydra/sufia/blob/master/app/controllers/sufia/resource_sync_controller.rb

jcoyne: No, I'm talking about https://github.com/projecthydra/sufia/blob/master/app/controllers/concerns/sufia/works_controller_behavior.rb

mjgiarlo: :+1: to that.

straleyb: Oh! I see! I was looking for the "Resource Controller" and couldn't find it. So I was confused what you were talking about. We actually thought about putting it there. We are looking into the patch fix for the routes, and if that works, it seems as though we can remove this module all together. (from what I understand :P)

luisgreg99:
### Screenshots
**Show page:**

image

**Edit page:**

image

app/controllers/api/generic_works_controller.rb
jcoyne: The API::ItemsController is for Arkivio integration only.

app/presenters/sufia/work_show_presenter.rb
```diff
@@ -17,6 +17,20 @@ def tweeter
       end
     end
 
+    def collection_presenter_types
+      [
+        "genericwork",
```
jcoyne: Assuming genericwork is not a valid assumption anymore.  There is no longer a GenericWork provided by Sufia.

app/views/curation_concerns/base/_form_child_work_relationships.html.erb
```diff
@@ -0,0 +1,75 @@
...
+          "id":"2b88qc32v"
+        }
+      }
+%>
+<div class="form-group multi_value optional managed">
+  <%= link_to "Attach New Work", polymorphic_path([main_app, :new, :curation_concerns, :parent, curation_concern.model_name.singular], parent_id: params[:id]), target: "_blank", class: 'btn btn-primary' %>
+
+  <table class="table table-striped related-files relationships-api-enabled"
+         data-api-add-url=<%="/api/generic_works/:id/add_parent/#{params[:id]}"%>
+         data-api-remove-url=<%="/api/generic_works/:id/remove_parent/#{params[:id]}"%>
```
jcoyne: I think you're going to have to update these routes when you change the controller name.

app/views/curation_concerns/base/_form_relationships.html.erb
```diff
@@ -1,6 +1,15 @@
-<h3><%= t("sufia.#{controller_name}.#{action_name}.in_collections") %></h3>
-<div id="collection-widget">
+<%= render 'form_in_works', f: f %>
+<h3>This Work in Collections</h3>
```
jcoyne: Why aren't we using `t("sufia.#{controller_name}.#{action_name}.in_collections")` anymore?

app/views/curation_concerns/base/_form_relationships.html.erb
```diff
@@ -1,6 +1,15 @@
...
+<div id="child_members">
   <%= f.input :collection_ids, as: :select,
       collection: available_collections(nil),
       input_html: { class: 'form-control', multiple: true } %>
 </div>
+
+<% if params[:id] %>
+  <h3>Other Works in This Work</h3>
```
jcoyne: Add to i18n

app/views/curation_concerns/base/_form_relationships.html.erb
```diff
@@ -1,6 +1,15 @@
...
+  <%= render 'form_child_work_relationships' %> 
+
+  <h3>This Work in Other Works</h3>
```
jcoyne: Add to i18n

app/views/curation_concerns/base/_relationships.html.erb
```diff
@@ -1,22 +1,7 @@
...
-            <% end %>
-          </td>
-        </tr>
-        </tbody>
-      </table>
-    <% end %>
+<h2><%= t('.header') %></h2>
+<table class="table table-striped relationships">
+  <tbody>
+    <%= render 'relationships_collection_rows', :presenter => presenter %>
```
jcoyne: Use ruby 1.9 hash syntax please.

app/views/records/edit_fields/_in_works_ids.html.erb
```diff
@@ -0,0 +1,77 @@
...
+          "title":["asdf"],
+          "path":"/concern/generic_works/2b88qc32v",
+          "id":"2b88qc32v"
+        }
+      }
+%>
+<% if f.object.class.multiple? key %>
+  <div class="form-group multi_value optional generic_work_in_works_ids managed">
+    <table class="table table-striped related-files relationships-api-enabled"
+           data-api-add-url=<%="/api/generic_works/#{f.object.model.id}/add_parent/:id"%>
```
jcoyne: These URIs will have to change

app/views/records/edit_fields/_in_works_ids.html.erb
```diff
@@ -0,0 +1,77 @@
...
+           data-api-remove-url=<%="/api/generic_works/#{f.object.model.id}/remove_parent/:id"%>
```
jcoyne: Why not just use `f.object.id`

config/routes.rb
```diff
@@ -149,6 +149,12 @@
...
+    get 'generic_works/:id/remove_parent/:parent_id' => 'generic_works#remove_parent'
```
jcoyne: You should use a `delete` action, not a `get` for request that removes a resource.

spec/javascripts/helpers/test_fixtures.js.coffee
```diff
@@ -0,0 +1,9 @@
+window.TestFixtures = {
+  form_add_remove_table_rows: {
+    child_id: 'child1234',
+    parent_id: 'parent5678'
+  }
+}
+
+window.TestFixtures.form_add_remove_table_rows["table_html"] = "<table class='table table-striped related-files relationships-api-enabled' data-works-context='parent' data-api-add-url='/api/generic_works/#{TestFixtures.form_add_remove_table_rows.child_id}/add_parent/:id' data-api-remove-url=data-api-url='/api/generic_works/#{TestFixtures.form_add_remove_table_rows.child_id}/remove_parent/:id'> <thead> <tr> <th>Parent Work</th> <th>Actions</th> </tr> </thead> <tbody> <tr class='new-row'> <td> <a href='' class='title hidden'></a> <input class='new-form-control string multi_value optional related_works_ids form-control multi-text-field' name='' value='' aria-labelledby='generic_work_in_works_ids_label' type='text'> <div class='message has-warning hidden'></div> </td> <td> <div class='child-actions'> <a href='' class='edit hidden btn btn-default' target='_blank'>Edit</a> <a class='btn btn-danger btn-remove-row hidden'>Remove</a> <a class='btn btn-primary btn-add-row'>Add</a></div></td></tr></tbody></table>"
```
jcoyne: Don't forget to change these URLs.

app/views/curation_concerns/base/_relationships_collection_row.html.erb
```diff
@@ -0,0 +1,16 @@
+<tr>
+<th><%= t(".#{model_name.downcase}", type: model_name) %></th>
```
jcoyne: Do you want to use `model_name.param_key` (e.g. `GenericWork.model_name.param_key`)?

spec/views/curation_concerns/base/_relationships.html.erb_spec.rb
```diff
@@ -14,10 +18,30 @@
...
+      expect(rendered).to match %r{This Work is not currently a part of any works\.}
+    end
+  end
+
+  context "when collections are present and no parents are present" do
+    let(:collection_presenters) { [collection] }
     let(:page) { Capybara::Node::Simple.new(rendered) }
     before do
+      allow(view).to receive(:contextual_path).and_return("/collections/456")
+      allow(model_col).to receive(:model_name).and_return("Collection")
```
jcoyne: the `model_name` method should return an `ActiveModel::Name` object, not a string

spec/views/curation_concerns/base/_relationships.html.erb_spec.rb
```diff
@@ -27,5 +51,74 @@
...
+      expect(page).to have_content "This Work is not currently a part of any works."
+    end
+  end
+
+  context "when parents are present and no collections are present" do
+    let(:collection_presenters) { [generic_work] }
+    let(:page) { Capybara::Node::Simple.new(rendered) }
+    before do
+      allow(view).to receive(:contextual_path).and_return("/concern/generic_works/456")
+      allow(model_gw).to receive(:model_name).and_return("GenericWork")
```
jcoyne: the `model_name` method should return an `ActiveModel::Name` object, not a string

spec/views/curation_concerns/base/_relationships.html.erb_spec.rb
```diff
@@ -27,5 +51,74 @@
...
+      expect(page).not_to have_content "This Work is not currently a part of any works."
+    end
+  end
+
+  context "when parents are present and collections are present" do
+    let(:collection_presenters) { [generic_work, collection] }
+    let(:page) { Capybara::Node::Simple.new(rendered) }
+    before do
+      allow(view).to receive(:contextual_path).and_return("/concern/generic_works/456")
+      allow(model_gw).to receive(:model_name).and_return("GenericWork")
```
jcoyne: the `model_name` method should return an `ActiveModel::Name` object, not a string

spec/views/curation_concerns/base/_relationships.html.erb_spec.rb
straleyb: Would that return something like look something like `allow(model_col).to receive(:model_name).and_return(double(ActiveModel::Name("collection)))`?

jcoyne: I have no idea. I'd suggest `...and_return(Collection.model_name)`

tpendragon:
I think I'm :+1: here. No controller changes makes me happy, and while I'm a little confused about what the JS is actually doing in some places, it seems to work well in the video. It's a little confusing to me when certain actions "take", like adding an ID (does that save immediately, or just when you hit the save button?), but those things can be worked out later.

mjgiarlo:
Thanks, @tpendragon!

@revgum @luisgreg99 @straleyb @wickr @vantuyls I'd like @jcoyne to comment as well when he is available given how involved he's been. I suspect we're not far off from merging this.

revgum:
Thanks @tpendragon @mjgiarlo! 

As far as the question about the JS, clicking "Add" or "Remove" serializes the ID's on the appropriate table (child / parent) then makes the PATCH call to the controller on the server. I added some bindings so that pressing <enter> on the field will essentially "click" the Add button, pressing <esc> clears the field and hides any visible error related to that table. In general though, the functionality is pretty basic but designed to make extending this into something a bit more featureful (search, paginated results, etc) not too complicated from the front-end perspective.

mjgiarlo:
@jcoyne Any final thoughts on this PR?

app/assets/javascripts/sufia/relationships_table.es6
```diff
@@ -0,0 +1,190 @@
...
+    let $form = this.$element.parents("form");
+
+    $form.on("keyup keypress", "input.related_works_ids", function(event) {
+      let $row = $(this).parents("tr:first");
+      let key_code = event.keyCode || event.which;
+
+      // ENTER key was pressed, wait for keyup to click the Add button
+      if (key_code === 13) {
+        if (event.type == "keyup") {
+          $row.find(".btn-add-row").click();
```
jcoyne: Can you encapsulate this in a method?

app/assets/javascripts/sufia/relationships_table.es6
```diff
@@ -0,0 +1,190 @@
...
+        }
+        event.preventDefault();
+        return false;
+      }
+
+      // ESC key was pressed, clear the input field and hide the error
+      if (key_code === 27 && event.type == "keyup") {
+        $(this).val("");
+        $row.find(".message").addClass("hidden");
```
jcoyne: Can you encapsulate this in a method?

app/assets/javascripts/sufia/relationships_table.es6
```diff
@@ -0,0 +1,190 @@
...
+  handleRemoveRowSuccess($this, args, json) {
+    args.row.remove();
+  }
+
+  // Add a new row to the table, query the server for details about the work to
+  // set the title and link for the new work that was added. Hide the input
+  // field and display the title and edit button
+  handleAddRowSuccess($this, args, json) {
+    let $clone = $this.cloneRow(args.row);
+    let $title = $clone.find("a.title:first");
```
jcoyne: can the DOM element be encapsulated by a new ES6 object? Can `cloneRow` be return a `RelationshipRow`?

app/assets/javascripts/sufia/relationships_table.es6
```diff
@@ -0,0 +1,190 @@
...
+    let $edit = $clone.find("a.edit:first");
+    let $input = $clone.find("input.related_works_ids:first");
+
+    $.getJSON(args.query_url, function(data) {
+      // Set the cloned input to have the proper name and value for posting the
+      // form to the server, and hide it.
+      $input.removeClass("new-form-control")
```
jcoyne: Have you considered putting this in a new function? That would make it more testable.

app/views/curation_concerns/base/_form_child_work_relationships.html.erb
```diff
@@ -0,0 +1,65 @@
...
+    <thead>
+    <tr>
+      <th>Child Work</th>
+      <th>Actions</th>
+    </tr>
+    </thead>
+    <tbody>
+    <tr class="hidden">
+      <td>
+        <% curation_concern.ordered_members.to_a.select { |k| k.class.to_s == "FileSet" }.each do |fileset| %>
```
jcoyne: This feels like it should be moved to the model level rather than in a view.

app/views/curation_concerns/base/_form_child_work_relationships.html.erb
```diff
@@ -0,0 +1,65 @@
...
+          <input class="related_works_ids generic_work_filesets_ids" name="generic_work[ordered_member_ids][]" type="hidden" value="<%= fileset.id %>">
+        <% end %>
+      </td>
+      <td></td>
+    </tr>
+    <% curation_concern.ordered_members.to_a.select { |k| k.class.to_s != "FileSet" }.each do |member| %>
```
jcoyne: Move this code out of the view.

app/views/curation_concerns/base/_form_child_work_relationships.html.erb
```diff
@@ -0,0 +1,65 @@
...
+        <tr>
+          <td>
+            <%= link_to "#{member.title.first}", polymorphic_path([main_app, :curation_concerns, curation_concern.model_name.singular], id: member.id) %>
```
jcoyne: 1) Why are you interpolating the string rather than just `member.title.first`
2) Can you just `link_to title, member`?

app/views/curation_concerns/base/_form_child_work_relationships.html.erb
```diff
@@ -0,0 +1,65 @@
...
+            <input class="string multi_value optional form-control related_works_ids generic_work_child_members_ids form-control multi-text-field hidden" value="<%= member.id %>" id="generic_work_child_members_ids" aria-labelledby="generic_work_child_members_ids_label" name="generic_work[ordered_member_ids][]" type="text">
+          </td>
+          <td>
+            <div class="child-actions">
+              <%= link_to "Edit", edit_polymorphic_path([main_app, :curation_concerns, curation_concern.model_name.singular], id: member.id), target: "_blank", class: 'btn btn-default' %>
```
jcoyne: `link_to "Edit", [main_app, :edit, member]`

app/views/curation_concerns/base/_relationships_collection_row.html.erb
```diff
@@ -0,0 +1,16 @@
+<tr>
+<th><%= t(".#{model_name}", type: model_name) %></th>
+<td>
+  <% if collection.blank? %>
+    <p><%= t('.empty', type: model_name) %></p>
+  <% else %>
+    <ul class="tabular">
+      <% collection.each do |col| %>
+        <li class='attribute title'>
+          <%= link_to col.title.first, contextual_path(col, nil) %>
```
jcoyne: Why are we calling contextual_path with a `nil` parent?

app/views/curation_concerns/base/_relationships_member_rows.html.erb
```diff
@@ -0,0 +1,24 @@
...
+  </tr>
+<% else %>
+  <% member_presenters.each_pair do |model_name, members| %>
+    <tr>
+      <th><%= t(".#{model_name.downcase}", type: model_name) %></th>
+      <td>
+        <ul class="tabular">
+          <% members.each do |member| %>
+            <li class='attribute title'>
+              <%= link_to member.title.first, contextual_path(member, nil) %>
```
jcoyne: Why are we calling contextual_path with a `nil` parent?

app/views/curation_concerns/base/_relationships_collection_row.html.erb
revgum: This line might render a child work, parent work, or a Collection.. Collections don't have the parent presenter as far as I know. I've done some debugging and also tried using polymorphic_path([col]) in this case and it doesn't work properly. What am I missing, can you suggest a better way to render this link?

app/assets/javascripts/sufia/relationships/table.es6
```diff
@@ -0,0 +1,206 @@
...
+  clickAddbutton($row) {
+    $row.find(".btn-add-row").click();
+  }
+
+  /**
+   * Handle click events by the "Remove" buttons in the table, and calling the
+   * server to handle the request
+   */
+  bindRemoveButton() {
+    let $this = this;
```
jcoyne: In ES6 there's no need to capture the pointer.  Use the double arrow operator instead. 
```es6
...on('click', (event) => {
```
that will bind `this` to the correct scope.

app/assets/javascripts/sufia/relationships/table.es6
```diff
@@ -0,0 +1,206 @@
...
+  }
+
+  /**
+   * Handle keyup and keypress events at the form level to prevent the ENTER key
+   * from submitting the form. ENTER key within a relationships table should
+   * click the "Add" button instead. ESC key should clear the input and hide the
+   * error message.
+   */
+  bindKeyEvents() {
+    let $this = this;
```
jcoyne: Use the double arrow operator instead.

app/assets/javascripts/sufia/relationships/table.es6
```diff
@@ -0,0 +1,206 @@
...
+        args.on_success($this, args, json);
+      })
+      .fail(function(jqxhr, status, err) {
+        args.on_error($this, args, jqxhr, status, err);
+      });
+  }
+
+  /**
+   * Set a warning message to alert the user on an error
+   * @param {jQuery} $this the RelationshipsTable class instance
```
jcoyne: There is no need to pass this as a parameter, just use `this`.

app/assets/javascripts/sufia/relationships/table.es6
```diff
@@ -0,0 +1,206 @@
...
+    let message = jqxhr.statusText;
+    if(jqxhr.responseJSON){
+      message = jqxhr.responseJSON.description;
+    }
+    $this.setWarningMessage(args.row, message);
+  }
+
+  /**
+   * Remove the row when the API returns this type of success
+   * @param {jQuery} $this the RelationshipsTable class instance
```
jcoyne: Don't pass this as a parameter.

app/assets/javascripts/sufia/relationships/table.es6
```diff
@@ -0,0 +1,206 @@
...
+   */
+  handleRemoveRowSuccess($this, args, json) {
+    args.row.remove();
+  }
+
+  /**
+   * Add a new row to the table, query the server for details about the work to
+   * set the title and link for the new work that was added. Hide the input
+   * field and display the title and edit button
+   * @param {jQuery} $this the RelationshipsTable class instance
```
jcoyne: Don't pass this as a parameter.

app/forms/curation_concerns/forms/work_form.rb
```diff
@@ -0,0 +1,83 @@
+module CurationConcerns
+  module Forms
+    class WorkForm
```
jcoyne: Add class level documentation. What is this class used for?

jcoyne: This already exists in CurationConcerns. Why are you redefining it here?

mjgiarlo: @jcoyne Looks like they added two methods, `#ordered_fileset_members` and `#ordered_work_members`.

jcoyne: Then this should be a new class that extends CurationConcerns::Forms::WorkForm, rather than overwrites it.

app/views/curation_concerns/base/_relationships_collection_row.html.erb
jcoyne: What do you mean by not working properly. What are you seeing? It certainly should work.

app/views/curation_concerns/base/_relationships_collection_row.html.erb
```diff
@@ -0,0 +1,16 @@
+<tr>
+<th><%= t(".#{model_name}", type: model_name) %></th>
+<td>
+  <% if collection.blank? %>
+    <p><%= t('.empty', type: model_name) %></p>
+  <% else %>
+    <ul class="tabular">
+      <% collection.each do |col| %>
```
jcoyne: If this could be something other than a collection, should it have a more descriptive name?

jcoyne: Also, why are we iterating over a collection?  What is the expected type for `collection`?

app/views/records/edit_fields/_in_works_ids.html.erb
```diff
@@ -0,0 +1,59 @@
...
+  input.new-form-control : Input field for client client side validation and additional features
+  .message.has-warning : Used to display UI errors related to input values and server errors
+HTML Properties:
+  table:
+    data-query-url : URL base to append newly added work ID to for querying details (title, href)
+%>
+<% if f.object.class.multiple? key %>
+  <div class="form-group multi_value optional generic_work_in_works_ids managed">
+    <table class="table table-striped related-files relationships-ajax-enabled"
+           data-query-url="<%=polymorphic_path([main_app, :curation_concerns, curation_concern.model_name.singular], id: '$id') %>">
```
jcoyne: What if you have more than one type of model here? Will this still work?

app/views/records/edit_fields/_in_works_ids.html.erb
```diff
@@ -0,0 +1,59 @@
...
+          <th>Parent Work</th>
+          <th>Actions</th>
+        </tr>
+      </thead>
+      <tbody>
+        <% curation_concern.in_works.each do |parent| %>
+          <tr>
+            <td>
+              <%= link_to "#{parent.title.first}", polymorphic_path([main_app, :curation_concerns, curation_concern.model_name.singular], id: parent.id) %>
+              <input class="string multi_value optional form-control related_works_ids form-control multi-text-field hidden" name="generic_work[in_works_ids][]" value="<%= parent.id %>" id="generic_work_in_works_ids" aria-labelledby="generic_work_in_works_ids_label" type="text">
```
jcoyne: Do not hardcode `generic_work`. It may be a work of a different type.

app/views/curation_concerns/base/_form_child_work_relationships.html.erb
```diff
@@ -0,0 +1,65 @@
...
+    <tr>
+      <th>Child Work</th>
+      <th>Actions</th>
+    </tr>
+    </thead>
+    <tbody>
+    <tr class="hidden">
+      <td>
+        <% f.object.ordered_fileset_members.each do |fileset| %>
+          <input class="related_works_ids generic_work_filesets_ids" name="generic_work[ordered_member_ids][]" type="hidden" value="<%= fileset.id %>">
```
jcoyne: Do not hardcode `generic_work` anywhere.

config/locales/sufia.en.yml
```diff
@@ -339,9 +345,20 @@ en:
...
-        empty:       "This %{type} is not currently in any collections."
-        collections: "In collections:"
+       header: Relationships
+       empty:       "This %{type} is not currently in any collections."
+      relationships_collection_row:
+       collection: "In collections:"
+       generic_work: "In works:"
+      relationships_collection_row_empty:
+       collection: "In collections:"
+       generic_work: "In works:"
```
jcoyne: avoid generic_work in the translation file.

config/locales/sufia.en.yml
```diff
@@ -339,9 +345,20 @@ en:
...
+       empty_collection:       "This %{type} is not currently in any collections."
+       empty_generic_work:       "This %{type} is not currently a part of any works."
```
jcoyne: avoid generic_work in the translation file. How about `empty_work`?

config/locales/sufia.en.yml
```diff
@@ -339,9 +345,20 @@ en:
...
+      relationships_member_rows:
+       member: "Has related members:"
+       generic_work: "Has child works:"
```
jcoyne: avoid `generic_work` here.

app/assets/javascripts/sufia/relationships/table.es6
revgum: I need both reference to the RelationshipsTable class instance, and the button which fired the click event. The double arrow would give reference to the button, but I would still need reference to the RelationshipsTable class instance to `.callAjax`, right? 

These references are passed to the error and success callbacks so that they can do things like setting the warning message, etc.

app/views/curation_concerns/base/_relationships_collection_row.html.erb
revgum: collection in this sense isn't necessarily a `Collection`, so the variable naming here probably needs to be changed to maybe `list` and `item` when iterating. The type for a collection, is a grouping of `Collection`, child works, or parent works. The result is in the show page on the relationships table you would see a row with a list of all of the `Collection`, Children, and Parents.

app/views/curation_concerns/base/_relationships_collection_row.html.erb
revgum: When the partial is rendered in the context of a `Collection` it doesn't have a parent presenter like a regular work would. I was expecting `contextual_path(col)` to eventually just call `polymorphic_path([col])` but maybe there needs to be a fix in CC to check `parent_presenter.nil?` to make this all flow nicely without us needing to pass a `nil` in?

https://github.com/projecthydra/curation_concerns/blob/a1fdf2d2deedf089f2ba52dc8a03208f26e0b344/app/services/curation_concerns/contextual_path.rb#L12

app/views/records/edit_fields/_in_works_ids.html.erb
revgum: Would something like `name="<%= curation_concern.model_name.singular %>[in_works_ids][]"` be appropriate?

jcoyne: I think `curation_concern.model_name.param_key` is prefered

app/assets/javascripts/sufia/relationships/table.es6
jcoyne: I guess I prefer where we keep the `this` as is,  use the double arrow function call and address the click target with `$(event.target)`

jcoyne:
One other thing that is missing is a feature test.  The feature is "You can add a work as a child of another work"

app/views/curation_concerns/base/_form_child_work_relationships.html.erb
jcoyne: `f.object.model_name.param_key` instead.

config/locales/sufia.en.yml
```diff
@@ -339,9 +345,20 @@ en:
   curation_concerns:
     base:
       relationships:
-        header: Relationships
-        empty:       "This %{type} is not currently in any collections."
-        collections: "In collections:"
+       header: Relationships
```
jcoyne: indent two spaces please.

awead:
Having looked at this a bit more closely, it's some really nice work. A lot more JS and ajax than we usually use, but that's no reason to hold it up. I think we should get this in the fold and start kicking it around more if we see more needs emerging.
👍

PR #2251, opened 2016-06-21
Use WorkRelation instead of GenericWork

jcoyne:

### DIFF
--- a/lib/sufia.rb
+++ b/lib/sufia.rb
@@ -41,4 +41,8 @@ module Sufia
 
     @config
   end
+
+  def self.primary_work_type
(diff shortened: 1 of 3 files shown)


lib/sufia.rb
```diff
@@ -41,4 +41,8 @@ def self.config(&block)
 
     @config
   end
+
+  def self.primary_work_type
+    CurationConcerns::WorkRelation::DummyModel.primary_concern
```
awead: If there are multiple concerns, this would be the first one listed in the config file?

jcoyne: Yes.

PR #2272, opened 2016-06-22
Use the correct i18n key for sorting featured works, fixes #2270

### LINKED ISSUE #2270: Unable to render homepage as administrator with featured items
### Descriptive summary

As an admin
If featured works exist
I am unable to view the homepage

### Expected behavior

As an admin
If featured works exist
I should be able to view the homepage
And I should be able to re-arrange/remove featured items

### Actual behavior

I18n::MissingTranslationData at /
translation missing: en.drag

1. Build the sample app
1. Edit config/role_map.yml and `admin: - [your email]`
1. Startup app and register your account with `[your email]`
1. Create a work and feature it
1. Visit /

### Related work

Making a previously featured work "Private" does not remove it from the featured works list. #2271

awead:
Fixes #2270 

Replaces `<%= t :drag %>` with something that seems slightly more reasonable, but I'm just making this up.


app/views/sufia/homepage/_sortable_featured.html.erb
```diff
@@ -1,6 +1,6 @@
 <% presenter = f.object.presenter %>
 <li class="dd-item dd3-item featured-item" data-id="<%= presenter.id %>">
-  <div class="dd-handle dd3-handle"><%= t :drag %></div>
+  <div class="dd-handle dd3-handle"><%= t 'sufia.sort_label' %></div>
```
jcoyne: I don't think this is the correct key. This is for the drag and drop handle.

jcoyne: No i18n key was added here: https://github.com/projecthydra/sufia/commit/844e582f66e76ba4832f0314bc964dbff357d544 but as this was copypasta from Spotlight, I think we ought to add the key from spotlight: https://github.com/projectblacklight/spotlight/blob/master/config/locales/spotlight.en.yml#L3

awead:
@jcoyne like this?

jcoyne:
@awead this is better, but I'm not sure we want the spotlight namespace too.

awead:
@jcoyne ok, how about this?

jcoyne:
🎉 This looks good.

PR #2268, opened 2016-06-22
Export GenericFile metdata, versions, and permissions

cam156:
Fixes #2153; fixes #2151; fixes #2154

Exports GenericFile to json format including metadata, version metadata, and permissions.

The intention of this class is to act as a way to transfer data from Sufia 6 so that it can be imported into Sufia 7. See #2136 for user story.

I pulled much of the code from a branch in ScholarSphere written by @hectorcorrea [export_s6](https://github.com/psu-stewardship/scholarsphere/tree/export_s6)

@projecthydra/sufia-code-reviewers

Issue #2136, opened 2016-06-03
Export Metadata from Sufia 6.x to prepare for migration


cam156:
I would like to be able to export metadata my Sufia 6.x repository prior to migration, so that I can move every object to my Sufia PCDM application. Part of user story #2135 

I would like to create a executable I can run that creates a json file for each object that I am migrating from Sufia 6.x to Sufia PCDM.

This will be done when I can run sufia_export.exe in my sufia 6.x application and I produce a directory of json files that contains a json representation of my object (Collections & GenericFiles)  metadata.

My executable should take the following options:
--limit= [# of objects,]                                       # limit the number of items exported (defaults to all)
--id= a list of Id (id1,id2)                                   # only export the given ids
--models= a list of models to be exported        # Models to be exported (defaults to [Collection,GenericFile])

![slide4](https://cloud.githubusercontent.com/assets/1599081/15783313/aab978aa-297b-11e6-905f-53c38bc1184f.png)

PR #2279, opened 2016-06-23
Exporting a Collection's metadata, permissions, and members

### LINKED ISSUE #2155: Migration::Export::CollectionConvertor
I would like to create an class that exports Sufia 6.0 exported Metadata of a Collection in json format. See #2147 for calling Actor.

* to_json - outputs the Collection as JSON including pointers to each member

This is done when I can:

* Migration::Export::CollectionConvertor.new(col) creates an instance of the class for the Collection
* instance.to_json converts the object to json including metadata and member pointers.

An example of Collection export can be found in ScholarSphere [s6_export](https://github.com/psu-stewardship/scholarsphere/blob/export_s6/lib/export/collection_metadata_export.rb)

cam156:
Fixes #2155 

Exports a Collection to json format including metadata, permissions and member ids (the members are expected to be exported separately see #2268).

The intention of this class is to act as a way to transfer data from Sufia 6 so that it can be imported into Sufia 7. See #2136 for user story.

I pulled much of the code from a branch in ScholarSphere written by @hectorcorrea export_s6

### DIFF
--- a/lib/sufia/export/converter.rb
+++ b/lib/sufia/export/converter.rb
@@ -1,0 +1,23 @@
+module Sufia
+  module Export
+    # a base class to convert an ActiveFedora object that contains permissions and allow for pretty json.
+    #
+    class Converter
+      # overrides to_json to optionally allow for a pretty version of the json to be outputted
+      #
+      # @param [Boolean] pretty pass true to output formatted json using pretty_generate
+      def to_json(options = {})
+        pretty = options.delete(:pretty)
+        json = super
+        return json unless pretty
+        JSON.pretty_generate(JSON.parse(json))
+      end
+
+      private
+
+        def permissions(object)
+          object.permissions.map { |p| PermissionConverter.new(p) }
+        end
+    end
+  end
+end
--- a/spec/lib/sufia/export/collection_converter_spec.rb
+++ b/spec/lib/sufia/export/collection_converter_spec.rb
@@ -1,0 +1,32 @@
+require 'spec_helper'
+
+describe Sufia::Export::CollectionConverter do
+  let(:collection) do
+    Collection.create(title: "title1", creator: ["creator1"], description: "description1") do |col|
+      col.apply_depositor_metadata("jilluser")
+    end
+  end
+  let(:permission) { collection.permissions.first }
+  let(:permission2) { collection.permissions.last }
+  let(:json) { "{\"id\":\"#{collection.id}\",\"title\":\"title1\",\"description\":\"description1\",\"creator\":[\"creator1\"],\"members\":[],\"permissions\":[{\"id\":\"#{permission.id}\",\"agent\":\"http://projecthydra.org/ns/auth/person#jilluser\",\"mode\":\"http://www.w3.org/ns/auth/acl#Write\",\"access_to\":\"#{collection.id}\"},{\"id\":\"#{permission2.id}\",\"agent\":\"http://projecthydra.org/ns/auth/group#public\",\"mode\":\"http://www.w3.org/ns/auth/acl#Read\",\"access_to\":\"#{collection.id}\"}]}" }
+  describe "to_json" do
+    subject { described_class.new(collection).to_json }
+    it { is_expected.to eq json }
+
+    context "pretty to_json" do
+      subject { described_class.new(collection).to_json(pretty: true) }
+      let(:json) { "{\n  \"id\": \"#{collection.id}\",\n  \"title\": \"title1\",\n  \"description\": \"description1\",\n  \"creator\": [\n    \"creator1\"\n  ],\n  \"members\": [\n\n  ],\n  \"permissions\": [\n    {\n      \"id\": \"#{permission.id}\",\n      \"agent\": \"http://projecthydra.org/ns/auth/person#jilluser\",\n      \"mode\": \"http://www.w3.org/ns/auth/acl#Write\",\n      \"access_to\": \"#{collection.id}\"\n    },\n    {\n      \"id\": \"#{permission2.id}\",\n      \"agent\": \"http://projecthydra.org/ns/auth/group#public\",\n      \"mode\": \"http://www.w3.org/ns/auth/acl#Read\",\n      \"access_to\": \"#{collection.id}\"\n    }\n  ]\n}" }
+      it { is_expected.to eq json }
+    end
+
+    context "with members" do
+      let(:file1) { create(:generic_file) }
+      let(:file2) { create(:generic_file) }
+      before do
+        collection.members = [file1, file2]
+      end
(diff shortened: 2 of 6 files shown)


lib/sufia/export/converter.rb
```diff
@@ -0,0 +1,21 @@
...
+  module Export
+    # a base class to convert an ActiveFedora object that contains permissions and allow for pretty json.
+    #
+    class Converter
+      # overrides to_json to optionally allow for a pretty version of the json to be outputted
+      #
+      # @param [Boolean] pretty pass true to output formatted json using pretty_generate
+      def to_json(pretty = false)
+        return super unless pretty
+        JSON.pretty_generate(JSON.parse(to_json))
```
mjgiarlo: Is this method recursive? Or should this `to_json` be `super`?

spec/lib/sufia/export/collection_converter_spec.rb
```diff
@@ -0,0 +1,32 @@
+require 'spec_helper'
+
+describe Sufia::Export::CollectionConverter do
+  let(:collection) do
+    Collection.create(title: "title1", creator: ["creator1"], description: "description1") do |col|
```
mjgiarlo: Should `title` be an array here?

lib/sufia/export/converter.rb
hectorcorrea: Good catch. I guess it never got into an infinite loop because the guard class the the top but I think it should be super.

spec/lib/sufia/export/collection_converter_spec.rb
mjgiarlo: Yes, I do believe they should, according to https://github.com/projecthydra/sufia/blob/master/spec/factories/collections_factory.rb#L23-L26

Would it be possible to use a factory here instead of creating a repository object? That'd help with suite speed.

lib/sufia/export/converter.rb
cam156: @mjgiarlo & @hectorcorrea I tried super and that was no good (which is strange) calling to_json without the true then calls super.  Do not ask me why that is different from calling to_json, but the tests fail when I tried the change with the last PR.

spec/lib/sufia/export/collection_converter_spec.rb
cam156: @mjgiarlo This is sufia 6, so there are no factories in existence. Also for Sufia 6 title and description are singular.

mjgiarlo: Ah, right! This context-switching is twisting my wig. Thanks, @cam156.

lib/sufia/export/converter.rb
mjgiarlo: :eyes:  Huh.

cam156: @mjgiarlo I fooled around with some different ways to call super and `super()` works `super` does not it gives a 
```
     TypeError:
       can't dup TrueClass
```
The result is the same either way.  Which is easier to understand? recalling to_json with no parameters or calling super()?

lib/sufia/export/converter.rb
```diff
@@ -0,0 +1,21 @@
...
+  module Export
+    # a base class to convert an ActiveFedora object that contains permissions and allow for pretty json.
+    #
+    class Converter
+      # overrides to_json to optionally allow for a pretty version of the json to be outputted
+      #
+      # @param [Boolean] pretty pass true to output formatted json using pretty_generate
+      def to_json(pretty = false)
+        return super unless pretty
+        JSON.pretty_generate(JSON.parse(super()))
```
mjgiarlo: Thoughts on the `super()` here, @projecthydra/sufia-code-reviewers @jcoyne @jeremyf etc.

lib/sufia/export/converter.rb
hectorcorrea: I would go for `super()` to remove the recursive call.

I think `super` is a keyword whereas `super()` is a function call. Ruby might not like the keyword inside another function call.

lib/sufia/export/converter.rb
cam156: The other choice is to call to_json again with no parameter or false...
```
JSON.pretty_generate(JSON.parse(to_json))
```
or
```
JSON.pretty_generate(JSON.parse(to_json(false)))
```

cam156: or
```
json = super()
return json unless pretty
JSON.pretty_generate(JSON.parse(json))
```

lib/sufia/export/converter.rb
```diff
@@ -0,0 +1,22 @@
...
+    # a base class to convert an ActiveFedora object that contains permissions and allow for pretty json.
+    #
+    class Converter
+      # overrides to_json to optionally allow for a pretty version of the json to be outputted
+      #
+      # @param [Boolean] pretty pass true to output formatted json using pretty_generate
+      def to_json(pretty = false)
+        json = super()
+        return json unless pretty
+        JSON.pretty_generate(JSON.parse(json))
```
cam156: @mjgiarlo Went with a third option suggested by @awead and approved as readable by @mtribone. 👌 ?

cam156:
@mjgiarlo, @awead & @hectorcorrea  I changed the method to match the calling parameters of the original to_json which allows for an options hash.  I moved pretty so that it would be an option.  This allows us to call super without the parens.  Hopefully the new solution ticks all the requests...

mjgiarlo:
Taking a look, @cam156!

lib/sufia/export/converter.rb
```diff
@@ -0,0 +1,23 @@
+module Sufia
+  module Export
+    # a base class to convert an ActiveFedora object that contains permissions and allow for pretty json.
+    #
+    class Converter
+      # overrides to_json to optionally allow for a pretty version of the json to be outputted
+      #
+      # @param [Boolean] pretty pass true to output formatted json using pretty_generate
+      def to_json(options = nil)
```
mjgiarlo: Why set `options` to `nil` here? Then you have to check if it's `#blank?` on the next line. You could make this less ask-y and more tell-y by doing this:

```ruby
def to_json(options = {})
  json = super
  return json unless options[:pretty]
  JSON.pretty_generate(JSON.parse(json))
end
```

I don't care strongly about this, though. Are you inclined to care about that, or :shipit:?

mjgiarlo:
@cam156 I have just one comment, but like I say, I don't care very much about it. Should I :shipit: or do you want to make that one change?

lib/sufia/export/converter.rb
cam156: @mjgiarlo I can make the change.

mjgiarlo:
TRAVIIIIIIIIIIIIIIIIIIS

PR #2281, opened 2016-06-23
TrophyPresenter should return a title, fixes #2280

### LINKED ISSUE #2280: Work titles are not appearing correctly in highlighted works tab
### Descriptive summary

When I highlight a file on my profile
Then I do not see its title on the highlighted works tab

highlight

### Expected behavior

I should see the title.

### Actual behavior

I do not.

1. Create a work
1. Highlight it using the dropdown menu from the work's dashboard entry
1. Visit your profile

awead:
Fixes #2280 

### DIFF
--- a/app/presenters/sufia/trophy_presenter.rb
+++ b/app/presenters/sufia/trophy_presenter.rb
@@ -7,6 +7,8 @@ module Sufia
 
     attr_reader :solr_document
 
+    delegate :to_s, to: :solr_document
+
     def self.find_by_user(user)
       work_ids = user.trophies.pluck(:work_id)
       query = ActiveFedora::SolrQueryBuilder.construct_query_for_ids(work_ids)
--- a/spec/presenters/sufia/trophy_presenter_spec.rb
+++ b/spec/presenters/sufia/trophy_presenter_spec.rb
@@ -16,7 +16,7 @@ RSpec.describe Sufia::TrophyPresenter do
   end
 
   let(:presenter) { described_class.new(solr_document) }
-  let(:solr_document) { SolrDocument.new(id: '123456', has_model_ssim: 'GenericWork') }
+  let(:solr_document) { SolrDocument.new(id: '123456', has_model_ssim: 'GenericWork', title_tesim: ['A Title']) }
 
   describe "id" do
     subject { presenter.id }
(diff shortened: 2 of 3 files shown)


app/presenters/sufia/trophy_presenter.rb
```diff
@@ -18,6 +18,10 @@ def thumbnail_path
       solr_document[CatalogController.blacklight_config.index.thumbnail_field]
     end
 
+    def to_s
+      solr_document.title.first
```
cam156: @awead I wonder if it would be better to just delegate to the solr_document. Curation concerns is already providing a to_s: https://github.com/projecthydra/curation_concerns/blob/master/app/models/concerns/curation_concerns/solr_document_behavior.rb

mjgiarlo: :+1: to what @cam156 said.

spec/presenters/sufia/trophy_presenter_spec.rb
```diff
@@ -16,7 +16,7 @@
   end
 
   let(:presenter) { described_class.new(solr_document) }
-  let(:solr_document) { SolrDocument.new(id: '123456', has_model_ssim: 'GenericWork') }
+  let(:solr_document) { SolrDocument.new(id: '123456', has_model_ssim: 'GenericWork', title_tesim: 'A Title') }
```
mjgiarlo: Aren't `has_model_ssim` and `title_tesim` multi-valued fields? Should they return arrays? I'm wondering whether this expectation sets up a mismatch between what we're testing and how the app behaves.

awead: @mjgiarlo yes, but in the end it doesn't matter since it's cast to an array anyway.

mjgiarlo: @awead In the `SolrDocumentBehavior`, you mean, or elsewhere?

awead:
done.

spec/presenters/sufia/trophy_presenter_spec.rb
awead: In Behavior. It does `Array.wrap` with title, so the expectation seems to be either string or array.

cam156:
👍 waiting on travis

PR #2314, opened 2016-06-30
Add google_analytics_id to Sufia configuration

### LINKED ISSUE #2295: Where do we put google_analytics_id?
It looks like this was removed from Sufia.config, but there are still multiple references:
https://github.com/projecthydra/sufia/blob/43a3271050de6db0bec82847882bf9d88640e3ad/app/views/_ga.html.erb#L3
https://github.com/projecthydra/sufia/blob/c3e0ffda0abe9b1bd8fadc8d8d4323eb3201c3af/app/services/sufia/analytics.rb#L62

Do we follow CurationConerns?

https://github.com/projecthydra/curation_concerns/blob/a1fdf2d2deedf089f2ba52dc8a03208f26e0b344/config/locales/curation_concerns.en.yml#L13

awead:
Fixes #2295 

This was removed in CurationConcerns, https://github.com/projecthydra/curation_concerns/pull/867

### DIFF
--- a/lib/sufia/configuration.rb
+++ b/lib/sufia/configuration.rb
@@ -86,6 +86,11 @@ module Sufia
       @always_display_share_button
     end
 
+    attr_writer :google_analytics_id
+    def google_analytics_id
+      @google_analytics_id ||= nil
+    end
+
     # Defaulting analytic start date to whenever the file was uploaded by leaving it blank
     attr_writer :analytic_start_date
     attr_reader :analytic_start_date
--- a/spec/helpers/blacklight_helper_spec.rb
+++ b/spec/helpers/blacklight_helper_spec.rb
@@ -22,6 +22,7 @@ RSpec.describe BlacklightHelper, type: :helper do
   let(:document) { SolrDocument.new(attributes) }
   before do
     allow(helper).to receive(:blacklight_config).and_return(blacklight_config)
+    allow(controller).to receive(:action_name).and_return('index')
   end
 
   describe "render_index_field_value" do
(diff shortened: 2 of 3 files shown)


mjgiarlo:
The fact that our test suite passes without this config change makes me wonder if this PR might be an opportunity to add a new test that would fail without the new `attr_writer`. Otherwise 👍

awead:
@mjgiarlo I've added tests for all config options.

spec/test_app_templates/lib/generators/test_app_generator.rb
```diff
@@ -57,4 +57,10 @@ def enable_arkivo_api
   def relax_routing_constraint
     gsub_file 'config/initializers/arkivo_constraint.rb', 'false', 'true'
   end
+
+  # Turbolinks 5.0 is causing tests to fail with PhantomJS 2.1.1
+  def restrict_turbolinks_version
```
mjgiarlo: You can add this to `Gemfile.extra`

mjgiarlo:
:red_circle:

spec/test_app_templates/lib/generators/test_app_generator.rb
awead: No, because it gets added automatically when Rails is generated.

awead:
@mjgiarlo yeah, I don't know how to make this work, other than removing PhantomJS from Travis, but that's not a viable solution.

spec/test_app_templates/lib/generators/test_app_generator.rb
mjgiarlo: :ghost: Dang.

mjgiarlo:
The error I'm seeing currently is:

```
You have requested:
  coffee-rails ~> 4.1.0
```

Is that related to the phantom thing?

PR #2316, opened 2016-07-01
Export Actor which will export ActiveFedora classes to json files

### LINKED ISSUE #2147: Migration::Export::Actor
I would like to create an class that exports Sufia 6.0 exported Metadata in json format into a file. See #2136 for executable.

* call - loops through all the instances of a model and exports the metadata
  * parameters:
     * model_types # a list of models to export
     * opts  # either a limit number or a list of ids to export
* register_convertor
  * parameters
    * model_class  # the class of the model to register
    * convertor       # the class that will convert the object's metadata to json

The class instance will allow for additional convertors to be registered so that default conversion can be overridden.  By default there will be  GenericFile & Collection convertor registered.

This is done when I can:

* Migration::Export::Actor.new creates an instance of the class
* instance.register_convertor registers a new conversion for the actor
* instance.call converts all objects to json files including metadata and permissions, along with pointers to content and versions.  There will be one file per object with the model type and ID as the file name.

An example of an Export Service can be found in ScholarSphere [s6_export](https://github.com/psu-stewardship/scholarsphere/blob/export_s6/lib/export/service.rb)

cam156:
Fixes #2147 for Sufia 6.x

Creates the Actor to export all the ActiveFedora Models using the Converters to json files.

Uses the latest Active Fedora 6.7.x branch to get the correct operation for the where(id: [id1,id2]), which would create a solr query like `q: (id_si: id1 OR id_si: id2)` instead of the released version which creates a solr query like  `q:  (id_si: id1 AND id_si: id2)` 

### DIFF
--- a/lib/sufia/export/actor.rb
+++ b/lib/sufia/export/actor.rb
@@ -1,0 +1,98 @@
+module Sufia
+  module Export
+    # Convert a GenericFile including metadata, permissions and version metadata into a PORO
+    # so that the metadata can be exported in json format using to_json
+    #
+    class Actor
+      attr_reader :converter_registry, :limit, :ids
+      attr_accessor :conversion_count
+
+      # initialize the class with the default registry
+      def initialize
+        @converter_registry = default_registry
+      end
+
+      # register a converter for a new class or overwrite the converter for a default class
+      #
+      # @param [Class] model_class     The ActiveFedora model class to be converter to json
+      # @param [Class] converter_class The class that will convert from ActiveFedora to json
+      def register_converter(model_class, converter_class)
+        raise(RegistryError, "Model (#{model_class.name}) for conversion must be an ActiveFedora::Base") unless model_class.ancestors.include?(ActiveFedora::Base)
+        raise(RegistryError, "Converter (#{converter_class.name}) for conversion must be an Sufia::Export::Converter") unless converter_class.ancestors.include?(Sufia::Export::Converter)
+        converter_registry[model_class] = converter_class
+      end
+
+      # Convert the classes using the registered converters from ActiveFedora to json files
+      #
+      # @param [Array] model_class_list list of classes to be converter
+      # @param [Hash]  opts
+      # @option opts [Number] :limit Limits the number of conversion done (defaults to -1 or all)
+      # @option opts [Array]  :ids   List of ids to be converted.  Can be from any model
+      def call(model_class_list = converter_registry.keys, opts = {})
+        @conversion_count = 0
+        validate_class_list(model_class_list)
+        parse_options(opts)
+        export_models(model_class_list)
+      end
+
+      private
+
+        def parse_options(opts)
+          @limit = opts[:limit] || -1
+          @ids = opts[:ids]
+        end
+
+        def model_scope(model_class)
+          scope = model_class.all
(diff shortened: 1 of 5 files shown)


Gemfile
```diff
@@ -7,6 +7,7 @@ gemspec
 gem 'kaminari', github: 'jcoyne/kaminari', branch: 'sufia'
 gem 'sufia-models', path: './sufia-models'
 gem 'slop', '~> 4.2' # This just helps us generate a valid Gemfile.lock when Rails 4.2 is installed (which requires byebug which has a dependency on slop)
+gem 'active-fedora', github: 'projecthydra/active_fedora', branch: '9.7.x-stable'
```
awead: Shall we do a 9.7.x+1 release so don't have to pin?

lib/sufia/export/actor.rb
```diff
@@ -0,0 +1,99 @@
...
+
+        def file_name(model)
+          File.join(file_path, "#{model.class.name.underscore}_#{model.id}.json")
+        end
+
+        def default_registry
+          { ::GenericFile => Sufia::Export::GenericFileConverter, ::Collection => Sufia::Export::CollectionConverter }
+        end
+
+        def file_path
```
awead: guard clause here: `return @file_path unless @file_path.blank?`

lib/sufia/export/actor.rb
```diff
@@ -0,0 +1,99 @@
...
+          if @file_path.blank?
+            @file_path = "tmp/export"
+            FileUtils.mkdir @file_path unless File.exist?(@file_path)
```
awead: mkdir_p won't raise an error if the directory exists, so you could eliminate the unless.

awead:
Looks great! Just a few tidbits. Also, can you correct the typo in the commit message?

Gemfile
cam156: @awead sure!  That would be 9.7.2 since 9.7.1 was released in January. Do you have the desire?

lib/sufia/export/actor.rb
```diff
@@ -0,0 +1,98 @@
...
+      private
+
+        def parse_options(opts)
+          @limit = opts[:limit] || -1
+          @ids = opts[:ids]
+        end
+
+        def model_scope(model_class)
+          scope = model_class.all
+          scope  = scope.where(id: ids) unless ids.blank?
```
jcoyne: Why not scope.find(ids) ?

cam156: @jcoyne becuase that would error if the ID is not from that particular class. We could change the ids to take a hash with like ids: {GenericFile: ['abc123', '111222'], Collection: ['ccc111']} and know which class to expect the ids from, but the weher seemed simpler.

jcoyne: Okay, that also sounds like an AF bug, but carry on.

cam156: @jcoyne: It is the same behavior as ActiveRecord. If you find with an id that does not exist it throws an exception. If you use where there is no exception if nothing is found...

jcoyne: Yep, you are right:
```
User.find([1, 2])
ActiveRecord::RecordNotFound: Couldn't find all Users with 'id': (1, 2) (found 1 results, but was looking for 2)
```

Issue #2317, opened 2016-07-01
Blacklight 6.3


awead:
### Descriptive summary

Blacklight 6.3 fixes problems with turbolinks 5.0, but there are some breakages. Some of the expectations with `itemprop` results have changed. Particularly, the contributor term isn't rendering properly.

atz:
Pls add milestone.  Otherwise it is unclear what the blocker is blocking.

atz:
Note: contributor rendering may have just been our own error in a helper_method.

PR #2318, opened 2016-07-01
Add Google analytics config and update to Blacklight 6.3

### LINKED ISSUE #2295: Where do we put google_analytics_id?

awead:
Fixes a number of problems: #2295, #2317, #2323, and #2322

1. Adds config value for GA back into Sufia config
2. Gets Sufia working with Blacklight 6.3.
3. Adds `contributor` facet for consistency with other searched fields and to display microdata properly.

Note the changes in the helper expectations. These seem to be more correct than the previous ones.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -10,10 +10,10 @@ group :development, :test do
 end
 
 # BEGIN ENGINE_CART BLOCK
-# engine_cart: 0.8.0
-# engine_cart stanza: 0.8.0
+# engine_cart: 0.10.0
+# engine_cart stanza: 0.10.0
 # the below comes from engine_cart, a gem used to test this Rails engine gem in the context of a Rails app.
-file = File.expand_path("Gemfile", ENV['ENGINE_CART_DESTINATION'] || ENV['RAILS_ROOT'] || File.expand_path(".internal_test_app", File.dirname(__FILE__)))
+file = File.expand_path('Gemfile', ENV['ENGINE_CART_DESTINATION'] || ENV['RAILS_ROOT'] || File.expand_path('.internal_test_app', File.dirname(__FILE__)))
 if File.exist?(file)
   begin
     eval_gemfile file
@@ -24,13 +24,22 @@ if File.exist?(file)
 else
   Bundler.ui.warn "[EngineCart] Unable to find test application dependencies in #{file}, using placeholder dependencies"
 
-  gem 'rails', ENV['RAILS_VERSION'] if ENV['RAILS_VERSION']
+  if ENV['RAILS_VERSION']
+    if ENV['RAILS_VERSION'] == 'edge'
+      gem 'rails', github: 'rails/rails'
+      ENV['ENGINE_CART_RAILS_OPTIONS'] = '--edge --skip-turbolinks'
+    else
+      gem 'rails', ENV['RAILS_VERSION']
+    end
+  end
 
-  if ENV['RAILS_VERSION'].nil? || ENV['RAILS_VERSION'] =~ /^4.2/
-    gem 'responders', "~> 2.0"
-    gem 'sass-rails', ">= 5.0"
-  else
-    gem 'sass-rails', "< 5.0"
+  case ENV['RAILS_VERSION']
+  when /^4.2/
+    gem 'responders', '~> 2.0'
+    gem 'sass-rails', '>= 5.0'
+    gem 'coffee-rails', '~> 4.1.0'
+  when /^4.[01]/
+    gem 'sass-rails', '< 5.0'
   end
 end
 # END ENGINE_CART BLOCK
--- a/app/helpers/sufia/sufia_helper_behavior.rb
+++ b/app/helpers/sufia/sufia_helper_behavior.rb
@@ -107,7 +107,7 @@ module Sufia
     # @param [String,Hash] text either the string to escape or a hash containing the
     #                           string to escape under the :value key.
     def iconify_auto_link(text, showLink = true)
-      text = presenter(text[:document]).render_values(text[:value], text[:config]) if text.is_a? Hash
+      text = presenter(text[:document]).field_value(Array.wrap(text[:value]).first, text[:config]) if text.is_a? Hash
       # this block is only executed when a link is inserted;
       # if we pass text containing no links, it just returns text.
       auto_link(html_escape(text)) do |value|
--- a/spec/helpers/blacklight_helper_spec.rb
+++ b/spec/helpers/blacklight_helper_spec.rb
@@ -22,6 +22,7 @@ RSpec.describe BlacklightHelper, type: :helper do
   let(:document) { SolrDocument.new(attributes) }
   before do
     allow(helper).to receive(:blacklight_config).and_return(blacklight_config)
+    allow(controller).to receive(:action_name).and_return('index')
   end
 
   describe "render_index_field_value" do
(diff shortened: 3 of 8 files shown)


awead:
There are still two failures here which are eluding me at present.

spec/helpers/blacklight_helper_spec.rb
```diff
@@ -44,17 +45,17 @@ def search_action_path(stuff)
...
 
       context "subject_tesim" do
         let(:field_name) { 'subject_tesim' }
-        it { is_expected.to eq "<span itemprop=\"about\"><a href=\"/catalog?f%5Bsubject_sim%5D%5B%5D=Awesome\"><span itemprop=\"about\">Awesome</span></a></span>" }
+        it { is_expected.to eq "<span itemprop=\"about\"><a href=\"/catalog?f%5Bsubject_sim%5D%5B%5D=Awesome\">Awesome</a></span>" }
       end
 
       context "creator_tesim" do
         let(:field_name) { 'creator_tesim' }
-        it { is_expected.to eq "<span itemprop=\"creator\"><a href=\"/catalog?f%5Bcreator_sim%5D%5B%5D=Justin\"><span itemprop=\"creator\">Justin</span></a></span> and <span itemprop=\"creator\"><a href=\"/catalog?f%5Bcreator_sim%5D%5B%5D=Joe\"><span itemprop=\"creator\">Joe</span></a></span>" }
```
cbeer: This seems like a regression to me; @jcoyne, think this is related to the presenter work you did?

awead: It looks "better" without the nested spans, only using the itemprop attribute once. Although, this may not be correct.

jcoyne: Yeah, the nested said and was a bug I fixed

spec/helpers/blacklight_helper_spec.rb
```diff
@@ -44,17 +45,17 @@ def search_action_path(stuff)
...
+        it { is_expected.to eq "<span itemprop=\"creator\"><a href=\"/catalog?f%5Bcreator_sim%5D%5B%5D=Justin\">Justin</a></span> and <span itemprop=\"creator\"><a href=\"/catalog?f%5Bcreator_sim%5D%5B%5D=Joe\">Joe</a></span>" }
       end
```
jcoyne: Spans

spec/helpers/blacklight_helper_spec.rb
cbeer: It seems like putting all the values into a single span changes the meaning of the schema.org markup, doesn't it? It turns a multivalued property into a delimited single-valued one.

jcoyne: It's a single wrapper per item now. Previously it was double wrapping each item

jcoyne: It's a single wrapper per item now. Previously it was double wrapping each item

cbeer: 👍 sorry, misread the diff.

atz:
Note: I'm fixing some misaligned helper expectations in my `help_the_helper` branch.

spec/helpers/blacklight_helper_spec.rb
mjgiarlo: I'm finding the documentation of schema.org to be... lacking. But StackOverflow seems to suggest that the preferred markup format for multiple keywords might now be:

```html
  <span itemprop="keywords">
    <a href="/catalog?f%5Bkeyword_sim%5D%5B%5D=taco">taco</a>, 
    <a href="/catalog?f%5Bkeyword_sim%5D%5B%5D=mustache">mustache</a>
  </span>
```

IOW, one use of `itemprop="keywords"` and a comma-separated list of keywords.

awead:
@mjgiarlo this is ready for a re-review

mjgiarlo:
Should we pin AF::Noid to < 2?

Gemfile
```diff
@@ -24,13 +24,22 @@ if File.exist?(file)
 else
   Bundler.ui.warn "[EngineCart] Unable to find test application dependencies in #{file}, using placeholder dependencies"
 
-  gem 'rails', ENV['RAILS_VERSION'] if ENV['RAILS_VERSION']
+  if ENV['RAILS_VERSION']
```
mjgiarlo: I don't think you need the outer `if` here. If `ENV` doesn't contain the key `RAILS_VERSION`, the `if` on the following line won't blow up; it'll just return `false`.

mjgiarlo:
I have one minor comment about the Gemfile but I'd be happy to merge this as is. Anything else to say about this, @atz, @jcoyne, @cbeer, or :shipit:?

Gemfile
cbeer: This is from upstream in `engine_cart`. I think the goal is to /not/ generate any of these dependencies unless `ENV['RAILS_VERSION']` is explicitly specified.

mjgiarlo: :ok_hand:

app/helpers/sufia/sufia_helper_behavior.rb
```diff
@@ -107,7 +107,7 @@ def link_to_index_field(name, value)
     # @param [String,Hash] text either the string to escape or a hash containing the
     #                           string to escape under the :value key.
     def iconify_auto_link(text, showLink = true)
-      text = presenter(text[:document]).render_values(text[:value], text[:config]) if text.is_a? Hash
+      text = presenter(text[:document]).field_value(Array.wrap(text[:value]).first, text[:config]) if text.is_a? Hash
```
cbeer: Where are we pinning to blacklight ~> 6.3?

cbeer:
Left a comment we should address before merging https://github.com/projecthydra/sufia/pull/2318/files#r69831905.. maybe it's happening, I just don't see it explicit.

sufia.gemspec
```diff
@@ -18,7 +18,7 @@ Gem::Specification.new do |spec|
   spec.version       = Sufia::VERSION
   spec.license       = 'Apache2'
 
-  spec.add_dependency 'curation_concerns', '>= 1.0.0', '< 2'
+  spec.add_dependency 'curation_concerns', '>= 1.1.0', '< 2'
```
cbeer: Why not `~> 1.1`?

app/helpers/sufia/sufia_helper_behavior.rb
cbeer: Looks like curation_concerns is pinning it for us.

atz:
The middle commit roughly matches changes I made in the `help_the_helper` branch, with at least one improvement here in declaring the contributor facet.  I have a lot more documentation, testing and refactoring of the helper methods, but should be able to rebase OK.  👍

awead:
@atz maybe you could rebase your PR off of this? I'd like to see yours get merged as it has improvements as well.

sufia.gemspec
awead: Can do.

awead:
@cbeer @mjgiarlo all fixed up.

atz:
@awead basically no different than rebasing off master after this is merged, but yes.

PR #2319, opened 2016-07-02
Help the helper

### LINKED ISSUE #2309: Link to contributor on index result produces an broken search
When I click on this link in the search result:

screen shot 2016-06-28 at 8 49 26 pm

It takes me to:
http://localhost:3000/catalog?frank="contributor"&search_field=advanced

which isn't filtering in any way. It's returning all results.

@mtribone any ideas?

### LINKED ISSUE #2313: sufia_helper has issues
`master` branch.
Related to: #2309, #2312 

- Documentation coverage lacking: contains 5 of the top 6 "Please take a look" targets from [inch](https://inch-ci.org/github/projecthydra/sufia)  
- Test coverage lacking.  
-  `link_to_facet_list`: the `html_safe`-ness of the returned string varies because the default empty message is not marked html_safe.  
- `search_state_with_facets`: accepts a hash of facets, but only uses the first, silently ignores the rest.
- `link_to_index_field(name, value)` is basically `link_to_field(name, value, name)`, but with more copy/paste.

atz:
Fixes #2309 
Fixes #2313

Begins to address underlying breakage from upstream Blacklight changes.  Removes double-wrapped spans in field rendering.  Corrects expectations that were *never* valid.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -90,4 +90,3 @@ RSpec/DescribeClass:
     - 'spec/features/**/*'
     - 'spec/config/sufia_events_spec.rb'
 
-
--- a/app/helpers/sufia/sufia_helper_behavior.rb
+++ b/app/helpers/sufia/sufia_helper_behavior.rb
@@ -28,28 +28,28 @@ module Sufia
     end
 
     # Only Chrome supports this
+    # @return [Boolean]
+    # @todo Replace uses with more granular client-side detection (as jQuery-fileuploader already does)
     def browser_supports_directory_upload?
       user_agent.include? 'Chrome'
     end
 
-    # @param [Hash] options a list of options that blacklight passes from helper_method
-    #                       invocation.
+    # A Blacklight helper_method
+    # @param [Hash] options from blacklight invocation of helper_method
+    # @return [Date]
     def human_readable_date(options)
       Date.parse(options[:value]).to_formatted_s(:standard)
     end
 
     def error_messages_for(object)
-      if object.try(:errors) && object.errors.full_messages.any?
-        content_tag(:div, class: 'alert alert-block alert-error validation-errors') do
-          content_tag(:h4, I18n.t('sufia.errors.header', model: object.class.model_name.human.downcase), class: 'alert-heading') +
-            content_tag(:ul) do
-              object.errors.full_messages.map do |message|
-                content_tag(:li, message)
-              end.join('').html_safe
-            end
-        end
-      else
-        '' # return empty string
+      return '' unless object.try(:errors) && object.errors.full_messages.any?
+      content_tag(:div, class: 'alert alert-block alert-error validation-errors') do
+        content_tag(:h4, I18n.t('sufia.errors.header', model: object.class.model_name.human.downcase), class: 'alert-heading') +
+          content_tag(:ul) do
+            object.errors.full_messages.map do |message|
+              content_tag(:li, message)
+            end.join('').html_safe
+          end
       end
     end
 
@@ -62,6 +62,7 @@ module Sufia
       end
     end
 
+    # @return [Boolean]
     def has_collection_search_parameters?
       !params[:cq].blank?
     end
@@ -70,21 +71,31 @@ module Sufia
       ActiveFedora::Base.where(DepositSearchBuilder.depositor_field => user.user_key).count
     end
 
-    def link_to_facet(field, field_string)
-      path = search_action_path(search_state.add_facet_params_and_redirect(field_string, field))
-      link_to(field, path)
+    # @param item [Object]
+    # @param field [String]
+    # @return [ActiveSupport::SafeBuffer] the html_safe link
+    def link_to_facet(item, field)
+      path = search_action_path(search_state.add_facet_params_and_redirect(field, item))
+      link_to(item, path)
     end
 
-    # @param values [Array] The values to display
-    # @param solr_field [String] The name of the solr field to link to without its suffix (:facetable)
-    # @param empty_message [String] ('No value entered') The message to display if no values are passed in.
-    # @param separator [String] (', ') The value to join with.
-    def link_to_facet_list(values, solr_field, empty_message = "No value entered", separator = ", ")
+    # @param values [Array{String}] strings to display
+    # @param solr_field [String] name of the solr field to link to, without its suffix (:facetable)
+    # @param empty_message [String] message to display if no values are passed in
+    # @param separator [String] value to join with
+    # @return [ActiveSupport::SafeBuffer] the html_safe link
+    def link_to_facet_list(values, solr_field, empty_message = "No value entered".html_safe, separator = ", ")
       return empty_message if values.blank?
       facet_field = Solrizer.solr_name(solr_field, :facetable)
       safe_join(values.map { |item| link_to_facet(item, facet_field) }, separator)
     end
 
+    # @param name [String] field name
+    # @param value [String] field value
+    # @param label [String] defaults to value
+    # @param facet_hash [Hash] first argument to Blacklight::SearchState.new
+    # @return [ActiveSupport::SafeBuffer] the html_safe link
+    # @see Blacklight::SearchState#initialize
     def link_to_field(name, value, label = nil, facet_hash = {})
       label ||= value
       params = { search_field: 'advanced', name => "\"#{value}\"" }
@@ -92,20 +103,19 @@ module Sufia
       link_to(label, main_app.search_catalog_path(state))
     end
 
-    def index_field_link(options = {})
-      field_name = options[:config][:field_name]
+    # A Blacklight helper_method
+    # @param options [Hash] Blacklight sends :document, :field, :config, :value and whatever else was in options
+    # @option options [Hash] :config including {:field_name => "my_name"}
+    # @option options [Array{String}] :value
+    # @return [ActiveSupport::SafeBuffer] the html_safe link
+    def index_field_link(options)
+      raise ArgumentError unless options[:config] && options[:config][:field_name]
+      name = options[:config][:field_name]
       values = options[:value]
-      safe_join(values.map { |item| link_to_index_field(item, field_name) }, ", ")
+      safe_join(values.map { |item| link_to_field(name, item, item) }, ", ".html_safe)
     end
 
-    def link_to_index_field(name, value)
-      params = { search_field: 'advanced', name => "\"#{value}\"" }
-      state = search_state.params_for_search(params)
-      link_to(name, main_app.search_catalog_path(state))
-    end
-
-    # @param [String,Hash] text either the string to escape or a hash containing the
-    #                           string to escape under the :value key.
+    # @param text [String,Hash] string to escape or a hash containing that string under the :value key.
     def iconify_auto_link(text, showLink = true)
       text = index_presenter(text[:document]).field_value(Array.wrap(text[:value]).first, text[:config]) if text.is_a? Hash
       # this block is only executed when a link is inserted;
@@ -116,6 +126,7 @@ module Sufia
     end
 
     # @param [String,User,Hash] args if a hash, the user_key must be under :value
+    # @return [ActiveSupport::SafeBuffer] the html_safe link
     def link_to_profile(args)
       user_or_key = args.is_a?(Hash) ? args[:value].first : args
       user = case user_or_key
@@ -125,18 +136,13 @@ module Sufia
                ::User.find_by_user_key(user_or_key)
              end
       return user_or_key if user.nil?
-
-      text = if user.respond_to? :name
-               user.name
-             else
-               user_or_key
-             end
-
+      text = user.respond_to?(:name) ? user.name : user_or_key
       link_to text, Sufia::Engine.routes.url_helpers.profile_path(user)
     end
 
-    # @param [Hash] options hash from blacklight passes from helper_method
-    #                       invocation. Maps rights URIs to links with labels.
+    # A Blacklight helper_method
+    # @param [Hash] options from blacklight helper_method invocation. Maps rights URIs to links with labels.
+    # @return [ActiveSupport::SafeBuffer] rights statement links, html_safe
     def rights_statement_links(options = {})
       options[:value].map { |right| link_to RightsService.label(right), right }.to_sentence.html_safe
     end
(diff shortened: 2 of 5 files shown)


.rubocop.yml
```diff
@@ -61,6 +61,11 @@ Style/CollectionMethods:
 Style/WordArray:
   Enabled: false
 
+# in rspec, expect{...} is conventional
```
mjgiarlo: That doesn't match my expectation (so to speak), @atz. I would expect to see `expect {...}` vs. `expect{...}`.

app/helpers/sufia/sufia_helper_behavior.rb
```diff
@@ -70,42 +72,52 @@ def number_of_deposits(user)
...
-    # @param empty_message [String] ('No value entered') The message to display if no values are passed in.
-    # @param separator [String] (', ') The value to join with.
-    def link_to_facet_list(values, solr_field, empty_message = "No value entered", separator = ", ")
+    # @param values [Array{String}] strings to display
+    # @param solr_field [String] name of the solr field to link to, without its suffix (:facetable)
+    # @param empty_message [String] message to display if no values are passed in
+    # @param separator [String] value to join with
+    # @return [ActiveSupport::SafeBuffer] the html_safe link
+    def link_to_facet_list(values, solr_field, empty_message = "No value entered".html_safe, separator = ", ")
       return empty_message if values.blank?
```
mjgiarlo: Any reason not to move the `#html_safe` call here (vs. in the method signature)?

app/helpers/sufia/sufia_helper_behavior.rb
```diff
@@ -70,42 +72,52 @@ def number_of_deposits(user)
...
     end
 
-    def index_field_link(options = {})
-      field_name = options[:config][:field_name]
+    # A Blacklight helper_method
+    # @param options [Hash] Blacklight sends :document, :field, :config, :value and whatever else was in options
+    # @option options [Hash] :config including {:field_name => "my_name"}
+    # @option options [Array{String}] :value
+    # @return [ActiveSupport::SafeBuffer] the html_safe link
+    def index_field_link(options)
```
mjgiarlo: Why change `options` to a required parameter?

app/helpers/sufia/sufia_helper_behavior.rb
```diff
@@ -70,42 +72,52 @@ def number_of_deposits(user)
...
-      params = { search_field: 'advanced', name => "\"#{value}\"" }
-      state = search_state.params_for_search(params)
-      link_to(name, main_app.search_catalog_path(state))
+      safe_join(values.map { |item| link_to_field(name, item, item) }, ", ".html_safe)
     end
 
-    # @param [String,Hash] text either the string to escape or a hash containing the
-    #                           string to escape under the :value key.
+    # @param text [String,Hash] string to escape or a hash containing that string under the :value key.
+    # @param showLink [Boolean]
```
mjgiarlo: While you're tweaking this, could you change `showLink` to `show_link`?

sufia.gemspec
```diff
@@ -31,6 +31,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'select2-rails', '~> 3.5.9'
   spec.add_dependency 'json-schema'
   spec.add_dependency 'oauth'
+  spec.add_dependency 'devise', '~> 4.2'
```
mjgiarlo: How is this change related?

mjgiarlo:
Overall, very glad to see this PR. Have a few questions and comments.

app/helpers/sufia/sufia_helper_behavior.rb
atz: Why allow the variable to be assigned in a state that you never want to use?

mjgiarlo: I'm more wondering if we might want to call `#html_safe` on user-provided values of `empty_message` too.

sufia.gemspec
atz: There is a deprecation warning from spec_helper in certain versions of devise.  The warning references an alternative invocation that I used, but the alternative invocation is not available on *all* versions of devise.  This specification allows the deprecation to be avoided, but also not crash the build.

.rubocop.yml
atz: Contrived example:
```ruby
expect{object.method_a}.not_to raise_error
expect(object.method_b).to eq 5
```
vs.
```ruby
expect { object.method_a }.not_to raise_error
expect(object.method_b).to eq 5
```
I prefer the alignment.  Making a "word" of `}.not_to` feels off to me (and adding a space is worse).

sufia.gemspec
mjgiarlo: I'm -0 on specifying this change, which has an effect on all downstream apps, to avoid a deprecation warning in the test suite. How about we go the Blacklight route here? https://github.com/projectblacklight/blacklight/blob/master/spec/spec_helper.rb#L56-L61

app/helpers/sufia/sufia_helper_behavior.rb
atz: Because you were going to throw an exception in the next line anyway.  In practice, it was already required.

.rubocop.yml
mjgiarlo: Yeah? I prefer the latter example to the former, which helps recognize the distinction between parentheses and braces, and that is our current practice in Sufia (and CurationConcerns and ActiveFedora, too, for instance). There are 0 uses of `expect{` in these codebases.

app/helpers/sufia/sufia_helper_behavior.rb
mjgiarlo: :ok_hand: I'm cool with this.

app/helpers/sufia/sufia_helper_behavior.rb
atz: I'm fine with allowing the user to get back the exact argument they supplied.  I do *not* want to call html_safe on a string that may or may not contain HTML, therefore may provide a vector for injection.  I would oppose this PR making the base case of this method just a blind `html_safe` call. It is possible that calling **`html_escape`** might improve the consistency of the return, but I don't want to encourage building html and passing it to this method.  (Honestly, I think the logic in this helper method is a pain in the ass overall.  The message should be in a view and translated, for example.)

mjgiarlo: Good points, all. :+1:

.rubocop.yml
jcoyne: Yep, keep the spaces until we have a discussion about changing the agreed upon style.  I'm personally happy with the existing convention.

sufia.gemspec
atz: I opened a related ticket:
https://github.com/projectblacklight/blacklight/issues/1428

I think that conditional is inapplicable.  Our version of rails didn't change, but version of devise did.  I would fix Blacklight to specify a SemVer of devise that didn't require such a janky workaround.

mjgiarlo: Fair point.

More broadly... ScholarSphere's `sufia7` branch is using devise 3.5.10. I don't see why fixing a deprecation warning in Sufia's test suite should force every Sufia adopter to update their devise dependency to `~> 4.2`.

.rubocop.yml
atz: Obviously, there are no cases because the rubocop/travis rules would fail the build on it.  So we don't need to grep to know that.  Otherwise, we wouldn't be having this conversation.  

When I'm reading specs, differentiation between paren and curly following `expect` is a non-issue.  It's not something I need to highlight.  But the difference between `method_a` and `method_b` is.  The formatting guideline comes from writing ruby blocks and hashes, not this particular invocation inside rspec DSL.  

As for extending community standards, how about starting with Blacklight where AllCops are excluded for all specs?  Here I'm covering concretely failing methods and correcting expectations that were never once legitimate.  I'm not going to go to war over whitespacing two lines, but this is kind of thing where it feels trifling.

app/helpers/sufia/sufia_helper_behavior.rb
```diff
@@ -70,42 +72,52 @@ def number_of_deposits(user)
...
     def iconify_auto_link(text, showLink = true)
       text = presenter(text[:document]).render_values(text[:value], text[:config]) if text.is_a? Hash
```
atz: I need some help with this line, since Blacklight broke SemVer and removed `.render_values`.  @jcoyne recommended `.field_value` as an alternative, and while that method exists, it fails in full tests with:
```
  1) BlacklightHelper render_index_field_value metadata index links description_tesim 
     Failure/Error: text = presenter(text[:document]).field_value(text[:value], text[:config]) if text.is_a? Hash
     
     NoMethodError:
       undefined method `each_pair' for ["This links to http://example.com/ What about that?"]:Array
     # /Users/atz/.rvm/gems/ruby-2.2.0/gems/blacklight-6.3.0/lib/blacklight/configuration/null_field.rb:9:in `initialize'
     # /Users/atz/.rvm/gems/ruby-2.2.0/gems/blacklight-6.3.0/app/presenters/blacklight/index_presenter.rb:66:in `new'
     # /Users/atz/.rvm/gems/ruby-2.2.0/gems/blacklight-6.3.0/app/presenters/blacklight/index_presenter.rb:66:in `block in field_config'
     # /Users/atz/.rvm/gems/ruby-2.2.0/gems/blacklight-6.3.0/app/presenters/blacklight/index_presenter.rb:66:in `fetch'
     # /Users/atz/.rvm/gems/ruby-2.2.0/gems/blacklight-6.3.0/app/presenters/blacklight/index_presenter.rb:66:in `field_config'
     # /Users/atz/.rvm/gems/ruby-2.2.0/gems/blacklight-6.3.0/app/presenters/blacklight/index_presenter.rb:44:in `field_value'
     # ./app/helpers/sufia/sufia_helper_behavior.rb:122:in `iconify_auto_link'
```

This String-or-Hashness of the 1st argument is problematic.

sufia.gemspec
atz: No current implementation is forced to do anything.  Sufia 7 is unreleased and still in beta, so the time to modify dependencies is **before release**.  This allows the released version to have a longer lifespan and more consistent profile.  ScholarSphere's tracking *might* be an argument if if Sufia 7 were actually released, or if this PR were targeting a `6.x.x` bugfix, but for now it is just a question of whether consumers want to be tracking the leading edge of development or not.  If a project can't bump deps with a major version release, then it can't do it at all.  

If it were up to me, I would go through and bump up as many dependencies major versions as possible before release.  The release is happening *now* and should use components that are current now.  I.e., get this badge green, or closer to it (especially the oauth gem!): https://gemnasium.com/projecthydra/sufia

Specifically, this would also surface deprecations that we could clean up now rather than allow to creep in with a later `bundle update` or fresh install.

app/helpers/sufia/sufia_helper_behavior.rb
mjgiarlo: @atz It looks like @awead's PR would also change this line. What do you think of the solution he used?

https://github.com/projecthydra/sufia/pull/2318/files#diff-dd1cb84d4202ddaa5d482ce3da352033R110

(And if you can include `s/showLink/show_link/g` in your next push, that'd be `:gravy:`.

sufia.gemspec
mjgiarlo: We do have Sufia 7 releases out and people are using them. If someone's gotten their app upgraded to use `7.0.0.beta4` and they are using devise 3.5, I don't want to complicate their path to 7.0.0 proper. I don't disagree with where you're coming from, though.

atz: We have v7 betas only.  I'm distinguishing between a beta and a fully released version.  People shouldn't get too attached to betas.  Tracking your app against a beta dependency is an **experiment**.  If you want to do it, cool, you can give feedback and peer into a possible future.  But there is materially no social (or SemVer) contract of what you see in that beta being preserved in a production release.  Not features, not APIs, not UI, not documentation and not dependencies.  The fact of having conducted an experiment does not induce this expectation in the project, even if it has that psychological effect on the experimenter.  There may be a dozen more betas between that version and release.  The very next beta may totally alter or erase whatever I looked at in the last one.  In fact, that project in beta may *never* see a proper release at all (e.g. indie video games, hopefully not us!).  

If we were between release candidates instead, I might feel differently, but we aren't.

app/helpers/sufia/sufia_helper_behavior.rb
atz: That looks fine, and I wish I would have got that answer when I started asking about what to replace the removed method with last week.

jcoyne:
This PR needs to be rebased.

sufia.gemspec
cbeer: > I don't see why fixing a deprecation warning in Sufia's test suite should force every Sufia adopter to update their devise dependency to ~> 4.2.

👏  since the application works fine with older versions of devise, forcing people to upgrade to save ourselves a couple lines of (test-only) code doesn't seem worth the inconvenience it can cause downstream. Now, if it were a functional change to production code, I might be more wary of trying to maintain compatibility, but this is entirely sufia-developer facing and doesn't strike me as an undue burden.

atz: The application also hasn't released a new major version in over a year.  Consumers cannot reasonably complain "you bumped a dependency in a major version after a year!"  And if you cannot bump deps with a major version, you literally can **never** do it.  There are no doubt plenty of other dependencies that have jumped major versions in that time with breaking changes that you don't seem to care about.  You might as well complain that `hydra-batch-edit` has bumped a version or that CurationConcerns was introduced *at all*.  Run `git diff v6.0.0 sufia.gemspec` to see everything that's been allowed to change since the last new major version.  They still have to navigate that and this is supposed to signify a **bigger** change, but then requiring an update to Devise is somehow unthinkable?  Not feeling it.

jcoyne: I don't think we should list transitive dependencies here. I don't want to have to release a new version of Blacklight *AND* Sufia, when Devise decides to do an update.

spec/helpers/sufia_helper_spec.rb
```diff
@@ -1,3 +1,7 @@
+def new_state
```
mjgiarlo: What would you think of reworking this as:

```ruby
let(:new_state) { Blacklight::SearchState.new({}, CatalogController.blacklight_config) }
```

around line 23?

atz: I agree that would be better, but it would change the semantics where `new_state` is invoked twice in one test.  It was written as a method previously, and there are only so many turds I care to polish as ancillary work to my actual purpose.  Previously it was buried in the middle of a test definition, I pulled it out so that people would notice it and say, as you have, "hey, that doesn't belong there!"  But I'm not entertaining fixing that now.

There are similar instances in various test files, btw.  If you want to ticket improving the rspec conventions of the set, I could conceivably work on them as a group.

atz: see `def create_models` in the same file, for example.

mjgiarlo: I see. I grepped for `new_state` and didn't see it in the code before, but now I see you reworked it. OK, I'm cool with this.

hackmastera:
shout out to @atz for adding comments!!

PR #2320, opened 2016-07-05
Adding executable to run export actor from the command line

### LINKED ISSUE #2136: Export Metadata from Sufia 6.x to prepare for migration

cam156:
Fixes #2136  for Sufia 6

Create an executable to run for exporting models from the Sufia 6 repository


exe/sufia_export
```diff
@@ -0,0 +1,65 @@
+#!/usr/bin/env ruby
+require 'rails'
+
+# get the current running directory & load the environment from there
+require ::File.join(Dir.pwd, 'config/environment')
```
mjgiarlo: Can you add instructions that clarify that `sufia_export` *must* be run from the application root, else the environment won't be picked up? Either that or use `Rails.root` or the like to find the environment.

sufia.gemspec
```diff
@@ -9,12 +9,14 @@ Gem::Specification.new do |gem|
...
 
   gem.files         = `git ls-files | grep -v ^sufia-models`.split($\)
+  gem.bindir        = 'exe'
   gem.executables   = gem.files.grep(%r{^bin/}).map{ |f| File.basename(f) }
   gem.test_files    = gem.files.grep(%r{^(test|spec|features)/})
   gem.name          = "sufia"
   gem.require_paths = ["lib"]
   gem.version       = version
   gem.license       = 'Apache2'
+  gem.executables   = gem.files.grep(%r{^exe/}) { |f| File.basename(f) }
```
mjgiarlo: This should replace line 13.

exe/sufia_export
```diff
@@ -0,0 +1,65 @@
...
+
+  opts.on("-v", "--[no-]verbose", "Run verbosely") do |v|
+    options[:verbose] = v
+  end
+
+  opts.on("--limit NUMBER_OF_OBJECTS", Integer, "Export a maximum number of objects (defaults to all)") do |v|
+    options[:limit] = v
+  end
+
+  opts.on("--ids ID_LIST", Array, "Expoprt only the objects in the list of IDs") do |v|
```
mjgiarlo: `s/Expoprt/Export/`

Can you clarify in the help text what `ID_LIST` is supposed to look like? The example you provide for `MODEL_LIST` is a good model

exe/sufia_export
```diff
@@ -0,0 +1,65 @@
...
+    options[:ids] = v
+  end
+
+  opts.on("--models MODEL_LIST", Array, "Specify the models to be output (defaults to [Collection,GenericFile]",
+          "To override the default converter for a model or to register a new converter add the converter after the model.",
+          "For example: --models \"GenericFile Mine::NewGenericFileConverter\"",
```
mjgiarlo: So if I want to pass in two models with new converters, would I invoke `sufia_export` like this?

```bash
$ sufia_export ... --models [Collection Mine::NewCollectionConverter,GenericFile Mine::NewGenericFileConverter]
```

Am unclear on whether `MODEL_LIST` is supposed to look like a Ruby array or a string or what.

sufia.gemspec
cam156: I was following the location of the executables, but now that you point out line 13 I will move my executable to bin.

awead: @mjgiarlo @cam156 "exe" is the new location for gem executables so that it keeps it separate from the other like rake and rails. Are we saying we don't want to adopt the new convention or is there another reason we're sticking with the old way? http://bundler.io/blog/2015/03/20/moving-bins-to-exe.html

exe/sufia_export
cam156: @mjgiarlo The Rails.root is nil at that point.  After much digging I found Bundler.default_gemfile which I will change to using.

cam156:
@mjgiarlo @atz Changed up the help text here is the full output.  Does it make more sense?
```
Exports model metdata from fedora to json files in tmp/export. Usage: sufia_export [options]
    -v, --[no-]verbose               Run verbosely
        --limit NUMBER_OF_OBJECTS    Export a maximum number of objects (defaults to all)
        --ids ID_LIST                Export only the objects in the list of IDs
                                     For example to export a Collection with the id abc123 and a GenericFile with an id def456:
                                       --ids abc123,def456
        --models MODEL_LIST          Specify the models to be output (defaults to GenericFile,Collection)
                                     To override the default converter for a model or to register a new converter add the converter after the model.
                                     For example to convert GenericFiles with a new converter and Collections with the default converter:
                                       --models GenericFile=Mine::NewGenericFileConverter,Collection
```

cam156:
@awead @mjgiarlo The location of the executables in exe or bin is ok with me.  I can add a commit to this PR to move them to exe, or we could ticket moving the set since that seems a bit out of scope...

cam156:
@mjgiarlo actually there is a bin directory in sufia 6: https://github.com/projecthydra/sufia/tree/6.x-stable/bin

mjgiarlo:
@cam156 Ah, good catch. Then that file will need to move into the new `exe` directory too or it will no longer be included in the list o' executables.

cam156:
@awead & @mjgiarlo Added a second commit to move the executable out of the bin directory.  Good?

mjgiarlo:
:clap: 

Thanks, @cam156!

PR #2324, opened 2016-07-06
Adding ActiveRecord class for Sufia::Migration::Survey::item

### LINKED ISSUE #2131: Migration::Survey::Item Class
Create an ActiveRecord class to save and update a Migration::Survey::Item in the database.  This class is needed in Sufia 6.x and Sufia PCDM applications. See #2128 and #2129 for use of this class.

This class should have the following attributes:
 * object_id
 * type
 * title (for debug purposes)
 * status

This will be done when I can:
* Migration::Survey::Item.new # creates instance of object
* attributes are available on an instance
* save to store an instance in database
* find by object_id to retrieve an instance

cam156:
Fixes #2131

Adding an ActiveRecord Class so we can create a survey of our ActiveFedora objects before we migrate.  This survey will be used to verify that all our objects made it after the migration has completed.

### DIFF
--- a/db/migrate/20160706132422_create_sufia_migration_survey_items.rb
+++ b/db/migrate/20160706132422_create_sufia_migration_survey_items.rb
@@ -1,0 +1,12 @@
+class CreateSufiaMigrationSurveyItems < ActiveRecord::Migration
+  def change
+    create_table :sufia_migration_survey_items do |t|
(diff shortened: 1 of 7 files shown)


mjgiarlo:
LGTM! Other @projecthydra/sufia-code-reviewers ?

PR #2332, opened 2016-07-07
Pass an id to FileSetAuditService, ref #2326

### LINKED ISSUE #2326: Enable Solr-only viewing of FileSets
### Descriptive summary

We're able to view the work page by only using Solr, and not having to make any calls to Fedora. Why not make this happen for the file set pages as well?

Currently, there are at least two calls to Fedora, one for the parent object, such as the containing work, and another call for the file set, to obtain all the versions of the file for the audit trail.

It would nice if we didn't need those Fedora calls and could load the page exclusively from Solr.

### Details

This call to `FileSet.find(id)` causes the file set to load from Solr:

https://github.com/projecthydra/sufia/blob/master/app/presenters/sufia/file_set_presenter.rb#L49

We can remove the parent to load from Fedora in #2331

awead:
Fixes #2326 

FileSet show page now made with 💯  Solr!

NB: Requires a CC 1.1.1 release.

### DIFF
--- a/app/presenters/sufia/file_set_presenter.rb
+++ b/app/presenters/sufia/file_set_presenter.rb
@@ -45,9 +45,7 @@ module Sufia
     end
 
     def audit_service
-      # model = solr_document.to_model # See https://github.com/projecthydra-labs/hydra-pcdm/issues/197
-      model = FileSet.find(id)
-      @audit_service ||= CurationConcerns::FileSetAuditService.new(model)
+      @audit_service ||= CurationConcerns::FileSetAuditService.new(id)
     end
   end
 end


Gemfile
```diff
@@ -3,6 +3,8 @@ source 'https://rubygems.org'
 # Please see sufia.gemspec for dependency information.
 gemspec
 
+gem 'curation_concerns', github: 'projecthydra/curation_concerns', branch: 'master'
```
jcoyne: Let's resolve this before merging the PR.

spec/test_app_templates/Gemfile.extra
```diff
@@ -1,5 +1,7 @@
 # Use this file to reference specific commits of gems.
 
+gem 'curation_concerns', github: 'projecthydra/curation_concerns', branch: 'master
```
jcoyne: Let's resolve this before merging the PR.

Gemfile
jcoyne: Just as a note to @awead, I don't think there's any need for this if you put it in Gemfile.extra.

awead:
@jcoyne agreed on all points above. This is just to have Travis run it to see if I missed anything.

PR #2333, opened 2016-07-08
Remove deprecated Devise::TestHelpers

awead:


jcoyne:
@awead I think you need to do something like this: https://github.com/projectblacklight/blacklight/commit/de1499b6f82776f642f27aac099942577960777d

cbeer:
Only better; either by version-checking devise, or checking if the new class exists?

mjgiarlo:
:tada:

PR #2335, opened 2016-07-08
Adding FedoraIdService to create a list of all the ids in fedora

### LINKED ISSUE #2132: Migration::Survey::FedoraIdService
Create a class To walk the Fedora 4 Sufia 6.x repository to gather all of the ids of Sufia's default objects (Collection and GenericFile). See #2128 for full use case.

* register_model - register an additional object type beyond Collection and GenericFile
  * parameters: - Object.class
* call - produce a list of ids for the registered models
  * parameters: limit_number defaults to all
  * return: a list of IDs 

This class should have a model registry that is initialized with Collection.class & GenericFile.class

This is done when I can:
* Migration::Survey::FedoraIdService.new and create an instance of the survey
* instance.register_model and add additional models to the registry
* instance.call produces a list of ids in the fedora 4 repository limited by the registerd models and optionally limited by the limit number

cam156:
Fixes #2132

This service creates a list of ids for all the top level objects in a fedora repository.  By default top level object is defined as GenericFiles and Collections.  You may register additional top level objects before you call the service.

### DIFF
--- a/lib/sufia/migration/survey/fedora_id_service.rb
+++ b/lib/sufia/migration/survey/fedora_id_service.rb
@@ -1,0 +1,62 @@
+# Class to extract all the ids from fedora for registered classes.  By default GenericFile and Collection are registered.
+#
+module Sufia
+  module Migration
+    module Survey
+      class FedoraIdService
+        attr_accessor :model_registry
+
+        # initialize the service with the default models (GenericFile & Collection) registered
+        def initialize
+          @model_registry = default_registry
+        end
+
+        # regesiter an additional ActiveFedora Model to extract ids for
+        #
+        # @param [Class] model_class additional class that you would like to be in the output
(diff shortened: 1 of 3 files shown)


lib/sufia/migration/survey/fedora_id_service.rb
```diff
@@ -0,0 +1,62 @@
...
+        # @raise [RegistryError] if the class is not an ActiveFedora based class
+        def register_model(model_class)
+          raise(RegistryError, "Model (#{model_class.name}) for conversion must be an ActiveFedora::Base") unless model_class.ancestors.include?(ActiveFedora::Base)
+          @model_registry << model_class
+        end
+
+        # returns a list of ids for all the registered classes in the repository
+        #
+        # @param [Number] limit limits the number of results (default is all)
+        def call(limit = -1)
```
jcoyne: why `-1` why not `nil` or `:all`?

PR #2340, opened 2016-07-08
Adding a surveyor to create survey items for an id list

### LINKED ISSUE #2133: Migration::Survey::Surveyor Class
Create a class that surveys a fedora 4 instance of a Sufia 6 application based on an id list to generate a Migration::Survey::Item #2131 for each id in the list. See #2128 for full use case.

* call - loops through the ids generating items for each id
  * parameter: id_list   # an array of ids to process

This is done when I can:
* Migration::Survey::Surveyor.call produces a Migration::Survey::Item for each id in the input list

cam156:
fixes #2133 

Creates a Sufia::Migration::Survey::Item for each id passed into the call.  If any id does not exists the entire call should fail.

The id list is generated by the code in #2335.

### DIFF
--- a/lib/sufia/migration/survey/item.rb
+++ b/lib/sufia/migration/survey/item.rb
@@ -3,15 +3,16 @@
 # @attr [String] object_id fedora id of the object being migrated
 # @attr [String] object_class fedora class of the object being migrated (Collection, GenericFile)
 # @attr [String] object_title title of the object being migrated
-# @attr [int]    migration_status - Status of the object's migration
-# @option migration_status -1 initial state before migration
-# @option migration_status 0 migrated successfully
-# @option migration_status 1 Missing when verified
-# @option migration_status 2 Migrated but wrong type
+# @attr [enum]   migration_status - Status of the object's migration
+# @option migration_status :initial_state initial state before migration
+# @option migration_status :successful migrated successfully
+# @option migration_status :missing Missing when verified
+# @option migration_status :wrong_type Migrated but wrong type
 module Sufia
   module Migration
     module Survey
       class Item < ActiveRecord::Base
+        enum migration_status: [:initial_state, :successful, :missing, :wrong_type]
       end
     end
   end
--- a/lib/sufia/migration/survey/surveyor.rb
+++ b/lib/sufia/migration/survey/surveyor.rb
@@ -1,0 +1,19 @@
+# Class that will survey Fedora based on a list of ids
+module Sufia
+  module Migration
+    module Survey
+      class Surveyor
+        class << self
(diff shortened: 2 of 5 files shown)


lib/sufia/migration/survey/surveyor.rb
```diff
@@ -0,0 +1,19 @@
...
+  module Migration
+    module Survey
+      class Surveyor
+        class << self
+          # call causes the surveyor to create a survey item for each id in the list
+          #
+          # @param [Array] id_list a list of ids to be surveyed
+          def call(id_list)
+            ActiveFedora::Base.find(id_list).each do |object|
+              Item.create(object_id: object.id, object_class: object.class, object_title: object.title, migration_status: -1)
```
jcoyne: It seems that meaning has been encoded in `migration_status`, but it's not clear to me what `-1` represents.

jcoyne: Does it make sense to encode this in an [ActiveRecord::Enum](http://api.rubyonrails.org/v4.1/classes/ActiveRecord/Enum.html)?

awead:
@cam156 needs a rebase.

lib/sufia/migration/survey/item.rb
```diff
@@ -3,15 +3,16 @@
...
+# @attr [enum]   migration_status - Status of the object's migration
+# @option migration_status :initial_state initial state before migration
+# @option migration_status :successful migrated successfully
+# @option migration_status :missing Missing when verified
+# @option migration_status :wrong_type Migrated but wrong type
 module Sufia
   module Migration
     module Survey
       class Item < ActiveRecord::Base
+        enum migration_status: [:initial_state, :successful, :missing, :wrong_type]
```
jcoyne: 👍 Much better!

spec/lib/sufia/migration/survey_item_spec.rb
```diff
@@ -7,13 +7,13 @@
...
   context "with specific values" do
-    let(:item) { described_class.create(object_id: "myid", object_class: "MyClass", object_title: "My Title", migration_status: 99) }
+    let(:item) { described_class.create(object_id: "myid", object_class: "MyClass", object_title: "My Title", migration_status: :successful) }
     subject { described_class.find(item.id) }
     it "stores the attributes" do
       expect(subject.object_id).to eq "myid"
       expect(subject.object_class).to eq "MyClass"
       expect(subject.object_title).to eq "My Title"
-      expect(subject.migration_status).to eq 99
+      expect(subject.successful?).to be_truthy
```
jcoyne: this works, but it could also be `expect(subject).to be_successful`

PR #2341, opened 2016-07-08
Remove spec that tests Blacklight, not Sufia

jcoyne:


awead:
@jcoyne do we have any integration tests that verify the facet search links work?

jcoyne:
I don't know, but if we merge #2343, we won't even be using the `render_index_field_value` method from Blacklight.

jcoyne:
In other words, not merging this PR, does not get us any closer to having an integration spec.

mjgiarlo:
@awead @jcoyne I'm in favor of merging, but you raise good points about integration testing. Perhaps one of you could create a new issue for this, so we don't blindly trust Blacklight or lose track of this need?

awead:
@jcoyne makes sense. so the doc_presenter in #2343 is now taking care of rendering the facet links, so having this test here is going to be superfluous anyway, right?

jcoyne:
@awead that's correct.

awead:
OK, I've put this here to remind us: #2345

atz:
This is ill advised.  This tests **sufia's configuration of blacklight** and it represents tests that were directly cut out of different file and moved here.  The remaining tests are entirely superficial and do not test the different field configurations, helper_methods, itemprops, etc.  

Is preferable to REVERT to retain history rather than me adding this back in as a branch.  The fact that there is more/better integration testing to do doesn't get you out of actually addressing the cases that are covered here.

PR #2352, opened 2016-07-11
Simplify iconify_auto_link

jcoyne:

### DIFF
--- a/app/helpers/sufia/sufia_helper_behavior.rb
+++ b/app/helpers/sufia/sufia_helper_behavior.rb
@@ -132,7 +132,10 @@ module Sufia
     # @return [ActiveSupport::SafeBuffer]
     # @todo stop being a helper_method, start being part of the Blacklight render stack?
     def iconify_auto_link(text, show_link = true)
-      text = index_presenter(text[:document]).field_value(text[:field].to_sym, text[:config]) if text.is_a? Hash
+      if text.is_a? Hash
+        options = text[:config].separator_options || {}
+        text = text[:value].to_sentence(options)
+      end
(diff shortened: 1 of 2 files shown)


jcoyne:
@awead what do you think of this? Am I missing anything?

app/helpers/sufia/sufia_helper_behavior.rb
```diff
@@ -132,7 +132,7 @@ def index_field_link(options)
     # @return [ActiveSupport::SafeBuffer]
     # @todo stop being a helper_method, start being part of the Blacklight render stack?
     def iconify_auto_link(text, show_link = true)
-      text = index_presenter(text[:document]).field_value(text[:field].to_sym, text[:config]) if text.is_a? Hash
+      text = text[:value] if text.is_a? Hash
```
cbeer: Do we need to worry about multivalued fields?

atz:
There are some pending tests that now might pass if this works.  Or if they don't, should be revisited for whether they are considered correct.  Basically iconify was not testable for legitimate Hash args.

awead:
@jcoyne you know, that's what I initially tried but I thought to myself, that can't be right... there has to be some Blacklight voodoo magick that I'm skipping. So yeah, 👍. You can do a join for multivalued fields.

jcoyne:
@awead I've updated to handle arrays. It appears that's whats coming through with these `_tesim` fields.

awead:
@jcoyne looks good. Should there be a separator for the join? Just a space?

jcoyne:
@awead sure, separator added

PR #2353, opened 2016-07-12
Adding executable to run surveyor from the command line.

### LINKED ISSUE #2128: Sufia 6.x_stable Survey
I would like to be able to survey my Sufia 6.x repository prior to migration, so that I can verify that every object has been moved.  Part of user story #2127 

I would like to create a executable I can run that creates a database table with survey information.

This will be done when I can run sufia_survey.exe in my sufia 6.x  application and I produce a database table that contains a list of my object ids and types [Collections & Files].

My executable should take the following options:
  --limit= [all (default), # of objects,]   # limit the number of items surveyed
  --id=  a list of Id (id1,id2)                 # only survey the given ids

cam156:
Fixes #2128

Adds an executable to run the surveyor from the command line.  In addition this PR fixes a "bug" in the surveyor, to update an Survey::Item instead of always creating one.  That way if you run the sufia_survey 3 times you only get one item per Fedora object instead of 3.

### DIFF
--- a/exe/sufia_survey
+++ b/exe/sufia_survey
@@ -1,0 +1,87 @@
+#!/usr/bin/env ruby
+require 'rails'
+
+require ::File.join(File.dirname(Bundler.default_gemfile), 'config/environment')
+
+options = {}
+
+subtext = <<HELP
+Commonly used commands are:
+   clean :     cleans all Sufia::Migration::Survey::Items from the database
+See 'sufia_survey COMMAND --help' for more information on a specific command.
+HELP
+
+args = OptionParser.new do |opts|
+  opts.banner = "Survey model objects in fedora and store survey results in an ActiveRecord table. Usage: sufia_survey [options]"
+
+  opts.on("-v", "--[no-]verbose", "Run [not] verbosely") do |v|
+    options[:verbose] = v
+  end
+
+  opts.on("--limit NUMBER_OF_OBJECTS", Integer, "Survey a maximum number of objects (defaults to all)") do |v|
+    options[:limit] = v
+  end
+
+  opts.on("--ids ID_LIST", Array, "Survey only the objects in the list of IDs",
+          "For example to survey a Collection with the id abc123 and a GenericFile with an id def456:",
+          "  --ids abc123,def456") do |v|
+    options[:ids] = v
+  end
+
+  opts.on("--models MODEL_LIST", Array, "Specify the models to be surveyed (defaults to GenericFile,Collection)") do |v|
+    options[:models] = v
+  end
+
+  opts.separator ""
+  opts.separator subtext
+end
+
+subcommands = {
+  'clean' => OptionParser.new do |opts|
+    opts.banner = "Usage: sufia_survey clean"
+  end
+}
+
+begin
+  args.order!
+  command = ARGV.shift
+  if command
+    raise "Bad subcommand: #{command}" unless subcommands[command]
+    subcommands[command].order!
+  end
+  args.parse!
+rescue => error
+  $stderr.puts "ERROR: #{error}\n"
+  $stderr.puts args.help
+  exit 1
+end
+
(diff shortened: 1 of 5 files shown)


exe/sufia_survey
```diff
@@ -0,0 +1,91 @@
+#!/usr/bin/env ruby
+require 'rails'
+
+require ::File.join(File.dirname(Bundler.default_gemfile), 'config/environment')
+
+options = {}
+
+subtext = <<HELP
+Commonly used command are:
```
mjgiarlo: `s/command/commands/`

exe/sufia_survey
```diff
@@ -0,0 +1,91 @@
...
+   clean :     cleans all Sufia::Migration::Survey::Items from the database
+See 'sufia_survey COMMAND --help' for more information on a specific command.
+HELP
+
+args = OptionParser.new do |opts|
+  opts.banner = "Survey model objects in fedora and store it in an ActiveRecord table. Usage: sufia_survey [options]"
```
mjgiarlo: Store what? The results of the survey? Maybe reword as `s/store it/store survey results/`?

exe/sufia_survey
```diff
@@ -0,0 +1,91 @@
...
+
+  opts.on("-v", "--[no-]verbose", "Run verbosely") do |v|
```
mjgiarlo: So `--no-verbose` and `--verbose` both trigger verbose output? Maybe I misunderstand what OptionParser is doing here.

exe/sufia_survey
```diff
@@ -0,0 +1,91 @@
...
+    options[:limit] = v
+  end
+
+  opts.on("--ids ID_LIST", Array, "Survey only the objects in the list of IDs",
+          "For example to survey a Collection with the id abc123 and a GenericFile with an id def456:",
+          "  --ids abc123,def456") do |v|
+    options[:ids] = v
+  end
+
+  opts.on("--models MODEL_LIST", Array, "Specify the models to be surveyed (defaults to GenericFile,Collection)") do |v|
```
mjgiarlo: Looks like lines 31-33 are identical to lines 35-37.

exe/sufia_survey
```diff
@@ -0,0 +1,91 @@
...
+  end
+  args.parse!
+rescue => error
+  $stderr.puts "ERROR: #{error}\n"
+  $stderr.puts args.help
+  exit 1
+end
+
+if command == 'clean'
+  $stderr.puts "cleaning Sufia::Migration::Survey::Item"
```
mjgiarlo: `s/Item/Items/` ?

mjgiarlo: Should this go to `$stdout` instead of `$stderr`?

exe/sufia_survey
```diff
@@ -0,0 +1,91 @@
...
+
+$stdout.puts "running with #{options}" if verbose
+id_service = Sufia::Migration::Survey::FedoraIdService.new
+
+if options[:models]
+  model_options = options.delete(:models)
+  $stdout.puts "processing models #{model_options}" if verbose
+  model_options.each do |model_str|
+    model_class = Object.const_get(model_str)
+    id_service.register_model(model_class) unless id_service.model_registry.keys.include?(model_class)
```
mjgiarlo: I might like to see the ID service have a method so you could simplify this to `id_service.registered?(model_class)` but I don't care *too* much about that.

exe/sufia_survey
cam156: @ mjgiarlo the [no] is essentially a not, so no-verbose forces not verbose.  I actually copied this out of solr-wrapper so that the executables would be consistent across hydra.  I'll change the wording to include [not].

exe/sufia_survey
cam156: @mjgiarlo good 👀 .. copy pasta...

PR #2366, opened 2016-07-19
Render the batch create form

### LINKED ISSUE #2363: BatchUploads error in 7.0.0.rc1
I did a fresh install of 7.0.0.rc1 this morning, using these instructions:  https://github.com/projecthydra/sufia/tree/v7.0.0.rc1#creating-a-sufia-based-app

All is working so far, except for Batch Uploads... When I visit `batch_uploads/new`, I get:

```
NoMethodError in Sufia::BatchUploads#new
Showing /Users/vanmiljf/.rvm/gems/ruby-2.3.1@sufia-app/gems/sufia-7.0.0.rc1/app/views/curation_concerns/base/new.html.erb where line #1 raised:
undefined method `human_readable_type' for #<BatchUploadItem:0x007ff15630c628>
```

Prior to triggering these errors, I created two users, set up a proxy, and added two works.

awead:
Fixes #2363 

I'm not convinced this is the way to fix this, but it works.

### DIFF
--- a/app/models/batch_upload_item.rb
+++ b/app/models/batch_upload_item.rb
@@ -2,7 +2,9 @@
 # It should never actually be persisted in the repository.
 # The properties on this form should be copied to a real work type.
 class BatchUploadItem < ActiveFedora::Base
-  include CurationConcerns::BasicMetadata
+  include ::CurationConcerns::WorkBehavior
+  include ::CurationConcerns::BasicMetadata
+  include Sufia::WorkBehavior
 
   # This mocks out the behavior of Hydra::PCDM::PcdmBehavior
   def in_collection_ids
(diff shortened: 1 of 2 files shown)


mjgiarlo:
@awead can you comment here on what form the bug takes, and how this fixes it? Your solution looks good to me, but I'm puzzled as to how the code got into this broken state and eluded our test suite.

awead:
@mjgiarlo issue that's linked here summarizes it. The page just doesn't render because it's expecting things like `human_readable_type` on BatchUploadItem. I don't know why this wasn't happening before, but basically, because it's extending AR::Base, and only including BasicMetadata instead of the other bits that are included with GenericWork, adding the other modules in corrects the issue.

I'm further puzzled that simply adding these two, fairly large, modules into and AR::Base object works. I guess things are better than they seem!

jcoyne:
Have you attempted just mixing in `CurationConcerns::HumanReadableType`?

awead:
@jcoyne there were others beyond that, such as permissions. It was expecting BatchUploadItem to respond to permissions.

PR #2369, opened 2016-07-19
Making global javascript name longer so it does not conflict with other gems

cam156:
We are getting feature test error in ScholarSphere becuase for some reason the stat variable is getting set by another gem we are using and the flot javascript is erroring since it is set but not with the expected data.

Since it is a global variable, making the name more verbose solves the issue.


awead:
I think this problem originated from me, here https://github.com/projecthydra/sufia/blob/1caf7ab2dcb5317a1a5ab8cfa91131271bb27a7f/app/views/generic_files/_flot_stats.js.erb

I think the solution is fine for now, but we ought to make a follow-up issue that refactors this so it namespaced, and maybe update it to load via Blacklight or, or update it to EMCA6.

PR #2372, opened 2016-07-21
Verify we have a tab to select so we do not get a js error

cam156:
This code is for the tabs on the home page. On the dashboard the tabs are defined differently and showing a tab caused a javascript error.

To replicate the javascript error you need to got to the works listing, then click on the collections listing, then hit the back button and the error will show in the js console. One of the feature tests in ScholarSphere just happened to use the back button, which then showed the js error.

### DIFF
--- a/app/assets/javascripts/sufia/tabs.js
+++ b/app/assets/javascripts/sufia/tabs.js
@@ -1,22 +1,30 @@
+// This code is to implement the tabs on the home page
+
+// navigate to the selected tab or the first tab
+function tabNavigation(e) {
+    var activeTab = $('[href="' + location.hash + '"]');
+    if (activeTab.length) {
+        activeTab.tab('show');
+    } else {
+        var firstTab = $('.nav-tabs a:first');
+        // select the first tab if it has an id and is expected to be selected
+        if ((firstTab[0] !== undefined) && (firstTab[0].id != "")){
+          $(firstTab).tab('show');
+        }
+    }
+}
+
 Blacklight.onLoad(function () {
   // When we visit a link to a tab, open that tab.
   var url = document.location.toString();
   if (url.match('#')) {
     $('.nav-tabs a[href="#' + url.split('#')[1] + '"]').tab('show');
   }
-  
+
   // Change the url when a tab is clicked.
   $('a[data-toggle="tab"]').on('click', function(e) {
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/sufia/tabs.js
```diff
@@ -16,7 +17,11 @@ Blacklight.onLoad(function () {
     if (activeTab.length) {
       activeTab.tab('show');
     } else {
-      $('.nav-tabs a:first').tab('show');
+      var firstTab = $('.nav-tabs a:first');
+      // select the first tab if it has an id and is expected to be selected
+      if (firstTab[0].id != ""){
```
jcoyne: Why do we care if it has an id? Can we just check that the typeof is not undefined?

cam156: @jcoyne The item exists on the Work listing, but the id is set to empty string. So it is defined, just not set to a valid value.

mjgiarlo:
@cam156 Can you write a spec that triggers the javascript error you reference, so that if the code you've added is removed or changed, the suite fails?

cam156:
@mjgiarlo ok.  Will try to get that done tomorrow.  I can create a feature test that does it, since it was a feature that failed in SS, but I'll go for a JS test.

mjgiarlo:
A feature test would be OK, though likely slower than a JS test. Either would be sufficient to get this PR merged, IMO!

cam156:
@mjgiarlo Added a test.  I needed to move the javascript around a bit more to make it easier to test...

app/assets/javascripts/sufia/tabs.js
```diff
@@ -1,22 +1,30 @@
+// This code is to implement the tabs on the home page
+
+// navigate to the selected tab or the first tab
+function tabNavigation(e) {
+    $('.nav-tabs').addClass('blah');
```
mjgiarlo: Blah? 😄

cam156: Yeah blah...  I had made the change locally to remove that, but it did not get pushed...

mjgiarlo:
Looks good to me, @cam156! :watch: on Travis.

cam156:
@mjgiarlo travis is all 😃

PR #2406, opened 2016-08-08
Add work_file_required config, fixes #1715

### LINKED ISSUE #1715: Config work requires files
https://wiki.duraspace.org/display/hydra/New+Work+-+Metadata+Tab+-+Phase+1+-+UI+Wireframes#NewWork-MetadataTab-Phase1-UIWireframes-VariantsofSaveWorkwidget

Add new configuration option `config.work_requires_files`
  * if true, show in Requirements area and set to red exclamation icon if 0 files attached or green check if at least 1 file attached. Needs to be able to be changed if number of files changes.
  * if false, do not show in Requirements

Ref #1562

hackmastera:
Guidance on where testing is needed would be appreciated!

### DIFF
--- a/app/assets/javascripts/sufia/save_work/uploaded_files.es6
+++ b/app/assets/javascripts/sufia/save_work/uploaded_files.es6
@@ -5,6 +5,11 @@ export class UploadedFiles {
     $('#fileupload').bind('fileuploadcompleted', callback)
   }
 
+  get hasFileRequirement() {
+    let fileRequirement = this.form.find('li#required-files')
+    return fileRequirement.size() > 0
+  }
+
   get hasFiles() {
     let fileField = this.form.find('input[name="uploaded_files[]"]')
     return fileField.size() > 0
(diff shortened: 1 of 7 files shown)


app/assets/javascripts/sufia/save_work/uploaded_files.es6
```diff
@@ -2,6 +2,8 @@ export class UploadedFiles {
   // Monitors the form and runs the callback if any files are added
   constructor(form, callback) {
     this.form = form
+    this.fileRequirement = form.find('li#required-files')
+    this.hasFileRequirement = this.fileRequirement.size() > 0
```
awead: I think you could make `hasFileRequirement` a separate method which returns true or false, and then test it via Jasmine. It would look something like:
``` javascript
it("returns true", function() {
  up = new UploadFiles();
  loadFixtures("html_with_selector.html");
  expect(up.hasFileRequirement).toBe(true);
});

hackmastera:
I added the test recommended by @awead but it doesn't come up red before applying the code change so that's probably not good.

hackmastera:
@mjgiarlo this is ready.

PR #2458, opened 2016-08-10
Do not use deprecated #load_instance_from_solr method

### LINKED ISSUE #2446: load_instance_from_solr deprecated
During testing the warning is:
```
DEPRECATION WARNING: load_instance_from_solr is deprecated and will be removed from a future release. 
```
With two origins:
```
called from block in initialize_combined_fields at /Users/atz/repos/sufia/app/forms/sufia/forms/batch_edit_form.rb:31
called from to_model at /Users/atz/.rvm/gems/ruby-2.2.1/gems/curation_concerns-1.3.0/app/models/concerns/curation_concerns/solr_document_behavior.rb:26
```

mjgiarlo:
Fixes #2446

### DIFF
--- a/app/forms/sufia/forms/batch_edit_form.rb
+++ b/app/forms/sufia/forms/batch_edit_form.rb
@@ -26,9 +26,9 @@ module Sufia
         def initialize_combined_fields(batch)
           combined_attributes = {}
           permissions = []
-          # For each of the files in the batch, set the attributes to be the concatination of all the attributes
+          # For each of the files in the batch, set the attributes to be the concatenation of all the attributes
           batch.each do |doc_id|
-            work = model_class.load_instance_from_solr(doc_id)
+            work = model_class.find(doc_id)
             terms.each do |key|
               combined_attributes[key] ||= []
               combined_attributes[key] = (combined_attributes[key] + work[key]).uniq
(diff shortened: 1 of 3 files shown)


app/forms/sufia/forms/batch_edit_form.rb
```diff
@@ -26,9 +26,9 @@ def initialize(model, current_ability, batch)
         def initialize_combined_fields(batch)
           combined_attributes = {}
           permissions = []
-          # For each of the files in the batch, set the attributes to be the concatination of all the attributes
+          # For each of the files in the batch, set the attributes to be the concatenation of all the attributes
           batch.each do |doc_id|
-            work = model_class.load_instance_from_solr(doc_id)
+            work = model_class.find(doc_id).to_solr
```
jcoyne: I don't think you need the `to_solr` here, it's expecting an instance of `GenericWork` (or some kind of work)

mjgiarlo: Ah, I slapped this together without a second thought while the builds were broken. Assumed I needed a Solr doc here because of the hash-like `work[key]` invocation on line 34, but perhaps I overthought it. Thx.

PR #2468, opened 2016-08-15
Move child works from 'relationships' to 'items' on work show page

### LINKED ISSUE #2459: Integrate child works and child FileSets into a single list on work show page
### Descriptive summary
Sufia 7. Since these can be intermingled in ordering, they should display in the same area.

### Expected behavior
'Items' display should include fileset and works children.

### Actual behavior
'Items' area contains only fileset objects; child works are listed under 'relationships' as 'Has related work'

### Related work

#2321

hackmastera:
Fixes #2459


app/views/curation_concerns/base/_items.html.erb
```diff
@@ -1,17 +1,17 @@
 <h2><%= t('.header') %></h2>
-<% if presenter.file_set_presenters.present? %>
+<% if presenter.member_presenters.present? %>
   <table class="table table-striped related-files">
     <thead>
       <tr>
-        <th>File</th>
-        <th>Filename</th>
+        <th></th>
+        <th>Title</th>
```
cam156: It would be nice if this was I18N instead of hardcoded.

spec/views/curation_concerns/base/_items.html.erb_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+  end
+  let(:blacklight_config) { CatalogController.new.blacklight_config }
+  let(:blacklight_configuration_context) do
+    Blacklight::Configuration::Context.new(controller)
+  end
+
+  context "when children are present" do
+    before do
+      stub_template 'curation_concerns/base/_actions.html.erb' => 'Actions'
+      allow(presenter).to receive(:file_set_presenters).and_return([file_set])
```
cam156: @hackmastera Is this line needed now that the view uses member_presenter?

hackmastera: I guess not. The test wouldn't go red without it but i guess it makes sense to remove now.

cam156:
@hackmastera I'm trying to figure out what the column header that was removed looks like.  Could you possibly add in a screenshot before and after to the PR?

hackmastera:
sure I'll take a screenshot. It said 'File' and then the column contained thumbnails which link to the show page (same as the title)

hackmastera:
Columns before this PR:
screen shot 2016-08-16 at 11 44 26 am

After:
screen shot 2016-08-16 at 11 43 25 am

I could change it to "thumbnail" or something. I just realized this might be an accessibility error.

hackmastera:
@cam156 for reconsideration

cam156:
@hackmastera Looks good to me.

PR #2469, opened 2016-08-16
[WIP] remove outmoded from_field line

atz:
It doesn't mean anything to any other code in Sufia and it is inaccurate, since our 2 join queries have *different* from_fields.

### DIFF
--- a/app/search_builders/collection_search_builder.rb
+++ b/app/search_builders/collection_search_builder.rb
@@ -2,7 +2,6 @@
 class CollectionSearchBuilder < CurationConcerns::MemberSearchBuilder
   include BlacklightAdvancedSearch::AdvancedSearchBuilder
 
-  self.from_field = 'child_object_ids_ssim'
   self.default_processor_chain += [:include_contained_files]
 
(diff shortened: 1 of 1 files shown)


jcoyne:
I believe this is used here: https://github.com/projecthydra/curation_concerns/blob/master/app/search_builders/curation_concerns/member_search_builder.rb#L14

mjgiarlo:
Given what @jcoyne says, should we close this PR, @atz?

atz:
Recommend adding a test(!) and renaming the method to override the upstream one.  Actually, it is unclear why we are descending from a class where we want to ignore both attributes it touches and pre-empt its only method.  

Probably we should just descend from search builder and cut this CC piece out.

mjgiarlo:
@atz Could you write up an issue for that, or are you inclined to work on that in this PR?

atz:
Either way is fine.

mjgiarlo:
@atz How about we write this up as a separate technical debt ticket and close this PR, then? Thanks!

atz:
See #2480

PR #2476, opened 2016-08-16
Replace mail_form with an ordinary ActionMailer class

jcoyne:


app/models/sufia/contact_form.rb
```diff
@@ -0,0 +1,33 @@
...
+      ["Depositing content", "Depositing content"],
+      ["Making changes to my content", "Making changes to my content"],
+      ["Browsing and searching", "Browsing and searching"],
+      ["Reporting a problem", "Reporting a problem"],
+      ["General inquiry or request", "General inquiry or request"]
+    ].freeze
+
+    attr_accessor :contact_method, :category, :name, :email, :subject, :message
+    validates :email, :category, :name, :subject, :message, presence: true
+    validates :email, format: /\A([\w\.%\+\-]+)@([\w\-]+\.)+([\w]{2,})\z/i, allow_blank: true
```
cam156: @jcoyne I'm not sure we would want to allow blank.  How will we get back into contact with the user if they do not supply the email?

jcoyne: We're allowing blank here, because we're checking for presence on the previous line.  If we don't allow blank, the user would get two errors: "you must provide an email, email is an invalid format"

PR #2479, opened 2016-08-16
Support Rails 4 and Rails 5

jcoyne:


app/views/users/index.html.erb
```diff
@@ -16,7 +16,7 @@
         <% @users.each do |user| %> 
             <tr> 
               <td><%= link_to sufia.profile_path(user) do %>
-                    <%= image_tag user.avatar.url(:thumb), width: 30 %>
+                    <%= image_tag(user.avatar.url(:thumb), width: 30) if user.avatar.file %>
```
mjgiarlo: How's this related to Rails 4/5 support?

lib/generators/sufia/templates/catalog_controller.rb
```diff
@@ -4,9 +4,8 @@ class CatalogController < ApplicationController
   include Sufia::Catalog
   include BlacklightAdvancedSearch::Controller
 
-  # These before_filters apply the hydra access controls
-  before_filter :enforce_show_permissions, only: :show
-  skip_before_filter :default_html_head
```
mjgiarlo: This is no longer needed?

spec/spec_helper.rb
```diff
@@ -41,6 +47,8 @@
 Capybara.javascript_driver = :poltergeist # This is slower
 Capybara.default_max_wait_time = ENV['TRAVIS'] ? 30 : 15
 
+ActiveJob::Base.queue_adapter = :inline
```
mjgiarlo: Because the test suite assumes synchronous processing and the new default is `:async`?

spec/test_app_templates/Gemfile.extra
```diff
@@ -1,4 +1,7 @@
 # Use this file to reference specific commits of gems.
+gem 'breadcrumbs_on_rails', git: 'https://github.com/weppos/breadcrumbs_on_rails.git'
+gem 'blacklight', git: 'https://github.com/projectblacklight/blacklight'
```
mjgiarlo: Why do we need to point at Blacklight's master branch?

spec/views/curation_concerns/base/_relationships.html.erb_spec.rb
```diff
@@ -4,8 +4,8 @@
   let(:ability) { double }
   let(:solr_doc) { double(id: '123', human_readable_type: 'Work') }
   let(:presenter) { Sufia::WorkShowPresenter.new(solr_doc, ability) }
-  let(:generic_work) { GenericWork.new(id: '456', title: ['Containing work', 'barbaz']) }
```
mjgiarlo: Why'd you remove the second titles? Not used?

spec/views/users/show.html.erb_spec.rb
```diff
@@ -29,7 +29,11 @@
 
   context "with trophy" do
     let(:trophy_presenter) { Sufia::TrophyPresenter.new(solr_document) }
-    let(:solr_document) { SolrDocument.new(id: 'abc123', has_model_ssim: 'GenericWork') }
+    let(:solr_document) do
+      SolrDocument.new(id: 'abc123',
+                       has_model_ssim: 'GenericWork',
+                       thumbnail_path_ss: '/foo/bar.png')
```
mjgiarlo: Is this now required?

sufia.gemspec
```diff
@@ -18,7 +18,9 @@ Gem::Specification.new do |spec|
   spec.version       = Sufia::VERSION
   spec.license       = 'Apache2'
 
-  spec.add_dependency 'curation_concerns', '~> 1.3', '>= 1.3.3'
+  spec.add_dependency 'curation_concerns', '~> 1.4'
+  # This is not required, but helps bundler resolve a bundle faster:
```
mjgiarlo: Good note. Thanks!

app/views/users/index.html.erb
jcoyne: Rails 5 gives an exception if you call `image_tag(nil)`, so we check for nil.

lib/generators/sufia/templates/catalog_controller.rb
jcoyne: `default_html_head` doesn't exist, and Rails 5 gives you an error if you try to skip something that doesn't exist.

spec/spec_helper.rb
jcoyne: Yep.

spec/test_app_templates/Gemfile.extra
jcoyne: For this PR: https://github.com/projectblacklight/blacklight/pull/1457

spec/views/curation_concerns/base/_relationships.html.erb_spec.rb
jcoyne: This PR also offers RDF2 support, which means that statistically, the tests that check for `'Containing work'` fail 50% of the time.  As the two titles didn't seem particularly important to the test, the fix was to remove one of them.

spec/views/users/show.html.erb_spec.rb
jcoyne: Yeah, otherwise the view will attempt a `image_tag(nil)`, which is an error in Rails 5

spec/views/curation_concerns/base/_relationships.html.erb_spec.rb
mjgiarlo: OK, I documented this change in #2491.

PR #2488, opened 2016-08-17
Remove an irrelevant test

jcoyne:
This test was testing a view that is located in Blacklight and is tested
in Blacklight. It is not necessary to test it in Sufia too.


mjgiarlo:
@cam156 @awead y'all have thoughts on this one? I see Penn State added a change to this file:

https://github.com/projecthydra/sufia/commit/071f7159489b9dabdfb3cc291128a7a03a48a2a7

Are we worried that wiping out this test will introduce the possibility that we'll diverge from Blacklight, or some such?

mjgiarlo:
This PR is blocking another PR which is blocking two other, rather trivial PRs, so I'm inclined to merge this now. If there are issues with removing this test, let's recreate it after the fact and unblock all the things.

PR #2489, opened 2016-08-17
Reduce duplication in codebase

mjgiarlo:


app/controllers/sufia/resource_sync_controller.rb
```diff
@@ -1,26 +1,14 @@
...
-    body = Rails.cache.fetch("source_description_#{request.host}", expires_in: 1.week) do
-      build_source_description
-    end
-    render body: body, content_type: 'application/xml'
+    render_from_cache_as_xml(build_source_description)
   end
 
   def capability_list
-    # Caching based on host, for multitenancy support
-    body = Rails.cache.fetch("source_description_#{request.host}", expires_in: 1.week) do
```
jcoyne: These should have had different cache keys.  This is a bug.

app/controllers/sufia/resource_sync_controller.rb
```diff
@@ -1,26 +1,14 @@
...
-      build_capability_list
-    end
-    render body: body, content_type: 'application/xml'
+    render_from_cache_as_xml(build_capability_list)
   end
 
   def resource_list
-    # Caching based on host, for multitenancy support
-    body = Rails.cache.fetch("source_description_#{request.host}", expires_in: 1.week) do
```
jcoyne: Here too.

app/controllers/sufia/resource_sync_controller.rb
```diff
@@ -1,26 +1,14 @@
 class Sufia::ResourceSyncController < ApplicationController
   def source_description
-    # Caching based on host, for multitenancy support
-    body = Rails.cache.fetch("source_description_#{request.host}", expires_in: 1.week) do
-      build_source_description
-    end
-    render body: body, content_type: 'application/xml'
+    render_from_cache_as_xml(build_source_description)
```
jcoyne: This is obviating the cache.  It's running the build_source_description even if it's available in the cache.

app/controllers/sufia/resource_sync_controller.rb
```diff
@@ -1,26 +1,14 @@
...
   end
 
   def capability_list
-    # Caching based on host, for multitenancy support
-    body = Rails.cache.fetch("source_description_#{request.host}", expires_in: 1.week) do
-      build_capability_list
-    end
-    render body: body, content_type: 'application/xml'
+    render_from_cache_as_xml(build_capability_list)
```
jcoyne: It's not being cached if you call it here.

app/controllers/sufia/resource_sync_controller.rb
```diff
@@ -1,26 +1,14 @@
...
   end
 
   def resource_list
-    # Caching based on host, for multitenancy support
-    body = Rails.cache.fetch("source_description_#{request.host}", expires_in: 1.week) do
-      build_resource_list
-    end
-    render body: body, content_type: 'application/xml'
+    render_from_cache_as_xml(build_resource_list)
```
jcoyne: and not cached here.

app/controllers/sufia/resource_sync_controller.rb
mjgiarlo: Done below.

app/controllers/sufia/resource_sync_controller.rb
mjgiarlo: Done.

PR #2504, opened 2016-08-18
Assign a work to an admin set. Fixes #2503

### LINKED ISSUE #2502: An admin set should have a show page.
This should look just like the collection page
- [ ] metadata: title, description
- [ ] selected thumbnail
- [ ] first 10 members, with a way to paginate and search through the members.

### LINKED ISSUE #2503: A user should be able to assign their work to an admin set.

jcoyne:


cam156:
@jcoyne are admin sets being hard coded into the system?  Should admin sets be a configurable option?  I believe PSU would prefer for admins sets to be something we could turn on or off.

jcoyne:
@cam156 I added #2507 If we mark that as a release blocker for 7.2 is that sufficient?

cam156:
Seems reasonable.

mjgiarlo:
One question about this PR while it's still unmerged. I'm seeing the relationship between the work and the AS expressed on the work show view under Descriptions rather than under Relationships. Is that deliberate?

jcoyne:
@mjgiarlo no not deliberate.  I don't have any designs to work from, so I'm just making stuff up as I go.

mjgiarlo:
@jcoyne OK, just thought I'd ask.

PR #2511, opened 2016-08-19
Refer to files by their titles instead of their labels

hackmastera:


app/presenters/sufia/file_set_presenter.rb
```diff
@@ -6,6 +6,14 @@ class FileSetPresenter < ::CurationConcerns::FileSetPresenter
     delegate :depositor, :keyword, :date_created, :date_modified, :itemtype,
              to: :solr_document
 
+    def page_title
```
mjgiarlo: Can you add test coverage for your changes to the `FileSetPresenter`?

app/presenters/sufia/file_set_presenter.rb
```diff
@@ -6,6 +6,14 @@ class FileSetPresenter < ::CurationConcerns::FileSetPresenter
...
+      title.first
+    end
+
+    def link_name
+      current_ability.can?(:read, id) ? page_title : 'File'
+    end
+
```
awead: Could you write a view test that shows what this renders when `current_ability?` returns true or false?

mjgiarlo: I'm on the fence about this, but I could get down with adding coverage to a feature spec. I see view specs as unit tests for views, more or less, and feature specs as integration specs.

hackmastera: add to https://github.com/projecthydra/sufia/blob/master/spec/features/work_show_spec.rb ?

awead: Yeah, I'm in agreement with @mjgiarlo. Mostly I go with unit tests because they're just faster. Although there's already a feature test with the necessary setup, I think a unit test is better because it'll show exactly where this is occurring, which I'm still not clear on presently. You could do something like:

https://github.com/projecthydra/sufia/blob/master/spec/views/curation_concerns/base/_items.html.erb_spec.rb

But in the partial where the file's title is being rendered.

hackmastera: It doesn't sound like y'all agree. I'll do it wherever; just let me know.

awead: I think we're both saying "write a unit test." At least that's what I'm saying... I don't care what @mjgiarlo thinks. 😉

mjgiarlo: Haha.

I'm saying, "this def needs some unit tests in a presenter spec, *and* I think @awead is asking for integration spec coverage too, in which case I'm :+1: on adding a line to an existing feature spec and :-1: on using view specs for integration testing" though, as always, I am happy to hear counterargs!

mjgiarlo: I'd merge this PR w/o integration testing, though, because, ugh, slow feature specs.

awead: OK, I see where the confusion arose. I was saying use a view test that's "integration-ish" because feature tests are slow. Looks like that's already been done, so I'm happy. :) But, to @mjgiarlo's point, since the feature test is already there, adding an additional like to check for the link is probably the better option. However, it's fine the way it is!

hackmastera:
@mjgiarlo @awead here's another test.

mjgiarlo:
Happy to merge this once Travis is :green_apple:

hackmastera:
no idea why i am getting this fail.

mjgiarlo:
@hackmastera Ugh, we see that every now and then and I don't think it has anything to do with your PR. I'll restart that one build.

hackmastera:
thank you!

PR #2521, opened 2016-08-23
Remove 'My Highlights' filter from search; fixes #2470

### LINKED ISSUE #2470: Remove My Highlights Listing from search
### Descriptive summary

In sufia 7 (actually sufia 6 too) we can search on highlights, but since there is only 4 this seems like overkill.  Check #2413

### Expected behavior

Remove My Highlights from the search bar.

### Actual behavior

My Highlights is in the drop down list for search.
screen shot 2016-08-16 at 7 44 50 am

cbeer:


app/views/catalog/_search_form.html.erb
```diff
@@ -30,10 +30,6 @@
                 data: { "search-option" => sufia.dashboard_collections_path, "search-label" => t("sufia.search.form.option.my_collections.label_short") } %>
           </li>
           <li>
-            <%= link_to t("sufia.search.form.option.my_highlights.label_long"), "#",
-                data: { "search-option" => sufia.dashboard_highlights_path, "search-label" => t("sufia.search.form.option.my_highlights.label_short") } %>
```
jcoyne: Should we remove that i18n key from the locale file?

cbeer: 👌

mjgiarlo:
:clap: 

Happy to merge when Travis is :green_heart:

PR #2522, opened 2016-08-23
Remove catch-all error routing from engine routes; fixes #1906

### LINKED ISSUE #1906: catch-all routing for 404 should be appended in generator, not in the engine

cbeer:


jcoyne:
Does this need to be sufia 8 or can it be 7.1?

mjgiarlo:
I hope we can make it 7.2 if semver can allow it.

cbeer:
I think 7.x is fine; this is just swapping one error implementation (sufia's) with another (rails') with essentially the same end-user behavior.

mjgiarlo:
I'll buy that for a dollar.

jcoyne:
Curiously the Rails 4.2 build completely fell apart.

jcoyne:
Additionally remove `https://github.com/projecthydra/sufia/blob/master/app/views/errors/404.html.erb`

jcoyne:
And update the generator: https://github.com/projecthydra/sufia/blob/master/lib/generators/sufia/install_generator.rb#L108-L109

PR #2529, opened 2016-08-23
Update rspec tests and configurations for new defaults and practices

cbeer:


spec/spec_helper.rb
```diff
@@ -177,4 +177,21 @@ def main_app
...
   config.infer_spec_type_from_file_location!
+
+  config.expect_with :rspec do |expectations|
+    expectations.include_chain_clauses_in_custom_matcher_descriptions = true
+  end
+
+  config.default_formatter = 'doc' if config.files_to_run.one?
+
+  config.order = :random
+  Kernel.srand config.seed
```
awead: This'll be interesting... I don't think we were doing random before, but it's a good idea.

spec/spec_helper.rb
```diff
@@ -177,4 +177,21 @@ def main_app
...
+
+  config.shared_context_metadata_behavior = :apply_to_host_groups
+
+  config.filter_run_when_matching :focus
+
+  config.example_status_persistence_file_path = 'spec/examples.txt'
```
awead: how about about in tmp and leave .gitignore unchanged?

cbeer: We could, although this is the default path from upstream, so I assume this is where people would expect to find the file.

PR #2534, opened 2016-08-24
Disabling the save button after the form submit 

### LINKED ISSUE #2397: Multiple works created for one upload
### Descriptive summary

I can not repeat this.  So I'm documenting it incase someone else sees this happen.

I uploaded one work into Sufia 7/ScholarSphere and got two works when I hit save.

This is my local copy and the works were created within milliseconds of each other so I could not have created them via the UI and no one else has access to the system.

https://gist.github.com/cam156/6636f6d2ac8f7b90520126d2054435ee

cam156:
Fixes #2397
Disabling the save button after the form submit so the user can not create multiple works

If the save button is not disabled the user can double click, triple click, even click many more times and create multiple works from a single form

### DIFF
--- a/app/assets/javascripts/sufia/save_work/save_work_control.es6
+++ b/app/assets/javascripts/sufia/save_work/save_work_control.es6
@@ -43,6 +43,17 @@ export class SaveWorkControl {
   }
 
   /**
+   * Keep the form from being submitted many times.
+   *
+   */
+  preventSubmitIfAlreadyInProgress() {
+    this.form.on('submit', (evt) => {
+      if (this.isValid())
+        this.saveButton.prop("disabled", true);
+    })
+  }
+
+  /**
    * Is the form for a new object (vs edit an exisiting object)
    */
   get isNew() {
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/sufia/save_work/save_work_control.es6
```diff
@@ -43,6 +43,17 @@ export class SaveWorkControl {
   }
 
   /**
+   * Keep the form from being submitted many times.
+   *
+   */
+  preventSubmitIfAlreadyInProgress() {
+    this.form.on('submit', (evt) => {
+      if (this.isValid())
```
mjgiarlo: Is the validity check here just to prevent marking as disabled an already disabled control?

spec/javascripts/save_work_spec.js
```diff
@@ -180,6 +180,14 @@ describe("SaveWorkControl", function() {
         $('#new_generic_work').submit();
         expect(spyEvent).not.toHaveBeenPrevented();
       });
+
+      it("disables save after submission", function() {
+        spyOn(target, 'isValid').and.returnValue(true);      expect(target.saveButton).toBeDisabled();
```
mjgiarlo: Looks like you need a newline after the first semi-colon here.

app/assets/javascripts/sufia/save_work/save_work_control.es6
cam156: @mjgiarlo yes.  The code above this checks validity on form submit even though you should never be able to submit if the form is invalid.  I was following coding style...

PR #2540, opened 2016-08-24
Fix or ignore new rubocop-rspec style violations

cbeer:


backus:
Many of these cops you are disabling support `--auto-correct` or `--auto-gen-config`. Doing `--auto-gen-config` for things like `MaxExpectations` or `NestedGroups` will determine what your current highest offenses are (for example 6 expectations in one example or 3 levels of nesting) and make that your maximum. This should be better for you than disabling the cops entirely since then you can at least avoid even more egregious violations

cbeer:
Thanks @backus; I share your concern, but I don't think this community has agreed what, if any, limits we want to target.

backus:
@cbeer ok. I just released these updates and this was the first public PR that seemed to be implementing the changes so I figured I would give you a heads up. I hope the other cops were helpful :)

PR #2544, opened 2016-08-24
Rescue non-critical solr connection exceptions

cbeer:


jcoyne:
@mjgiarlo the duplication error from codeclimate seems silly.  I think the way the code is currently written is more readable than any refactoring that removes duplicated code.

mjgiarlo:
@jcoyne Feel free to add an exclusion:

``` yaml
engines:
  # ...
  duplication:
    enabled: true
    # ...
    exclude_fingerprints:
    - 3fe527befeb3a1f8ac773f6393f75e0c
   # ...
```

PR #2548, opened 2016-08-25
Switch to using an isolated namespace

jcoyne:


jcoyne:
This will break any application that has overridden views (they're now have `sufia/` in their path), so we may want to consider this for Sufia 8

mjgiarlo:
@jcoyne Agreed. I'm not ready to push Sufia 8 yet, so let's keep the branch around for when we're ready to start working on 8.0 (after we have a tested upgrade/migration path from 6 to 7 and have finished some feature work). Thanks!

cbeer:
If it's just overriding views, we ought to be able to add the current location to the lookup path.

jcoyne:
Did this in Hyrax

PR #2551, opened 2016-08-25
Add flip for dynamically configuring features

jcoyne:
This will allow us to turn off admin sets for example.

### DIFF
--- a/app/controllers/sufia/admin/strategies_controller.rb
+++ b/app/controllers/sufia/admin/strategies_controller.rb
@@ -1,0 +1,9 @@
+module Sufia
+  module Admin
+    class StrategiesController < Flip::StrategiesController
+      before_action do
+        authorize! :manage, Feature
+      end
+    end
+  end
+end
--- a/app/assets/stylesheets/sufia.scss
+++ b/app/assets/stylesheets/sufia.scss
@@ -1,6 +1,7 @@
 /* ...
  *= require jquery-ui
  *= require select2
+ *= require flip
  *= require_self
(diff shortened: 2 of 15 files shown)


app/controllers/sufia/admin/strategies_controller.rb
```diff
@@ -0,0 +1,9 @@
+module Sufia
+  module Admin
+    class StrategiesController < Flip::StrategiesController
+      before_action do
+        authorize! :manage, Feature
```
jeremyf: Why not `Sufia::Feature` as above?

jcoyne: It is Sufia::Feature since it's being referred to from within the Sufia module.

mjgiarlo:
@jcoyne Did you want your isolate_namespace PR merged in advance, or is this ready?

jcoyne:
@mjgiarlo this is ready. the isolate_namespace will take a bit more work and can come later.

app/controllers/sufia/admin/strategies_controller.rb
mjgiarlo: But aren't both of the new controllers in the `Sufia::Admin` namespace? I don't know why you'd use the explicit `Sufia::` namespace in one and not the other.

jcoyne: I just wasn't looking at the one when doing the other. The fully-qualified path is optional. Semantically they are the same.

PR #2557, opened 2016-08-25
showing works instead of files in user index view

### LINKED ISSUE #2453: Users page is still showing files
### Descriptive summary

IIn Sufia 7 the Users listing is showing the number of files deposited instead of works.

### Expected behavior

I would expect to see deposited works, or work and collections not files

### Actual behavior

Profile page showing the correct number of works
screen shot 2016-08-10 at 2 24 14 pm

User listing showing the number of files
screen shot 2016-08-10 at 2 24 00 pm

When I have works that include multiple files
And I go to my user profile
And I click on the All Users button at the bottom
Then I should see the number of works, not the number of files

jenlindner:
fixes #2453.

Shows User's works instead of uploaded files.

Uses number_of_works helper method instead of number_of_deposits.

### DIFF
--- a/app/views/users/index.html.erb
+++ b/app/views/users/index.html.erb
@@ -2,19 +2,19 @@
   <%= render partial: 'search_form' %>
   <h1><%= application_name %> Users</h1>
 </div>
-<table class="table table-striped"> 
-    <thead> 
-        <tr> 
-            <th>Avatar</th> 
+<table class="table table-striped">
+    <thead>
+        <tr>
+            <th>Avatar</th>
             <th class="sorts"><i id="name" class="<%=params[:sort].blank? ? 'caret up' : params[:sort]== "name desc" ? 'caret' : params[:sort]== "name" ? 'caret up' : ''%>"></i> User Name</th>
             <th class="sorts"><i id="login" class="<%=params[:sort]== "login desc" ? 'caret' : params[:sort]== "login" ? 'caret up' : ''%>"></i> User Id</th>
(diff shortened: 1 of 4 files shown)


app/views/users/index.html.erb
```diff
@@ -2,31 +2,31 @@
...
               </td>
                <td><%= link_to user.name, sufia.profile_path(user) %></td>
                <td><%= link_to user.user_key, sufia.profile_path(user) %></td>
                <td><%= user.department %></td>
-               <td><%= number_of_deposits(user) %></td>
-            </tr> 
-         <% end %> 
-    </tbody> 
-</table> 
+               <td><%= number_of_works(user) %></td>
```
jcoyne: This was the only use of `number_of_deposits`, can we remove that method now?

jcoyne:
@mjgiarlo something is wrong with dependency ci on this PR. Can you take a look?

mjgiarlo:
@jcoyne sure. but it looks like you have an unaddressed comment, so do we want to wait on a response to that before we merge?

app/views/users/index.html.erb
jenlindner: I'd say yes. It's possible Admin sets might have a use for it but if so it would be easy to re-instate.

jenlindner: I removed the method and its test.

spec/helpers/sufia_helper_spec.rb
```diff
@@ -191,10 +191,6 @@ def new_state
       create_models("Collection", user1, user2)
     end
 
-    it "finds only 3 files" do
```
jcoyne: Remove the whole block starting at line 186

app/views/users/index.html.erb
jenlindner: Removed the whole block.

PR #2558, opened 2016-08-26
Add an admin dashboard page

jcoyne:

Presently this is just a skeleton, but we'll eventually put more admin
reports and settings here.

The design looks like this:
screen shot 2016-08-25 at 4 41 07 pm

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -24,7 +24,7 @@ engines:
       extensions:
         - .es6
   fixme:
-    enabled: true
+    enabled: false
   rubocop:
     enabled: true
   coffeelint:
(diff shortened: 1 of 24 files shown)


mjgiarlo:
Rebase? Also, can you include a screenshot of what your changes look like? (It's not the screenshot above, right?)

jcoyne:
My version of the dashboard looks like this:
screen shot 2016-08-26 at 2 23 39 pm

jcoyne:
@mjgiarlo we can't have a comment that says `TODO`? that seems screwy.

mjgiarlo:
@jcoyne We can disable that rule entirely or add an ignore for that line. Here's how to disable it entirely in `.codeclimate.yml`:

```yaml
engines:
  # ... 
  fixme:
    # ... 
    enabled: false
```

PR #2559, opened 2016-08-26
Fix the fedora warning about path containing an empty element

jcoyne:
before merging, check the travis output, we don't want to see any instances of `InvalidResourceIdentifierExceptionMapper`


mjgiarlo:
@jcoyne there still appear to be 16 such instances of that warning, all about `/test/12/3//li/st/123/list_source`

PR #2567, opened 2016-08-26
Remove edit option from collection menu if not available to user.

### LINKED ISSUE #2400: Collections are showing options on Works that it should not
### Descriptive summary

When I view someone else's collection I should not get the option to edit or highlight the work.

### Expected behavior

No menu items available?  Or other ones, just not edit and highlight

### Actual behavior

Options on the menu
screen shot 2016-08-05 at 2 47 02 pm

Error screen I see when I try to edit the work
screen shot 2016-08-05 at 2 47 26 pm

andjsmit:
Fixes #2400 

Removes edit option when unavailable.

This should remove the edit option when a user clicks on the "Select an Action" for an individual work being viewed from a collection page if that user does not have edit rights to the work.

### DIFF
--- a/app/views/collections/_show_document_list_menu.html.erb
+++ b/app/views/collections/_show_document_list_menu.html.erb
@@ -1,16 +1,18 @@
 <div class="btn-group">
-  <button class="btn btn-default dropdown-toggle" data-toggle="dropdown">Select an action <span class="caret"></span>
+  <button class="btn btn-default dropdown-toggle" data-toggle="dropdown">Select an action<span class="caret"></span>
   </button>
   <ul class="dropdown-menu dropdown-menu-right">
-    <li>
-      <%= link_to [:edit, document],
-                  class: "itemicon itemedit",
-                  title: t('sufia.collection.document_list.edit'),
-                  id: "edit_work_link_#{document.id}" do %>
-        <i class='glyphicon glyphicon-pencil'></i> <%= t('sufia.collection.document_list.edit') %>
-      <% end %>
-    </li>
     <% if current_user %>
+      <% if can? :edit, document %>
+      <li>
+        <%= link_to [:edit, document],
+                    class: "itemicon itemedit",
+                    title: t('sufia.collection.document_list.edit'),
+                    id: "edit_work_link_#{document.id}" do %>
+          <i class='glyphicon glyphicon-pencil'></i> <%= t('sufia.collection.document_list.edit') %>
+        <% end %>
+      </li>
+      <% end %>
       <li>
         <%= display_trophy_link(current_user, document.id) do |text| %>
           <i class='glyphicon glyphicon-star'></i> <%= text %>
--- a/spec/views/collections/_show_document_list_menu.html.erb_spec.rb
+++ b/spec/views/collections/_show_document_list_menu.html.erb_spec.rb
@@ -1,18 +1,38 @@
 
 describe 'collections/_show_document_list_menu.html.erb', type: :view do
-  context 'when user logged in displaying the collections of current user' do
+  context 'when user is viewing a collection' do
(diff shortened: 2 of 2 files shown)


app/views/collections/_show_document_list_menu.html.erb
```diff
@@ -1,7 +1,8 @@
 <div class="btn-group">
-  <button class="btn btn-default dropdown-toggle" data-toggle="dropdown">Select an action <span class="caret"></span>
+  <button class="btn btn-default dropdown-toggle" data-toggle="dropdown">Select an action<span class="caret"></span>
   </button>
   <ul class="dropdown-menu dropdown-menu-right">
+    <% if current_user && document.edit_people.include?(current_user.email) %>
```
jcoyne: The way we check for permissions is by using the ability class.  So `can? :edit, document`. This way it will check the edit groups as well as the edit users.

spec/views/collections/_show_document_list_menu.html.erb_spec.rb
```diff
@@ -1,18 +1,35 @@
 
 describe 'collections/_show_document_list_menu.html.erb', type: :view do
-  context 'when user logged in displaying the collections of current user' do
+  context 'when user is viewing a collection' do
```
jcoyne: They are viewing a list of collections, not a single collection here, right?

spec/views/collections/_show_document_list_menu.html.erb_spec.rb
```diff
@@ -1,18 +1,35 @@
...
     let(:user) { create :user }
-    let(:document) { SolrDocument.new(id: '1234') }
+    let(:another_user) { create :user }
+    let(:document) { SolrDocument.new(id: '1234', edit_access_person_ssim: [user.email]) }
```
jcoyne: Instead of stubbing access on the document, try this: `allow(view).to receive(:can?).with(:edit, document).and_return(true/false)`

app/views/collections/_show_document_list_menu.html.erb
andjsmit: Done. Thanks.

spec/views/collections/_show_document_list_menu.html.erb_spec.rb
andjsmit: I think this drop down menu is generated in the action column of a row for an individual work on a single collection index page unless it is also used in additional contexts that I haven't seen.

andjsmit:
Updated test using allow(user).to receive versus stubbing in access to document.

app/views/collections/_show_document_list_menu.html.erb
```diff
@@ -1,7 +1,8 @@
 <div class="btn-group">
-  <button class="btn btn-default dropdown-toggle" data-toggle="dropdown">Select an action <span class="caret"></span>
+  <button class="btn btn-default dropdown-toggle" data-toggle="dropdown">Select an action<span class="caret"></span>
   </button>
   <ul class="dropdown-menu dropdown-menu-right">
+    <% if current_user && current_user.can?(:edit, document) %>
```
jcoyne: no need for the `current_user` just `can? ...`

jcoyne:
Once you've made that change can you squash all your commits into a single commit (and then force push to this branch)?

app/views/collections/_show_document_list_menu.html.erb
andjsmit: Didn't think that was possible as test failed without current_user. Looking closer, I found the issue was in the test not the execution. Added `let(:ability) { instance_double("Ability") }` and `allow(controller).to receive(:current_ability).and_return(ability)` to fix.

PR #2577, opened 2016-08-29
Re-add tab icons for creating works

### LINKED ISSUE #2449: Icons on FileSet Edit tabs
### Descriptive summary

In Sufia 7 on the edit screen of the FileSet there are icons on the tabs.  Since these were removed from the work screen we should remove them from the FileSet Screen too.

Other alignment could occur too, but just removing the icons would visually align the two pages as a quick fix.

We may also want to change the permissions tab to sharing.

### Expected behavior

screen shot 2016-08-10 at 12 06 37 pm

### Actual behavior

screen shot 2016-08-10 at 12 06 50 pm

cbeer:
In #2449, we said to remove the tab icons from the `FileSet` form, but it makes more sense to me to re-add icons to the work form. Was there was a compelling reason we removed them in the first place? 

screen shot 2016-08-29 at 10 48 47 am

Fixes #2449.

### DIFF
--- a/app/assets/stylesheets/sufia/_collections.scss
+++ b/app/assets/stylesheets/sufia/_collections.scss
@@ -82,3 +82,19 @@ form.per_page {
 .presenter-title {
   padding-left: 20px;
 }
+
+.icon-metadata {
+  @extend .fa-tags;
+}
+
+.icon-files {
+  @extend .fa-files-o;
+}
+
+.icon-relationships {
+  @extend .fa-link;
+}
+
+.icon-share {
+  @extend .fa-key;
+}
--- a/app/views/curation_concerns/base/_guts4form.html.erb
+++ b/app/views/curation_concerns/base/_guts4form.html.erb
@@ -12,7 +12,7 @@
           <li role="presentation">
         <% end %>
             <a href="#<%= tab %>" aria-controls="<%= tab %>" role="tab" data-toggle="tab">
-              <%= t("sufia.works.edit.tab.#{tab}") %>
+              <i class="fa icon-<%= tab %>"></i> <%= t("sufia.works.edit.tab.#{tab}") %>
             </a>
           </li>
       <% end %>


mjgiarlo:
Do y'all have thoughts on why we removed the icons? @projecthydra/sufia-ui-ux-advisors

mtribone:
In the past, I was pretty hell bent on nuking IconSphere. Unless it had to do with an action like upload, I _personally_ wanted to remove "unnecessary" icons.

I think in this example, it's simply because the wireframes didn't include icons because they were focused on the interactions. And we didn't make it as far as the filesets in the UI group, so that explains the icons being there.

justcolin:
I am unaware of any accessibility problems with having icons there as long as they are added in an accessible manor. That being said, they are pretty tiny, meaning that it would be hard for a number of people to figure out what they are.

From a UX standpoint I think they might help. Do you have any thoughts @heidisb or @bnhowell?

heidisb:
I have not done tons of work with icons, but my personal feelings are that they should be (1) easily recognizable (share, download, etc.) (2) be immediately meaningful **and/or** (3) tie concepts together in some way. To me...the icons on those tabs don't mean a whole lot unless they are used consistently with the accompanying labels throughout the interface.

bnhowell:
Usability tests last week (without icons) showed that users easily understood the Descriptions and Files tabs but were confused by/unsure about Relationships and Share tabs. Those tabs have multiple meanings that are difficult to disambiguate. +1 for @heidisb comments about using icons with immediate meaning / widespread recognition.

mjgiarlo:
@justcolin @heidisb @bnhowell Thanks much! Would y'all mind recommending alternative icons that might work better for Relationships and Share?

cbeer:
Does it make sense to ship this, and have a separate ticket to improve iconography globally?

mjgiarlo:
Works for me, @cbeer. Thanks!

Let's keep the discussion going here until said "separate ticket" is formed enough to be written.

hackmastera:
-1 to icons on tabs. i feel they add visual clutter without clear benefit.

PR #2600, opened 2016-08-31
Removing characterization override now that curation concerns inludes it

### LINKED ISSUE #2431: Characterization seems to be happening but not logged
### Descriptive summary

In Sufia 7 when I upload a text file or an mp3 (there may be others) characterization seems to run ( the mime_type is known) but the file set view always says it has not yet been run.

Characterization does get applied for an image.

### Expected behavior

Once characterization has been run the fileset view should show it.

### Actual behavior

screen shot 2016-08-09 at 2 26 18 pm

When I create a new work
And I choose a .txt or a .mp3 to attach to the work
And I apply required metadata
Then I go to the file set show page
I should see the characterization information

cam156:
Fixes #2431 

Removing overrides that are no longer needed since https://github.com/projecthydra/curation_concerns/pull/923 was merged.  That PR includes the code to actually fix the issue.


jcoyne:
Oh this is my favorite kind of commit. One that just removes code.

jeremyf:
@cam156 Gets all the positive emojis for this pull request.

PR #2608, opened 2016-09-01
AdminSetSearchBuilder#search_results is now public

jcoyne:
Since https://github.com/projecthydra/curation_concerns/pull/955


cbeer:
Should we get a release of CC, so we can pin version dependencies accordingly?

mjgiarlo:
Yes, and @awead was asking about cutting 1.6.0 yesterday.

PR #2626, opened 2016-09-02
Adds external link icon to Related URL field

### LINKED ISSUE #2472: Related url no longer displays as a link
### Descriptive summary

In Sufia 6 there was a view for this field which called iconify_auto_link. Since these views were replaced with renderers the behavior was lost.

1. create a work with a related url
1. view the work's show page

### Related work
https://github.com/projecthydra/sufia/issues/2374

andjsmit:
Fixes #2472 

Change in Curation Concerns removed external link icon from related url field.

This provides a new CC renderer using code similar to an existing Sufia helper (iconify_auto_link). A future issue may pull out the helper code and replace it with a renderer. 

### DIFF
--- a/app/views/curation_concerns/base/_attribute_rows.html.erb
+++ b/app/views/curation_concerns/base/_attribute_rows.html.erb
@@ -8,6 +8,6 @@
 <%= presenter.attribute_to_html(:keyword, render_as: :faceted) %>
 <%= presenter.attribute_to_html(:date_created, render_as: :linked, search_field: 'date_created_tesim') %>
 <%= presenter.attribute_to_html(:based_near, render_as: :faceted) %>
-<%= presenter.attribute_to_html(:related_url) %>
+<%= presenter.attribute_to_html(:related_url, render_as: :external_link) %>
 <%= presenter.attribute_to_html(:resource_type, render_as: :faceted) %>
 <%= presenter.attribute_to_html(:source) %>
--- a/spec/views/curation_concerns/base/_attribute_rows.html.erb_spec.rb
+++ b/spec/views/curation_concerns/base/_attribute_rows.html.erb_spec.rb
@@ -1,0 +1,19 @@
+require 'spec_helper'
+
+describe 'curation_concerns/base/_attribute_rows.html.erb', type: :view do
+  let(:url) { "http://example.com" }
+  let(:ability) { double }
+  let(:work) { stub_model(GenericWork, related_url: [url]) }
+  let(:solr_document) { SolrDocument.new(work.to_solr) }
+  let(:presenter) { Sufia::WorkShowPresenter.new(solr_document, ability) }
+
+  let(:page) do
+    render 'curation_concerns/base/attribute_rows', presenter: presenter
+    Capybara::Node::Simple.new(rendered)
+  end
+
+  it 'shows external link with icon for related url field' do
+    expect(page).to have_selector '.glyphicon-new-window'
+    expect(page).to have_link(url)
+  end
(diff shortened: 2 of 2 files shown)


app/presenters/sufia/presents_attributes.rb
```diff
@@ -0,0 +1,12 @@
+module Sufia
+  class PresentsAttributes < ::CurationConcerns::PresentsAttributes
```
jcoyne: Why do we need this class. I don't see it referenced anywhere.

spec/renders/curation_concerns/renderers/external_link_attribute_renderer_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+    let(:tr_content) {
+      "<tr><th>Name</th>\n" \
+       "<td><ul class='tabular'>" \
+       "<li class=\"attribute name\">"\
+       "<a href=\"http://example.com\">"\
+       "<span class='glyphicon glyphicon-new-window'></span>&nbsp;"\
+       "http://example.com</a></li>\n" \
+       "</ul></td></tr>"
+    }
+    it { expect(renderer).not_to be_microdata(field) }
```
jcoyne: Is this test necessary?  There are an infinite number of negative assertions we could make about this code, but few of them would have any value.

andjsmit: Was a cut and paste from a CC test for link attributes renderer. I'll remove.

app/presenters/sufia/presents_attributes.rb
andjsmit: Must have added while experimenting. Shux. I'll remove it.

mjgiarlo:
Looks good to me. I have just two questions:

1. Should we also add a view spec to cover the change made to the view?
2. Should we move the renderer and its spec from Sufia to CC, since Sufia doesn't currently have renderers?

Thanks, @andjsmit! Not far off, now. :)

hackmastera:
I'm not sure about moving it to curation concerns; there's currently no use of glyphicon-new-window there.

andjsmit:
This is the argument that convinced me back in issue #2472 to p;ace it in Sufia over CC.

andjsmit:
Added a views test.

mjgiarlo:
:speech_balloon: @hackmastera 
> I'm not sure about moving it to curation concerns; there's currently no use of glyphicon-new-window there.

What say ye, @tpendragon @escowles @jcoyne?

mjgiarlo:
Specs look great, y'all. Watching Travis do its thing now.

jcoyne:
@mjgiarlo I don't care.

andjsmit:
Failed when trying to load http://archive.apache.org/dist/lucene/solr/6.2.0/solr-6.2.0.zip.md5, but the file appears available now. I'm going to restart the test.

escowles:
+1 to moving this to CC, and consistently marking external links in CC.

mjgiarlo:
Thanks, @escowles!

andjsmit:
Renderer code moved to Curation Concerns.

andjsmit:
This code will not be ready until renderer code makes it through to curation concerns gem.

mjgiarlo:
CC PR merged, @andjsmit. I suspect we need to cut 1.6.0 today so this branch can pull over the latest CC version and test the changes in this branch.

mjgiarlo:
Pushed a rebased branch and re-running the tests.

PR #2636, opened 2016-09-06
Use a homepage layout with a smaller header

jcoyne:

screen shot 2016-09-06 at 3 16 50 pm

Additionally, this positions sufia to use a user uploaded header image.


mjgiarlo:
What say ye, @projecthydra/sufia-ui-ux-advisors? Do you like the direction this is heading in?

justcolin:
I think it's a good direction, though the nav links over that background image in the header need a translucent-colored background to guarantee they will always meet contrast specs.

Also, I like this new hound bot. It is picky in all the right ways. :-D

jcoyne:
@justcolin those links do have a translucent colored background `$navbar-transparent-bg` and `$navbar-transparent-link-hover-color` (also used for `:focus`)

hackmastera:
+1 this is a nice improvement

justcolin:
@jcoyne Apologies, I was just going by the screenshot which appears to only have the background against the home link.

justcolin:
Maybe just do a translucent bar all the way across the bottom part?

hackmastera:
@justcolin i see a translucent bar across the bottom in the screenshot. do you think it is too subtle?

justcolin:
@hackmastera I totally didn't notice that until I zoomed in on the screenshot, so yes, I think it is too subtle. In order to make sure accessibility specs are met we tend to set translucent backgrounds so even if the thing behind them is `#000` or `#FFF` you still get at least a 4.5:1 contrast ratio (WCAG 2.0 AA). That way people can pick whatever background image they want and it will always have good contrast and little flecks of black or white won't make it harder to read.

jcoyne:
The current version is darker than the screenshot due to adjustments in the gradient overlay.  I don't know how I would check the contrast ration though. Can you show me?

justcolin:
@jcoyne: Do you use a Mac? If so you can do this:

1. Since the text is white the worst case for contrast would be if the image in the background were all white. Replace the background image with `#FFF`.
2. Open the "Digital Color Meter" in `Applications/Utilities`. Select to `View > Display Values > as Hexadecimal`, make sure `Display Native Values` is selected from the dropdown and use that to measure the color of the translucent background over the white.
3. Use a [color contrast tool](http://webaim.org/resources/contrastchecker/) to measure the difference between that value and `#FFF`

jcoyne:
Okay, with that method, it looks okay, except for the active tab, which has too light of a background.

jcoyne:
Fixed the active tab, so we should be > 4.5:1 for this image. Your mileage may vary if you change the background image or any of the colors.

cbeer:
We could also leave it to the discretion of each install to choose an appropriate background image and/or override some scss variables to achieve an appropriate contrast.

mjgiarlo:
@jcoyne Could you update the screenshot above if you haven't already so we can see the changes you've made?

I'm inclined to say this is a good improvement on what we already have and get it merged today. Any last comments, @projecthydra/sufia-ui-ux-advisors? Thanks!

simosacchi:
I very much like to idea of a smaller header. Also agree that the menu bar should be more visible in itself. Without it, the menu entries look like tabs which I am sure is not the expected outcome.

+1 great job

justcolin:
My vote is to have the default colors ensure that any image will be
accessible in terms of contrast so it is accessible by default, then if
people want to change the colors they can on their own.

mjgiarlo:
@justcolin I'm not sure we're quite there yet, re: allowing anyone to upload any image. That sounds like a good follow-up PR. Thanks!

PR #2641, opened 2016-09-07
Admin set show page for admins

### LINKED ISSUE #2502: An admin set should have a show page.

jcoyne:

Fixes #2502

### DIFF
--- a/app/controllers/sufia/admin/admin_sets_controller.rb
+++ b/app/controllers/sufia/admin/admin_sets_controller.rb
@@ -1,7 +1,42 @@
 module Sufia
   class Admin::AdminSetsController < ApplicationController
+    include CurationConcerns::CollectionsControllerBehavior
     load_and_authorize_resource
     layout 'admin'
+    self.presenter_class = Sufia::AdminSetPresenter
+
+    def show
+      add_breadcrumb t(:'sufia.controls.home'), root_path
+      add_breadcrumb t(:'sufia.toolbar.admin.menu'), sufia.admin_path
+      add_breadcrumb t(:'sufia.admin.sidebar.admin_sets'), sufia.admin_admin_sets_path
+      add_breadcrumb 'View Set', request.path
+      super
+    end
+
+    # Override the default prefixes so that we use the collection partals.
+    def self.local_prefixes
+      ["sufia/admin/admin_sets", "collections", 'catalog']
+    end
+
+    # Overriding the way that the search builder is initialized
+    def collections_search_builder
+      collections_search_builder_class.new(self, :read)
+    end
+
+    # Used for the show action
+    def collection_search_builder_class
+      Sufia::SingleAdminSetSearchBuilder
+    end
+
+    # Used for the index action
+    def collections_search_builder_class
+      CurationConcerns::AdminSetSearchBuilder
+    end
+
+    # Used to get the members for the show action
+    def collection_member_search_builder_class
+      Sufia::AdminSetMemberSearchBuilder
+    end
 
     def index
       authorize! :manage, AdminSet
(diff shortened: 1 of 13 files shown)


app/controllers/sufia/admin/admin_sets_controller.rb
```diff
@@ -1,7 +1,42 @@
...
+    def show
+      add_breadcrumb t(:'sufia.controls.home'), root_path
+      add_breadcrumb t(:'sufia.toolbar.admin.menu'), sufia.admin_path
+      add_breadcrumb t(:'sufia.admin.sidebar.admin_sets'), sufia.admin_admin_sets_path
+      add_breadcrumb 'View Set', request.path
+      super
+    end
+
+    # Override the default prefixes so that we use the collection partals.
+    def _prefixes
```
jeremyf: Is this a public API of Rails? I'm wondering about `prepend_view_path` instead? (Not certain on the specfics). It also seams that the `@_prefixes ||=` is a possible point of confusion. Perhaps the following so we don't have incidental state related issues:

```ruby

DEFAULT_PREFIXES_FOR_COLLECTION_PARTIALS = ["sufia/admin/admin_sets", "collections", 'catalog'].freeze
def _prefixes
  DEFAULT_PREFIXES_FOR_COLLECTION_PARTIALS
end
```

app/controllers/sufia/admin/admin_sets_controller.rb
```diff
@@ -33,10 +68,10 @@ def repository
...
 
       def setup_create_form
-        add_breadcrumb  'Home', root_path
-        add_breadcrumb  'Repository Dashboard', sufia.admin_path
-        add_breadcrumb  'Administrative Sets', sufia.admin_admin_sets_path
-        add_breadcrumb  'New', sufia.new_admin_admin_set_path
+        add_breadcrumb t(:'sufia.controls.home'), root_path
+        add_breadcrumb t(:'sufia.toolbar.admin.menu'), sufia.admin_path
+        add_breadcrumb t(:'sufia.admin.sidebar.admin_sets'), sufia.admin_admin_sets_path
+        add_breadcrumb 'New', sufia.new_admin_admin_set_path
```
jeremyf: Should 'New' also be translated?

app/views/sufia/admin/admin_sets/_sort_and_per_page.html.erb
```diff
@@ -0,0 +1,3 @@
+<div id="sortAndPerPage" class="clearfix">
```
jeremyf: Do we want this extract diff if we don't show_pagination?
```erb
<% if show_pagination? %>
<div id="sortAndPerPage" class="clearfix">
  <%= render partial: "paginate_compact", object: @response %>
</div>
<% end %>
```

app/views/sufia/admin/admin_sets/show.html.erb
```diff
@@ -0,0 +1,38 @@
+<% content_for :page_header do %>
+  <h1><span class="fa fa-sitemap"></span> Administrative Set</h1>
```
jeremyf: Given that you are internationalizing "Administrative Sets" , should this also be internationalized?

app/controllers/sufia/admin/admin_sets_controller.rb
mjgiarlo: This is not a public API. We discussed this on Slack the other week and @cbeer :speech_balloon: 
> we override `#_prefix` because `.local_prefixes` will prepend the path, and we really want to append it.

mjgiarlo: I'm cool with modifying how we override `#_prefixes` in our controllers, but maybe that should be a separate ticket, which we can apply to all the places in CC and Sufia where we do this, rather than making that change in this PR.

jcoyne: `prepend_view_path` does not have the same effect as `_prefixes`. it's called here: https://github.com/rails/rails/blob/810dff7c9fa9b2a38eb1560ce0378d760529ee6b/actionpack/lib/action_controller/metal/implicit_render.rb#L34

Perhaps we could override `local_prefixes` instead:
https://github.com/rails/rails/blob/3fc0bbf008f0e935ab56559f119c9ea8250bfddd/actionview/lib/action_view/view_paths.rb#L27

PR #2644, opened 2016-09-07
Create AdminSet with public read and admin edit access

### LINKED ISSUE #2582: When an admin set is created it should have public read and admin group edit access.
This should be stored as a WebACL and indexed into Solr

jcoyne:
Fixes #2582

### DIFF
--- a/spec/controllers/sufia/admin/admin_sets_controller_spec.rb
+++ b/spec/controllers/sufia/admin/admin_sets_controller_spec.rb
@@ -43,25 +43,27 @@ describe Sufia::Admin::AdminSetsController do
     end
 
     describe "#create" do
+      let(:service) { instance_double(Sufia::AdminSetService) }
+      before do
+        allow(Sufia::AdminSetService).to receive(:new)
+          .with(AdminSet, user)
+          .and_return(service)
+      end
+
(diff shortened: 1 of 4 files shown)


spec/controllers/sufia/admin/admin_sets_controller_spec.rb
```diff
@@ -43,25 +43,27 @@
     end
 
     describe "#create" do
+      let(:service) { instance_double(Sufia::AdminSetService) }
+      before do
+        allow(Sufia::AdminSetService).to receive(:new)
```
jeremyf: If you create `Sufia::AdminSetService.create` method you won't require this before preamble and can instead add the expectation directly in line 55. This helps with the Law of Demeter.

```ruby
expect(Sufia::AdminSetService).to receive(:create).with(AdminSet, user).and_return(true)
```

jcoyne: I'm following the pattern layed out here: https://blog.engineyard.com/2014/keeping-your-rails-controllers-dry-with-services

I don't think the law of demeter is applicable, because neither option exposes more information than the other.

PR #2652, opened 2016-09-09
Configure the banner image to use

jcoyne:


lib/sufia/configuration.rb
```diff
@@ -7,6 +7,11 @@ def curation_concerns_config
     delegate(*(CurationConcerns.config.methods - Object.methods),
              to: :curation_concerns_config)
 
+    attr_writer :banner_image
+    def banner_image
+      @banner_image ||= "https://cloud.githubusercontent.com/assets/92044/18370978/88ecac20-75f6-11e6-8399-6536640ef695.jpg"
```
mjgiarlo: How much harder would it be to include a default image that is local to the application?

jcoyne: Not hard, but why? This just bloats the gem and we expect that most users are going to set their own image anyway.

mjgiarlo: @jcoyne If we're confident that link will be working for some time, it's no big deal.

PR #2654, opened 2016-09-09
Remove extraneous `puts` statements from specs.

mjgiarlo:


spec/features/create_work_spec.rb
```diff
@@ -33,9 +33,6 @@
       expect(page).to have_content('Please note, making something visible to the world (i.e. marking this as Public) may be viewed as publishing which could impact your ability to')
       # attach_file('Upload a file', fixture_file_path('files/image.png'))
       check('agreement')
-      puts "Required metadata: #{page.evaluate_script(%{$('#form-progress').data('save_work_control').requiredFields.areComplete})}"
```
jcoyne: These are not extraneous. These are debug for the script failing on travis. We need this if we ever hope to unskip this on travis.

awead: 1abadb

Issue #2532, opened 2016-08-24
Representative Media – Works and FileSets


mtribone:
### Descriptive summary

In Sufia 7, the style which controls the display of representative media from Curation Concerns is missing.

https://github.com/projecthydra/curation_concerns/blob/master/app/assets/stylesheets/curation_concerns/_positioning.scss#L137

### Expected behavior

Expect the representative media to be a block level element and have the actionable buttons below the representative media.

![screen shot 2016-08-24 at 10 17 11 am](https://cloud.githubusercontent.com/assets/4163828/17933930/f995a48c-69e3-11e6-843a-32c7cf87517b.png)

### Actual behavior

The representative media behaves inline and the actionable buttons wrap around the media.
### Steps to reproduce the behavior

1. Create a Work and add files to the work
1. View the work and fileset(s)

mtribone:
![screen shot 2016-08-24 at 10 07 50 am](https://cloud.githubusercontent.com/assets/4163828/17933793/6d05f79c-69e3-11e6-8e01-9de9531bfed1.png)

mtribone:
![screen shot 2016-08-24 at 10 14 14 am](https://cloud.githubusercontent.com/assets/4163828/17933822/8e4506be-69e3-11e6-83ec-77a9d8ef3505.png)

PR #2658, opened 2016-09-09
Add a style for representative media

jcoyne:
The fix from #2542 was too broad and brought too many styles that were
adversely affecting the Sufia layout.

See the initial issue: #2532


mjgiarlo:
@jcoyne can you show how the layout was adversely affected?

jcoyne:
@mjgiarlo for example, it was pulling in these styles, which we don't want:
https://github.com/projecthydra/curation_concerns/blob/master/app/assets/stylesheets/curation_concerns/_positioning.scss#L90-L99

PR #2663, opened 2016-09-12
checking for null or empty to work for selects and text areas

### LINKED ISSUE #2648: Javascript error When adding additional required fields
### Descriptive summary

In Sufia 7 master, when I add an additional required field then I start getting javascript errors.

The setup is the same as in #2643.

### Expected behavior

Additional field is required and I can still upload.

### Actual behavior

screen shot 2016-09-08 at 7 16 37 am

1. Modify .internal_test_app/app/forms/curation_concerns/generic_work_form.rb to include description as a required field

screen shot 2016-09-07 at 2 16 49 pm

1. Then go to the new work form

1. Look in the debug console after the page has loaded for the error

### Related work

#2643

cam156:
Fixes #2648 

Checks for null or length of zero to allow for other input types including multi select for resource_type.

### DIFF
--- a/app/assets/javascripts/sufia/save_work/required_fields.es6
+++ b/app/assets/javascripts/sufia/save_work/required_fields.es6
@@ -7,7 +7,11 @@ export class RequiredFields {
   }
 
   get areComplete() {
-    return this.requiredFields.filter((n, elem) => { return $(elem).val().length < 1 } ).length == 0
+    return this.requiredFields.filter((n, elem) => { return this.isValuePresent(elem) } ).length == 0
+  }
+
+  isValuePresent(elem) {
+    return ($(elem).val() === null) || ($(elem).val().length < 1)
   }
 
   // Reassign requiredFields because fields may have been added or removed.
--- a/spec/javascripts/required_field_spec.js
+++ b/spec/javascripts/required_field_spec.js
@@ -1,0 +1,41 @@
+describe('RequiredFields', function() {
+  var control = require('sufia/save_work/required_fields');
+  var target = null;
+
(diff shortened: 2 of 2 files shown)


app/assets/javascripts/sufia/save_work/required_fields.es6
```diff
@@ -7,7 +7,7 @@ export class RequiredFields {
   }
 
   get areComplete() {
-    return this.requiredFields.filter((n, elem) => { return $(elem).val().length < 1 } ).length == 0
+    return this.requiredFields.filter((n, elem) => { return ($(elem).val() == null) || ($(elem).val().length < 1) } ).length == 0
```
jcoyne: Can we extract this test into a new method? `isValuePresent(elem)` perhaps?

spec/javascripts/required_field_spec.js
```diff
@@ -0,0 +1,48 @@
...
+      });
+      it("is complete", function() {
+        expect(target.areComplete).toEqual(true);
+      });
+    });
+
+  });
+});
+
+function form(title, resource_type_selected) {
```
jcoyne: Please fix.

jcoyne: have you considered using `loadFixtures('fixtures/form_with_single_value.html')`?

cam156: @jcoyne are you suggesting putting this in a file?  I'm not 100% sure I can pass variables into the file.

jcoyne:
@cam156 I was suggesting having two html fixtures instead. It's NBD if you want to keep a function.

Issue #2491, opened 2016-08-17
Release notes for 7.2.0


mjgiarlo:
## Authority change
```
cp config/authorities/rights.yml config/authorities/licenses.yml
```

## Generator to run:
```
$ rails generate jquery:datatables:install bootstrap3
```

## Unreleased gem
```
gem 'flipflop', git: 'https://github.com/jcoyne/flipflop.git', branch: 'hydra'
```

https://github.com/voormedia/flipflop/issues/4
https://github.com/voormedia/flipflop/issues/8
https://github.com/voormedia/flipflop/issues/10

# New flipflop migration
```
$ bin/rails sufia:install:migrations
$ rake db:migrate
```

# Changed default layout

If you are using the sufia-one-column layout it has been moved: `app/views/layouts/sufia-one-column.html.erb → app/views/layouts/curation_concerns/1_column.html.erb`

Check your application.rb for `layout 'sufia-one-column'` and replace it with `with_themed_layout '1_column'`.

In addition if you have overridden sufia-one-column in your application you should move your override to the new location.

# Remove unneeded table (optional)
Sufia no longer includes `acts_as_follower` so if you're upgrading, and you don't want to keep this data, you can add a migration to remove the table:

```
class DropFollows < ActiveRecord::Migration
  def self.up
    drop_table :follows
  end
end
```

## Ordering changes
#2479 brought Rails 5 support to Sufia, and also brought support for RDF 2 which ends a long-standing bug in ActiveFedora that returns RDF statements in a predictable, and not directly alterable, order. In other words, if an item has more than one title, the first title may change whenever the item is loaded.

A workaround for preserving the old behavior is to add `gem 'rdf', '~> 1.99'` to your `Gemfile`, and ordering will be preserved.

jcoyne:
Blocked by 
- [ ] Release of AF 10.3 (which is blocked by https://github.com/projecthydra/active_fedora/pull/1140)
- [ ] https://github.com/projecthydra/curation_concerns/pull/912

mjgiarlo:
@jcoyne Now we just need a release of AF 10.3 -- those two PRs have been merged.

jcoyne:
AF 10.3.0.rc1 has been released and it seems to work, so I'll push 10.3.0.

mjgiarlo:
Need to include the new db migration that drops the `follows` table in the upgrade process. New generator?

jcoyne:
[...]

PR #2667, opened 2016-09-12
Delete controller override and just override layout

### LINKED ISSUE #1815: Override CurationConcerns PermissionsController layout without copypasta

cam156:
Fixes #1815 

Delete the controller override of CC that just sets the layout to nil.  This can be accomplished by overriding the 1_column layout with a copy of the sufia-one-column layout instead.

Also moves sufia-one-column to curation_concerns/1_column so we do not need to override the layout in all of the other controllers.


jcoyne:
I'd prefer we keep an extra controller rather than having to manage an overridden view.

cam156:
@jcoyne then you should comment on  #1815 to lobby to close it without change. I would think putting out own layout in would be something we should do.

cam156:
After a conversation in slack it was determined to move the sufia-one-column layout to the curation_concern/1_column so that sufia can better tap into the curation concerns theming.  Working on an update to the PR

cam156:
@jcoyne Better?

awead:
Discussed this morning.

awead:
@mjgiarlo Is there a way we can get the dependency check to run? We need this merged for the PSU sprint.

lib/generators/sufia/install_generator.rb
```diff
@@ -87,12 +87,6 @@ def inject_sufia_application_controller_behavior
       end
     end
 
-    def use_blacklight_layout_theme
-      file_path = "app/controllers/application_controller.rb"
-      return unless File.exist?(file_path)
-      gsub_file file_path, /with_themed_layout '1_column'/, "layout 'sufia-one-column'"
```
jcoyne: can you add a note about this in https://github.com/projecthydra/sufia/issues/2491 ?

jcoyne:
@cam156 Can we remove https://github.com/projecthydra/sufia/blob/master/app/views/layouts/sufia-one-column.html.erb now?

app/controllers/concerns/sufia/batch_uploads_controller_behavior.rb
```diff
@@ -5,7 +5,6 @@ module BatchUploadsControllerBehavior
     include CurationConcerns::CurationConcernController
 
     included do
-      layout "sufia-one-column"
```
mjgiarlo: Does this controller behavior not need `with_themed_layout '1_column'` or is that accounted for elsewhere?

app/controllers/concerns/sufia/works_controller_behavior.rb
```diff
@@ -7,7 +7,6 @@ module WorksControllerBehavior
 
     included do
       self.show_presenter = Sufia::WorkShowPresenter
-      layout "sufia-one-column"
```
mjgiarlo: Does this controller behavior not need `with_themed_layout '1_column'` or is that accounted for elsewhere?

lib/generators/sufia/install_generator.rb
mjgiarlo: :+1:

app/controllers/sufia/admin_sets_controller.rb
```diff
@@ -1,7 +1,6 @@
 module Sufia
   class AdminSetsController < ApplicationController
     include CurationConcerns::CollectionsControllerBehavior
-    layout "sufia-one-column"
```
mjgiarlo: Is this picking up the themed layout from one of its dependencies?

app/controllers/concerns/sufia/works_controller_behavior.rb
cam156: @mjgiarlo any controller that includes on inherits from CurationConcerns::CollectionsControllerBehavior gets `with_themed_layout '1_column'` by default.  No need to override or set again.

app/controllers/sufia/admin_sets_controller.rb
cam156: @mjgiarlo any controller that includes on inherits from CurationConcerns::CollectionsControllerBehavior gets `with_themed_layout '1_column'` by default.  No need to override or set again.

app/controllers/concerns/sufia/batch_uploads_controller_behavior.rb
cam156: @mjgiarlo any controller that includes on inherits from CurationConcerns::CollectionsControllerBehavior gets `with_themed_layout '1_column'` by default.  No need to override or set again.

cam156:
@jcoyne I moved it, so in essence it is getting removed.  That is all the sufia specific headers and menus so we want that code.  This PR just puts it in the place curation concerns will find it. from files changed: `app/views/layouts/sufia-one-column.html.erb → ...youts/curation_concerns/1_column.html.erb`

app/controllers/concerns/sufia/works_controller_behavior.rb
mjgiarlo: Excellent, thanks!

PR #2676, opened 2016-09-13
New works are inactive when mediation is on

### LINKED ISSUE #2616: When an object is deposited and mediated deposit is enabled, the object state should be set to Inactive
All works and file sets are marked as inactive (in draft form) when deposited, if the mediated deposit feature flipper is on. To mark an object as inactive, set its `state` property to `::RDF::URI('http://fedora.info/definitions/1/0/access/ObjState#inactive'`. See https://github.com/projecthydra/curation_concerns/pull/928/files for the work that enabled this feature.

Decomposes #2452. Further decomposed into

hortongn:
Fixes #2616 ; refs #2616

If mediated deposit is flipped on, new works are marked inactive.

### DIFF
--- a/app/actors/sufia/create_with_files_actor.rb
+++ b/app/actors/sufia/create_with_files_actor.rb
@@ -3,7 +3,7 @@ module Sufia
   class CreateWithFilesActor < CurationConcerns::Actors::AbstractActor
     def create(attributes)
       self.uploaded_file_ids = attributes.delete(:uploaded_files)
-      validate_files && next_actor.create(attributes) && attach_files
+      validate_files && mark_inactive && next_actor.create(attributes) && attach_files
     end
 
     def update(attributes)
@@ -41,6 +41,15 @@ module Sufia
       def uploaded_files
         return [] if uploaded_file_ids.empty?
         @uploaded_files ||= UploadedFile.find(uploaded_file_ids)
+      end
+
+      def mark_inactive
+        return true unless Flipflop.enable_mediated_deposit?
+        curation_concern.state = inactive_uri
+      end
+
+      def inactive_uri
+        ::RDF::URI('http://fedora.info/definitions/1/0/access/ObjState#inactive')
       end
   end
 end
--- a/spec/actors/sufia/create_with_files_actor_spec.rb
+++ b/spec/actors/sufia/create_with_files_actor_spec.rb
@@ -44,4 +44,22 @@ describe Sufia::CreateWithFilesActor do
       end
     end
   end
+
+  describe "mediated deposit" do
+    subject { actor.curation_concern.state }
+    let(:inactive_uri) { RDF::URI('http://fedora.info/definitions/1/0/access/ObjState#inactive') }
+    before do
+      allow(Flipflop).to receive(:enable_mediated_deposit?).and_return(mediation_enabled)
+      allow(AttachFilesToWorkJob).to receive(:perform_later).with(GenericWork, [uploaded_file1, uploaded_file2])
+      actor.create(attributes)
+    end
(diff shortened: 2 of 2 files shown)


hortongn:
@mjgiarlo Before I move ahead with this issue, I'm pushing it up to get some feedback on if I'm putting the code in the appropriate place.  The code works as it is to set new works as inactive if the mediated deposit flipper is turned on.  Since I'm still learning my way around Sufia, I'm not sure if CreatesWithFilesActor is the best place to set the object state.  Thoughts?

app/actors/sufia/create_with_files_actor.rb
```diff
@@ -42,5 +43,13 @@ def uploaded_files
         return [] if uploaded_file_ids.empty?
         @uploaded_files ||= UploadedFile.find(uploaded_file_ids)
       end
+
+      def set_publication_state
+        self.curation_concern.state = unpublished if Flipflop.enable_mediated_deposit?
```
mjgiarlo: This is very minor, but would you consider:

1. renaming the new methods;
1. removing the explicit `self`; and 
1. reworking this method with a guard clause?

```ruby
def mark_inactive
  return unless Flipflop.enable_mediated_deposit?
  curation_concern.state = inactive_uri
end

def inactive_uri
  ::RDF::URI('http://fedora.info/definitions/1/0/access/ObjState#inactive')
end
```

app/actors/sufia/create_with_files_actor.rb
```diff
@@ -2,6 +2,7 @@ module Sufia
   # Creates a work and attaches files to the work
   class CreateWithFilesActor < CurationConcerns::Actors::AbstractActor
     def create(attributes)
+      set_publication_state
       self.uploaded_file_ids = attributes.delete(:uploaded_files)
       validate_files && next_actor.create(attributes) && attach_files
```
mjgiarlo: @hortongn It makes sense to put your change in this actor, yes. We may eventually want to break this functionality out into its own actor but it's early days to work on that abstraction, I feel. You may want to add this to the chain of operations on line 7:

```ruby
validate_files && mark_inactive && next_actor.create(attributes) && attach_files
```

hortongn:
@mjgiarlo Thanks for the feedback!  Exactly what I was looking for.

spec/actors/sufia/create_with_files_actor_spec.rb
```diff
@@ -44,4 +44,28 @@
       end
     end
   end
+
+  context "with mediated deposit" do
+    let(:inactive) { ::RDF::URI('http://fedora.info/definitions/1/0/access/ObjState#inactive') }
+    context "flipped on" do
```
mjgiarlo: Could you apply a small refactoring to DRY the new specs out a little? Here's what I had in mind

```ruby
describe "mediated deposit" do
  subject { actor.curation_concern.state }
  let(:inactive_uri) { RDF::URI('http://fedora.info/definitions/1/0/access/ObjState#inactive') }
  before do
    allow(FlipFlop).to receive(:enable_mediated_deposit?).and_return(mediated_enabled)
    allow(AttachFilesToWorkJob).to receive(:perform_later).with(GenericWork, [uploaded_file1, uploaded_file2])
    actor.create(attributes)
  end
  context "when enabled" do
    let(:mediation_enabled) { true }
    it { is_expected.to eq inactive_uri }
  end
  context "when disabled" do
    let(:mediation_enabled) { false }
    it { is_expected.to be nil }
  end
end
```

hortongn:
@mjgiarlo I refactored the spec and re-pushed.  All checks have passed.  Thanks for the input!

PR #2679, opened 2016-09-13
Remove header. Panel is labeled by the tab

jcoyne:


mtribone:
@projecthydra/sufia-ui-ux-advisors We need some 👀  on this PR in the context of the home page. Do we need headings if the tabs are labeled?

mjgiarlo:
@jcoyne Would it help the @projecthydra/sufia-ui-ux-advisors if a screenshot were included illustrating the change?

jcoyne:
Before:
screen shot 2016-09-13 at 10 51 41 am

After:
screen shot 2016-09-13 at 10 52 24 am

justcolin:
Hmm, the heading is useful for screenreader users (they use it to skip around the page). Would it be possible to have the unlinked tab to be encoded as an `h#` element?

jcoyne:
@justcolin should there be similar headers on the other tabs? Currently there aren't.

justcolin:
@jcoyne: Yup! Almost every significant section should have a header.

justcolin:
That being said, if it were removed it wouldn't be a critical error, more of a nuisance.

mjgiarlo:
Given what @justcolin has shared, how would you like to proceed on this one, @jcoyne?

jcoyne:
Closing. I added the `sr-only` class to that `h2` (and added missing headers for other tabs) in a different PR

PR #2688, opened 2016-09-14
Adds total item count for admin sets.

### LINKED ISSUE #2633: on the admin set display page, show the number of works in each set
This requires a change to the indexing logic for an admin set.  Or override `total_items` on the `AdminSetPresenter` (see https://github.com/projecthydra/curation_concerns/blob/master/app/presenters/curation_concerns/collection_presenter.rb#L29-L31)

https://ggeisler.github.io/lerna-dashboard/sdr-admin-sets-list.html

Ref (#2501)

andjsmit:
Fixes #2633

Adds works count for each admin set on index page. 

### DIFF
--- a/app/presenters/sufia/admin_set_presenter.rb
+++ b/app/presenters/sufia/admin_set_presenter.rb
@@ -1,4 +1,7 @@
 module Sufia
   class AdminSetPresenter < CollectionPresenter
+    def total_items
+      ActiveFedora::SolrService.count("{!field f=isPartOf_ssim}#{id}", fl: id)
+    end
   end
 end
--- a/app/views/sufia/admin/admin_sets/index.html.erb
+++ b/app/views/sufia/admin/admin_sets/index.html.erb
@@ -19,6 +19,7 @@
                   <th>Title</th>
                   <th>Date created</th>
                   <th>Creator</th>
+                  <th>Works</th>
                 </tr>
               </thead>
               <tbody>
@@ -27,6 +28,7 @@
                     <td><%= link_to admin_set.title.first, [sufia, :admin, admin_set] %></td>
                     <td><%= admin_set.create_date %></td>
                     <td><%= safe_join(admin_set.creator) %></td>
+                    <td><%= controller.presenter_class.new(admin_set, current_ability).total_items %></td>
                   </tr>
                 <% end %>
               </tbody>
--- a/spec/views/sufia/admin/admin_sets/index.html.erb_spec.rb
+++ b/spec/views/sufia/admin/admin_sets/index.html.erb_spec.rb
@@ -1,0 +1,27 @@
+describe "sufia/admin/admin_sets/index.html.erb", type: :view do
+  context "when no admin sets exists" do
+    it "alerts users there are no admin sets" do
+      render
+      expect(rendered).to have_content("No administrative sets have been created.")
+    end
+  end
+
+  context "when an admin set exists" do
+    let(:admin_set) { build(:admin_set, id: '123', title: ['Example Admin Set'], creator: ['jdoe@example.com']) }
+    let(:solr_doc) { SolrDocument.new(admin_set.to_solr) }
+    let(:admin_sets) { [solr_doc] }
+    let(:presenter_class) { Sufia::AdminSetPresenter }
+    let(:ability) { instance_double("Ability") }
+    before do
+      allow(controller).to receive(:current_ability).and_return(ability)
+      allow(controller).to receive(:presenter_class).and_return(presenter_class)
+      assign(:admin_sets, admin_sets)
+    end
+    it "lists admin set" do
+      render
+      expect(rendered).to have_content('Example Admin Set')
+      expect(rendered).to have_content('jdoe@example.com')
+      expect(rendered).to have_css("td", text: /^0$/)
+    end
+  end
+end
--- a/spec/presenters/sufia/admin_set_presenter_spec.rb
+++ b/spec/presenters/sufia/admin_set_presenter_spec.rb
@@ -1,0 +1,26 @@
+require 'spec_helper'
+
+describe Sufia::AdminSetPresenter do
+  let(:admin_set) { build(:admin_set,
+                          id: '123',
+                          description: ['An example admin set.'],
+                          title: ['Example Admin Set Title']) }
+  let(:work) { build(:work, title: ['Example Work Title'], admin_set_id: '123') }
+  let(:solr_document) { SolrDocument.new(admin_set.to_solr) }
+  let(:ability) { double }
+  let(:presenter) { described_class.new(solr_document, ability) }
+
+  describe "total_items" do
+    subject { presenter.total_items }
+    context "empty admin set" do
+      it { is_expected.to eq 0 }
+    end
+    context "admin set with work" do
+      before do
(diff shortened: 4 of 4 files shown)


app/controllers/sufia/admin/admin_sets_controller.rb
```diff
@@ -30,6 +30,7 @@ def index
       add_breadcrumb t(:'sufia.toolbar.admin.menu'), sufia.admin_path
       add_breadcrumb t(:'sufia.admin.sidebar.admin_sets'), sufia.admin_admin_sets_path
       @admin_sets = CurationConcerns::AdminSetService.new(self).search_results(:read)
+      @presenter = presenter_class.new(self, current_ability)
```
jcoyne: The first argument to AdminSetPresenter#initialize is a SolrDocument (see https://github.com/projecthydra/curation_concerns/blob/master/app/presenters/curation_concerns/collection_presenter.rb#L11).  You've passed a instance of the controller.

app/presenters/sufia/admin_set_presenter.rb
```diff
@@ -1,4 +1,7 @@
 module Sufia
   class AdminSetPresenter < CollectionPresenter
+    def total_items(admin_set = self)
```
jcoyne: In a presenter the admin_set would be as `self.solr_document` (as you want the id you could just call `id`)

app/views/sufia/admin/admin_sets/index.html.erb
```diff
@@ -27,6 +28,7 @@
                     <td><%= link_to admin_set.title.first, [sufia, :admin, admin_set] %></td>
                     <td><%= admin_set.create_date %></td>
                     <td><%= safe_join(admin_set.creator) %></td>
+                    <td><%= @presenter.total_items(admin_set) %></td>
```
jcoyne: How about `controller.presenter_class.new(admin_set).total_items` here.

spec/controllers/sufia/admin/admin_sets_controller_spec.rb
```diff
@@ -33,6 +33,11 @@
         expect(response).to be_success
         expect(assigns[:admin_sets]).to be_kind_of Array
       end
+
+      it 'defines a presenter' do
+        get :index
+        expect(assigns[:presenter]).to be_kind_of Sufia::AdminSetPresenter
```
jcoyne: Please combine the assertion into the previous test.  Otherwise we have the overhead of multip test/request setup for a single assertion. Our test suite is too slow as is.

spec/views/sufia/admin/admin_sets/index.html.erb_spec.rb
```diff
@@ -0,0 +1,30 @@
+describe "sufia/admin/admin_sets/index.html.erb", type: :view do
+  let(:document) { SolrDocument.new }
+  let(:ability) { double }
+  let(:presenter) { Sufia::AdminSetPresenter(document, ability) }
```
jcoyne: missing `.new` ?

app/views/sufia/admin/admin_sets/index.html.erb
andjsmit: I wasn't able to get this to work. Is the idea to remove the need to declare @presenter in the controller? Would additional code be needed to define controller? Replaced as given did not work for me.

app/presenters/sufia/admin_set_presenter.rb
```diff
@@ -1,4 +1,7 @@
 module Sufia
   class AdminSetPresenter < CollectionPresenter
+    def total_items(admin_set = solr_document)
```
mjgiarlo: If I'm tracing this correctly (looking at `CollectionPresenter` in CC), the `AdminSetPresenter` should already have the necessary solr document as in ivar sitting behind an `attr_accessor`, so I'm not sure this method needs any args.

app/presenters/sufia/admin_set_presenter.rb
```diff
@@ -1,4 +1,7 @@
...
+      ActiveFedora::SolrService.query("{!field f=isPartOf_ssim}#{admin_set.id}", fl: admin_set.id).length
```
mjgiarlo: `admin_set` here can be `solr_document`.

app/controllers/sufia/admin/admin_sets_controller.rb
```diff
@@ -29,6 +29,7 @@ def index
       add_breadcrumb t(:'sufia.toolbar.admin.menu'), sufia.admin_path
       add_breadcrumb t(:'sufia.admin.sidebar.admin_sets'), sufia.admin_admin_sets_path
       @admin_sets = CurationConcerns::AdminSetService.new(self).search_results(:read)
+      @presenter = presenter_class.new(SolrDocument, current_ability)
```
mjgiarlo: @andjsmit I now realize the ticket you're working on could be clearer -- there are *two* bits of AdminSet work in Sufia at the moment: the stuff in the admin dashboard and the stuff outside the admin dashboard. I believe the ticket you're working on is for the stuff *not* in the admin dashboard, so you should be using the AdminSetsController not under `sufia/admin/`, if that makes sense.

app/views/sufia/admin/admin_sets/index.html.erb
jcoyne: The presenter is an object that presents a record. In this case that record should be `admin_set`, so you can either declare it here or in the controller, but there should be no need to pass arguments to it's methods.

app/controllers/sufia/admin/admin_sets_controller.rb
jcoyne: Actually, I think it's needed in both places.

mjgiarlo: I don't disagree; I thought it might be worth splitting that into two PRs.

jcoyne: The presenter class should take an instance of SolrDocument (the elements in `@admin_sets`), not the class as it's first argument.

app/presenters/sufia/admin_set_presenter.rb
jcoyne: Use `ActiveFedora::SolrService.count` instead of `.query`

app/views/sufia/admin/admin_sets/index.html.erb
andjsmit: OK, got it. Code is working by placing `<td><%= controller.presenter_class.new(admin_set, current_ability).total_items %></td>` in view. However, I'm not sure how to account for this in the view test. Test keeps giving me `undefinded method 'presenter_class'`

jcoyne: You are going to set an expectation on the controller like so:
```ruby
allow(controller).to receive(:presenter_class).and_return(WhateverPresenter)
```

andjsmit: This works as long as `include Devise::Test::ControllerHelpers` is added to the context. Otherwise test errors with Devise could not find the `Warden::Proxy` instance on your request environment.

mjgiarlo:
@jcoyne is this ready to go?

Issue #2640, opened 2016-09-07
Inactive objects (works and file sets) should not be discoverable via the catalog controller and their show pages should not render.


mjgiarlo:
(May require further decomposition.)

When the mediated deposit feature flipper is on and works are deposited, said works and their file sets are marked as inactive. These inactive works and file sets should not be discoverable via the catalog controller (searches should not turn up results and facet values should not appear).

The following issues address other functionality associated with the visibility of works in inactive state due to mediated deposit.

#2727 
When mediated deposit is enabled, and a work is in inactive state because it is in the mediated deposit workflow and has not yet been approved, the work’s show views should error, regardless of the visibility specified.

#2728 
When mediated deposit is enabled, and a work is in inactive state because it is in the mediated deposit workflow and has not yet been approved, but the work has been marked as visible to other user(s), the work should not be visible by those user(s) and those user(s) should not yet receive a notification that the work is available.

#2729 
When mediated deposit is enabled, and a work is in inactive state because it is in the mediated deposit workflow and has not yet been approved, it may appear in the dashboard of the depositor. (So that the owner can find them)

Follows from #2616

hortongn:
@blancoj Can you edit this issue's description to remove the parts that you moved into new issues?  Maybe reference those issue numbers in the description so we know they moved?

When I see your new PR for this issue, I'll review it again.  Thanks!

blancoj:
I've updated the description, describing the decomposition of this issue into further issues.

mjgiarlo:
Closing because it's been decomposed into non-epics.

PR #2691, opened 2016-09-14
make inactive records undiscoverable.

blancoj:
I'm working on #2640 (if work is inactive then don't have it show up in facet or search).  I think what I have to do is change app/search_builders/sufia/catalog_search_builder.rb so that it will not include the works that are inactive in the results by adding a method to filter them out.  But since the state is not in solr ( I don't think ), I'm not sure how to check for this.  Am I headed in the right direction?  Any suggestions?  Thank you! Jose

BTW, the only file of interest is this one.  I changed the other files in my local repository based on issues that have not yet been merged.

app/search_builders/sufia/catalog_search_builder.rb

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -9,6 +9,8 @@ group :development, :test do
   gem 'coveralls', require: false
 end
 
+gem 'hydra-role-management'
+
 # BEGIN ENGINE_CART BLOCK
 # engine_cart: 0.10.0
 # engine_cart stanza: 0.10.0
--- a/app/search_builders/sufia/catalog_search_builder.rb
+++ b/app/search_builders/sufia/catalog_search_builder.rb
@@ -2,7 +2,8 @@ class Sufia::CatalogSearchBuilder < Sufia::SearchBuilder
   self.default_processor_chain += [
     :add_access_controls_to_solr_params,
     :add_advanced_parse_q_to_solr,
-    :show_works_or_works_that_contain_files
+    :show_works_or_works_that_contain_files,
+    :show_only_active_records
   ]
 
   # show both works that match the query and works that contain files that match the query
@@ -10,6 +11,13 @@ class Sufia::CatalogSearchBuilder < Sufia::SearchBuilder
     return if solr_parameters[:q].blank?
     solr_parameters[:user_query] = solr_parameters[:q]
     solr_parameters[:q] = new_query
+  end
+
+  # show works that are in the active state.
+  def show_only_active_records(solr_parameters)
+    return unless Flipflop.enable_mediated_deposit?
+    solr_parameters[:fq] ||= []
+    solr_parameters[:fq] << '-suppressed_bsi:true'
   end
 
   private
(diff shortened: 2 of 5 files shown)


mjgiarlo:
@blancoj  an work's publication state is solrized here -- https://github.com/projecthydra/curation_concerns/blob/master/app/indexers/curation_concerns/work_indexer.rb#L10 -- so it should be in Solr.

app/search_builders/sufia/catalog_search_builder.rb
```diff
@@ -12,6 +13,15 @@ def show_works_or_works_that_contain_files(solr_parameters)
     solr_parameters[:q] = new_query
   end
 
+
+  # show works that are in the active state.
+  def show_works_or_works_that_are_active(solr_parameters)
```
mjgiarlo: Shorten the name to something like `show_only_active_works`

app/search_builders/sufia/catalog_search_builder.rb
```diff
@@ -12,6 +13,15 @@ def show_works_or_works_that_contain_files(solr_parameters)
...
+
+      #Not sure sure what goes here.
+      #I want to say something like this, but I know that is not right.
+      return false if actor.curation_concern.state == Inactive
```
mjgiarlo: I'd check with @jcoyne here, but I imagine you want something like this here:

```ruby
solr_parameters[:fq] ||= []
solr_parameters[:fq] << '-suppressed_bsi:true'
```

jcoyne:
It looks like you'll need to fix up the style issues that rubocop has found too.

hortongn:
@blancoj I tested this and it does prevent inactive works from being discoverable through the search box.  It also properly makes works discoverable when mediated deposit is flipped off.  So that part of this issue appears to be working correctly.  Code looks fine to me.

There are a few other parts of this issue that are not yet implemented (see below).  So you should either implement these pieces or open new issue(s) for them and change this issue so the scope is just hiding the work from search.

1. their show views should error, regardless of the visibility specified. 
1. If the objects are shared with other users, they are not yet visible by those users and those users should not yet receive a notification. 
1. They may appear in the dashboard of the depositor. (So the owner can find them)

jcoyne:
@njaffer I don't think this should have been merged. @mjgiarlo's comments were not addressed.

Gemfile
```diff
@@ -9,6 +9,8 @@ group :development, :test do
   gem 'coveralls', require: false
 end
 
+gem 'hydra-role-management'
```
jcoyne: Why is this required?

blancoj:
I needed to be able to enable and disable mediated deposit, and followed instructions here:

https://github.com/projecthydra/sufia/wiki/Making-Admin-Users-in-Sufia

that required this Gem.

jcoyne:
You could have just mocked the admin group in your test.  We don't want to introduce a dependency on that gem as many people handle group management in LDAP and don't want the database involved.

mjgiarlo:
@blancoj can you pls submit a follow-up PR that removes hydra-role-management from the Gemfile?

blancoj:
done

PR #2699, opened 2016-09-15
Use UTF8 encoding for Endnote files

awead:
Sorry, forgot to add this earlier.

### DIFF
--- a/app/controllers/concerns/sufia/works_controller_behavior.rb
+++ b/app/controllers/concerns/sufia/works_controller_behavior.rb
@@ -63,7 +63,7 @@ module Sufia
       def additional_response_formats(format)
         format.endnote do
           send_data(presenter.solr_document.export_as_endnote,
-                    type: "application/x-endnote-refer",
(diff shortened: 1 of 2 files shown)


awead:
I don't think this is needed because config.action_dispatch.default_charset should cover that.

cam156:
@awead do we have a test that makes sure config.action_dispatch.default_charset is set it does indeed end up in the output to the user?  I wonder if send data is overriding the default.

awead:
@cam156 no, the setting wasn't there before. FWIW, I'm not seeing any difference with the option set and/or with `charset=utf-8` in the header. The file is downloaded as `charset=us-ascii` each time, unless I'm not checking it correctly.

awead:
@cam156 ok, I can now confirm utf-8 download. I just needed to use non-ascii characters in the title. It's downloading as utf-8 _without_ adding the charset in the or even specifying it in the config... so meh. I'm going to add the config option in anyway in Scholarsphere, but we don't need the add the charset in the header, I'm closing this PR.

Issue #2728, opened 2016-09-21
The inactive state of mediated deposit workflow supersedes the visibility settings for the object


blancoj:
When mediated deposit is enabled, and a work is in inactive state because it is in the mediated deposit workflow and has not yet been approved, but the work has been marked as visible to other user(s), the work should not be visible by those user(s) and those user(s) should not yet receive a notification that the work is available.

mjgiarlo:
Include as part of 1+-step workflow.

mjgiarlo:
@vantuyls Should we defer this to Phase 2, operating under the assumption that Phase 1 is for 0-step?

mjgiarlo:
@vantuyls how much do we really care about this one? I'm moving it to phase 2 for now, but it'd be good to see if this is still relevant and highly desired. (I wonder if this can collapse into the "editing an object sends it back through workflow" story?)

vantuyls:
If we're sure that the object will remain hidden from view when it is in the workflow, then OK. Otherwise, this needs to be an issue we address explicitly.

mjgiarlo:
If user A shares an in-workflow work with user B, the work will be hidden from B in search results and the show view assuming B lacks permission to advance the work from its current workflow state. I'm not sure if it's also hidden from B's dashboard. This would be worth testing/QAing. Might you have cycles for this, @vantuyls?

PR #2733, opened 2016-09-22
shared works should not be available if in mediated deposit and inact…

blancoj:
This pull request is to address issue #2728 .  With this change when a work is shared, and the work is inactive and mediated deposit is enabled, the work is not available from the Dashboard/Shares page until the work is no longer inactive.  

Issue #2728 also talks about not having notifications sent out, but I did not see any notifications sent out when not in mediated deposit, and the work is shared and published, so I think for now there is no change needed regarding this.

### DIFF
--- a/app/search_builders/sufia/my_shares_search_builder.rb
+++ b/app/search_builders/sufia/my_shares_search_builder.rb
@@ -9,5 +9,7 @@ class Sufia::MySharesSearchBuilder < Sufia::SearchBuilder
     solr_parameters[:fq] += [
       "-" + ActiveFedora::SolrQueryBuilder.construct_query_for_rel(depositor: scope.current_user.user_key)
     ]
+
+    solr_parameters[:fq] += ['-suppressed_bsi:true'] if Flipflop.enable_mediated_deposit?
   end
 end
--- a/spec/search_builder/sufia/my_shares_search_builder_spec.rb
+++ b/spec/search_builder/sufia/my_shares_search_builder_spec.rb
@@ -8,18 +8,42 @@ describe Sufia::MySharesSearchBuilder do
                        current_user: me) }
   let(:builder) { described_class.new(scope) }
 
+  let(:solr_params) { { q: user_query } }
+
   before do
     allow(builder).to receive(:gated_discovery_filters).and_return(["access_filter1", "access_filter2"])
     allow(ActiveFedora::SolrQueryBuilder).to receive(:construct_query_for_rel)
       .with(depositor: me.user_key)
       .and_return("depositor")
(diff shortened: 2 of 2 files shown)


hortongn:
@blancoj please update your local mediated_deposit_epic branch and the rebase this branch against mediated_deposit_epic and push up again.  You have the same failing test I had on a recent PR and pulling in the other recent commits fixed it.  Thanks

app/search_builders/sufia/my_shares_search_builder.rb
```diff
@@ -9,5 +9,8 @@ def show_only_shared_files(solr_parameters)
     solr_parameters[:fq] += [
       "-" + ActiveFedora::SolrQueryBuilder.construct_query_for_rel(depositor: scope.current_user.user_key)
     ]
+
+    return unless Flipflop.enable_mediated_deposit?
```
mjgiarlo: This method used to return what was returned by:

```ruby
solr_parameters[:fq] += [
        "-" + ActiveFedora::SolrQueryBuilder.construct_query_for_rel(depositor: scope.current_user.user_key)
      ]
```

and now it returns `nil` if mediated deposit is off. Would this work instead:

```ruby
solr_parameters[:fq] += ['-suppressed_bsi:true'] if Flipflop.enable_mediated_deposit?
```

?

spec/search_builder/sufia/my_shares_search_builder_spec.rb
```diff
@@ -22,4 +24,28 @@
...
                            "-depositor"]
   end
+
+  describe "mediated deposit" do
+    before do
+      allow(Flipflop).to receive(:enable_mediated_deposit?).and_return(mediation_enabled)
+    end
+
+    context "with mediated deposit enabled" do
+      let(:user_query) { nil }
```
mjgiarlo: Since this `let()` statement has the same value in both of the new context blocks, should it sit above them, beneath the new `before` block?

spec/search_builder/sufia/my_shares_search_builder_spec.rb
```diff
@@ -22,4 +24,28 @@
...
+      let(:mediation_enabled) { true }
+      it "does includes suppressed switch" do
+        builder.show_only_shared_files(solr_params)
```
mjgiarlo: Should this also live in the parent `before` block since it's common to both new specs?

mjgiarlo:
@blancoj the build also looks like it needs to be fixed.

blancoj:
@mjgiarlo I fixed the rubocop  error.  It passes the tests now.

PR #2738, opened 2016-09-23
2627 mediated user notification

### LINKED ISSUE #2627: Send notification to user when item in workflow needs their attention

hortongn:
Fixes #2627 ; refs #2627

Notifies admin users when mediated works are created.

@vantuyls @blancoj @randalldfloyd @scholarworks 

### DIFF
--- a/app/actors/sufia/create_with_files_actor.rb
+++ b/app/actors/sufia/create_with_files_actor.rb
@@ -1,9 +1,12 @@
 module Sufia
   # Creates a work and attaches files to the work
   class CreateWithFilesActor < CurationConcerns::Actors::AbstractActor
+    include Rails.application.routes.url_helpers
+    include ActionDispatch::Routing::PolymorphicRoutes
+
     def create(attributes)
       self.uploaded_file_ids = attributes.delete(:uploaded_files)
-      validate_files && mark_inactive && next_actor.create(attributes) && attach_files
+      validate_files && next_actor.create(attributes) && attach_files && mark_inactive
     end
 
     def update(attributes)
@@ -46,10 +49,32 @@ module Sufia
       def mark_inactive
         return true unless Flipflop.enable_mediated_deposit?
         curation_concern.state = inactive_uri
+        send_inactive_message
       end
 
       def inactive_uri
         ::RDF::URI('http://fedora.info/definitions/1/0/access/ObjState#inactive')
+      end
+
+      def send_inactive_message
+        return true if receiving_users.empty?
+        receiving_users.each do |recipient|
+          message = "#{user.email} created <a href='#{work_path}'>#{curation_concern.to_s}</a>"
+          ::User.batch_user.send_message(recipient, message, 'New work to review')
+        end
+      end
+
+      # Gather all admin users (if the admin role is defined)
+      def receiving_users
+        recipients = []
+        ::User.find_each do |user|
+          recipients << user if (user.respond_to? :admin?) && user.admin?
+        end
+        recipients
+      end
+
+      def work_path
+        polymorphic_path(curation_concern)
       end
   end
 end
--- a/spec/actors/sufia/create_with_files_actor_spec.rb
+++ b/spec/actors/sufia/create_with_files_actor_spec.rb
@@ -48,6 +48,10 @@ describe Sufia::CreateWithFilesActor do
   describe "mediated deposit" do
     subject { actor.curation_concern.state }
     let(:inactive_uri) { RDF::URI('http://fedora.info/definitions/1/0/access/ObjState#inactive') }
+    let!(:admin) { create(:user) }
+    let!(:non_admin) { create(:user) }
+    let(:admin_inbox) { admin.mailbox.inbox }
+    let(:non_admin_inbox) { non_admin.mailbox.inbox }
     before do
       allow(Flipflop).to receive(:enable_mediated_deposit?).and_return(mediation_enabled)
       allow(AttachFilesToWorkJob).to receive(:perform_later).with(GenericWork, [uploaded_file1, uploaded_file2])
@@ -56,10 +60,27 @@ describe Sufia::CreateWithFilesActor do
     context "when enabled" do
       let(:mediation_enabled) { true }
       it { is_expected.to eq inactive_uri }
+      context "non-admin users" do
+        it "do not receive a new work notification" do
+          expect(non_admin_inbox.count).to eq(0)
+        end
+      end
+      context "admin users" do
+        before do
+          allow_any_instance_of(User).to receive(:admin?).and_return(true)
+          actor.create(attributes)
+        end
+        it "receive a new work notification" do
+          expect(admin_inbox.count).to eq(1)
+        end
+      end
     end
     context "when disabled" do
       let(:mediation_enabled) { false }
       it { is_expected.to be nil }
+      it "does not deliver a notification to admins" do
+        expect(admin_inbox.count).to eq(0)
+      end
     end
   end
 end
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -113,4 +113,5 @@ RSpec/AnyInstance:
     - 'spec/presenters/sufia/work_usage_spec.rb'
     - 'spec/presenters/sufia/file_usage_spec.rb'
     - 'spec/services/sufia/repository_audit_service_spec.rb'
+    - 'spec/actors/sufia/create_with_files_actor_spec.rb'
 


jcoyne:
Please hold off on merging this branch into master until we discuss merging this work with the workflows in curation_concerns.

app/actors/sufia/create_with_files_actor.rb
```diff
@@ -3,7 +3,7 @@ module Sufia
   class CreateWithFilesActor < CurationConcerns::Actors::AbstractActor
     def create(attributes)
       self.uploaded_file_ids = attributes.delete(:uploaded_files)
-      validate_files && next_actor.create(attributes) && attach_files
+      validate_files && next_actor.create(attributes) && attach_files && mark_inactive
```
jcoyne: This should go into a new actor, not amalgamated into this one.

app/actors/sufia/create_with_files_actor.rb
```diff
@@ -42,5 +42,36 @@ def uploaded_files
         return [] if uploaded_file_ids.empty?
         @uploaded_files ||= UploadedFile.find(uploaded_file_ids)
       end
+
+      def mark_inactive
+        return true unless Flipflop.enable_mediated_deposit?
+        curation_concern.state = inactive_uri
+        send_inactive_message
```
jcoyne: This should happen in a background job as it can be a slow operation.

app/actors/sufia/create_with_files_actor.rb
```diff
@@ -42,5 +42,36 @@ def uploaded_files
...
+        receiving_users.each do |admin|
+          message = "#{user.name} created <a href='#{work_path}'>#{curation_concern.title.first}</a>"
+          ::User.batch_user.send_message(admin, message, 'New work to review')
+        end
+      end
+
+      # Gather all admin users (if the admin role is defined)
+      def receiving_users
+        receivers = []
+        ::User.find_each do |user|
```
jcoyne: This is going to become an extremely slow operation as the size of the user table grows. Especially if this requires hitting LDAP for each user.

It probably makes sense to abstract this logic into a service like `FindMemberOfGroup.call('admin')` or similar.

app/models/concerns/curation_concerns/publishable.rb
```diff
@@ -0,0 +1,18 @@
+module CurationConcerns
+  module Publishable
```
jcoyne: This exists in CurationConcerns gem. We should not replace it here.

hortongn:
@jcoyne Sorry, the base branch for this PR should have been mediated_deposit_epic, not master.  I'll fix that now.

app/actors/sufia/create_with_files_actor.rb
```diff
@@ -46,10 +46,32 @@ def uploaded_files
...
         curation_concern.state = inactive_uri
+        send_inactive_message
       end
 
       def inactive_uri
         ::RDF::URI('http://fedora.info/definitions/1/0/access/ObjState#inactive')
       end
+
+      def send_inactive_message
+        return true if receiving_users == []
```
mjgiarlo: Instead of comparing with a literal here, could you ask if `receiving.users.empty?`

app/actors/sufia/create_with_files_actor.rb
```diff
@@ -46,10 +46,32 @@ def uploaded_files
...
+        receiving_users.each do |admin|
+          message = "#{user.name} created <a href='#{work_path}'>#{curation_concern.title.first}</a>"
```
mjgiarlo: Can we be sure users have a `#name` value? If so, :ok_hand:. If not, how about `user.email` instead?

mjgiarlo: Though the result returned will be different (joined titles rather than the first title), you could call `curation_concern.to_s` here instead of `curation_concern.title.first` and defer to the curation concern how it wishes to be coerced into a string.

app/actors/sufia/create_with_files_actor.rb
```diff
@@ -46,10 +46,32 @@ def uploaded_files
...
+        send_inactive_message
       end
 
       def inactive_uri
         ::RDF::URI('http://fedora.info/definitions/1/0/access/ObjState#inactive')
       end
+
+      def send_inactive_message
+        return true if receiving_users == []
+        receiving_users.each do |admin|
```
mjgiarlo: Admins are receivers at the moment, but the intent is to expand this to more types of users soon. Suggest renaming this to `recipient`.

app/actors/sufia/create_with_files_actor.rb
mjgiarlo: WIll the Sipity stuff in CC make this easier to ask what all users have a particular role?

app/actors/sufia/create_with_files_actor.rb
```diff
@@ -46,10 +46,32 @@ def uploaded_files
...
+      def receiving_users
+        receivers = []
+        ::User.find_each do |user|
+          receivers << user if (user.respond_to? :admin?) && user.admin?
+        end
+        receivers
+      end
+
+      def work_path
+        Rails.application.routes.url_helpers.curation_concerns_generic_work_path(curation_concern)
```
mjgiarlo: I suspect you want a polymorphic path here, since CC is not guaranteed to have a model named `GenericWork`.

hortongn: @mjgiarlo I'm using this path because when I tried polymorphic_path here with curation_concern it errored as an undefined method.  How can I use a polymorphic path here?

mjgiarlo: @hortongn The incantation is tricky to get right. How about something like `polymorphic_path([main_app, curation_concern])`? Tagging @jcoyne @cbeer @jeremyf.

hortongn: Nope, main_app is undefined here.  I've tried a few variations polymorphic_path with no luck.  @jcoyne @cbeer @jeremyf

jcoyne: `Rails.application.routes` is the same as `main_app`

hortongn: @jcoyne Thanks, but now it's back to complaining the polymorphic_path is undefined.  This is what I'm getting for `polymorphic_path(['Rails.application.routes', curation_concern])`
*** NoMethodError Exception: undefined method `polymorphic_path' for #<Sufia::CreateWithFilesActor:0x0000000d6b1a18>

curation_concern is
`#<GenericWork id: "j3860692c", head: [], tail: [], depositor: "user18@example.com", title: ["Test title"], date_uploaded: nil, date_modified: nil, state: #<ActiveTriples::Resource:0x6b515d4(#<ActiveTriples::Resource:0x0000000d6a2ba8>)>, owner: nil, label: nil, relative_path: nil, import_url: nil, part_of: [], resource_type: [], creator: [], contributor: [], description: [], keyword: [], rights: [], rights_statement: [], publisher: [], date_created: [], subject: [], language: [], identifier: [], based_near: [], related_url: [], bibliographic_citation: [], source: [], proxy_depositor: nil, on_behalf_of: nil, arkivo_checksum: nil, access_control_id: "38080b2d-642d-4d07-987e-0690c31d5d05", representative_id: nil, thumbnail_id: nil, admin_set_id: nil, embargo_id: nil, lease_id: nil>`

jcoyne: Elsewhere in sufia we do it by including both of these modules: https://github.com/projecthydra/sufia/blob/fa5467b94dceec8ab9c706a62da68409bece15f8/app/jobs/content_depositor_change_event_job.rb#L5-L6

app/actors/sufia/create_with_files_actor.rb
```diff
@@ -3,7 +3,7 @@ module Sufia
   class CreateWithFilesActor < CurationConcerns::Actors::AbstractActor
     def create(attributes)
       self.uploaded_file_ids = attributes.delete(:uploaded_files)
-      validate_files && mark_inactive && next_actor.create(attributes) && attach_files
+      validate_files && next_actor.create(attributes) && attach_files && mark_inactive
```
jcoyne: This should be a new actor, it should not go into this class. Add your actor to the stack here: https://github.com/projecthydra/sufia/blob/master/app/services/sufia/actor_factory.rb#L3

app/actors/sufia/create_with_files_actor.rb
```diff
@@ -46,10 +46,32 @@ def uploaded_files
       def mark_inactive
         return true unless Flipflop.enable_mediated_deposit?
         curation_concern.state = inactive_uri
+        send_inactive_message
```
jcoyne: Sending notifications in this class violates the Single Responsibility Principal, and it should be in the background because it's slow. Consider making a subclass of `ContentEventJob` for this (see https://github.com/projecthydra/sufia/blob/master/app/jobs/content_update_event_job.rb#L2 as an example)

spec/actors/sufia/create_with_files_actor_spec.rb
```diff
@@ -48,6 +48,10 @@
   describe "mediated deposit" do
     subject { actor.curation_concern.state }
     let(:inactive_uri) { RDF::URI('http://fedora.info/definitions/1/0/access/ObjState#inactive') }
+    let!(:admin) { create(:user) }
```
jcoyne: why do we need `let!` instead of `let` on these lines?

app/actors/sufia/create_with_files_actor.rb
hortongn: @jcoyne That works!  Including those modules fixes `polymorphic_path(curation_concern)` Thanks.  @mjgiarlo

PR #2744, opened 2016-09-27
Switch to do/end for multiline block

jcoyne:


awead:
awesome. Does that enforce braces for one-liners too?

jcoyne:
@awead I don't know. It's just closer to the default styles.

PR #2750, opened 2016-09-27
Non-admin users should not be able to see the admin page.

jcoyne:


app/controllers/sufia/admin/admin_sets_controller.rb
```diff
@@ -16,6 +17,9 @@ class Admin::AdminSetsController < ApplicationController
     self.list_search_builder_class = CurationConcerns::AdminSetSearchBuilder
 
     def show
+      # Even though the user can view this admin set, they may not be able to view
+      # it on the admin page.
```
cbeer: Could the same be true for the index action too (and, if so, should this get pulled up to a `before_action` callback?)

jcoyne: Sure, I've made the change.

PR #2753, opened 2016-09-28
Add derivatives_path to sufia initializer

hackmastera:
https://github.com/projecthydra/sufia/wiki/Sufia-Management-Guide#derivatives refers to `config.derivatives_path` in `config/initializers/curation_concerns.rb` but that initializer doesn't exist in a newly-generated sufia app.

### DIFF
--- a/lib/generators/sufia/templates/config/sufia.rb
+++ b/lib/generators/sufia/templates/config/sufia.rb
@@ -96,6 +96,10 @@ Sufia.config do |config|
   # This must be a lambda that returns a Pathname
   #  config.upload_path = ->() { Rails.root + 'tmp' + 'uploads' }
 
+  # Location on local file system where derivatives will be stored.
+  # If you use a multi-server architecture, this MUST be a shared volume.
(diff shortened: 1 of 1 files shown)


mjgiarlo:
@hackmastera @jcoyne I'm a bit confused, because `derivatives_path` is a CC concern, not a Sufia concern. And it should already be handled by this in CC: https://github.com/projecthydra/curation_concerns/blob/5f109a8e66d39c7a00cc323273174797a1aee2fd/lib/generators/curation_concerns/templates/config/curation_concerns.rb#L23-L25

I don't think the Sufia initializer should include this unless, e.g., we're trying to wrap up all CC configs in the Sufia initializer and shelter users from having to tweak multiple initializers.

jcoyne:
@mjgiarlo that's exactly it, if you run Sufia, you should never see or know about the curation_concerns initializer.

mjgiarlo:
Ah, yes. We discussed this in Slack and decided :arrow_up:. Remaining work has been ticketed in https://github.com/projecthydra/sufia/issues/2754. Thx, @jcoyne and @hackmastera and sorry for the confusion.

PR #2766, opened 2016-10-05
[WIP] Add the ability to switch to a different language

jcoyne:
screen shot 2016-10-05 at 10 18 31 am
Click the Spanish flag to switch to using the Spanish locale.

This requires someone to translate `sufia.es.yml` before we merge, otherwise you get an error when switching to Spanish.

### DIFF
--- a/app/controllers/concerns/sufia/controller.rb
+++ b/app/controllers/concerns/sufia/controller.rb
@@ -4,6 +4,8 @@ module Sufia::Controller
   included do
     # Adds Hydra behaviors into the application controller
     include Hydra::Controller::ControllerBehavior
+
+    before_action :set_locale
   end
 
   def current_ability
@@ -14,4 +16,15 @@ module Sufia::Controller
   def after_sign_in_path_for(_resource)
     sufia.dashboard_index_path
   end
+
+  # Ensure that the locale choice is persistent across requests
+  def default_url_options
+    { locale: I18n.locale }
+  end
+
+  private
+
+    def set_locale
+      I18n.locale = params[:locale] || I18n.default_locale
+    end
 end
--- a/app/helpers/sufia/sufia_helper_behavior.rb
+++ b/app/helpers/sufia/sufia_helper_behavior.rb
@@ -16,6 +16,12 @@ module Sufia
       t('sufia.institution_name_full', default: institution_name)
     end
 
+    # Which translations are available for the user to select
+    # @return [Hash<String,String>] locale abbreviations as keys and flags as values
+    def available_translations
+      { 'en' => '🇬🇧', 'es' => '🇪🇸' }
+    end
+
     def orcid_label(style_class = '')
       "#{image_tag 'orcid.png', alt: t('sufia.user_profile.orcid.alt'), class: style_class} #{t('sufia.user_profile.orcid.label')}".html_safe
     end
--- a/app/views/_user_util_links.html.erb
+++ b/app/views/_user_util_links.html.erb
@@ -1,4 +1,13 @@
 <ul id="user_utility_links" class="nav navbar-nav navbar-right">
+  <li>
+    <% available_translations.each do |lang, flag| %>
+      <% next if lang == I18n.locale.to_s %>
+      <span class="sr-only">Switch language:</span>
+      <%= link_to "?locale=#{lang}" do %>
+        <span class="locale-choice locale-choice-<%= lang %>"><%= flag %></span>
+      <% end %>
+    <% end %>
+  </li>
   <% if user_signed_in? %>
     <li>
       <%= link_to sufia.notifications_path do %>
--- a/app/controllers/sufia/resource_sync_controller.rb
+++ b/app/controllers/sufia/resource_sync_controller.rb
@@ -1,4 +1,9 @@
 class Sufia::ResourceSyncController < ApplicationController
+  # We don't need locale in here.
+  def default_url_options
+    super.except(:locale)
+  end
+
   def source_description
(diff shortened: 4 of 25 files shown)


jcoyne:
@hectorcorrea Would you be able to take a shot at filling out `sufia.es.yml`?

hectorcorrea:
@jcoyne sure think, I'll do a first pass early next week.

awead:
We're having to pass `locale: 'en'` everywhere in our routes. Is there way we can just use the default?

jcoyne:
@awead That is the way that is recommended in the rails guides.  See the warning here: http://guides.rubyonrails.org/i18n.html#storing-the-locale-from-the-session-or-cookies

botimer:
I don't mean to be contrarian here, but _forcing_ locales on URLs or routes to deal with locales seems pretty gnarly. In my travels in internationalized apps where people spend significant time, the general case user expectation is literally opposite to the "click this link I've shared with you and you will get it in my language" behavior in section 2.2.5.

I acknowledge the purity of it, and where a locale is included in a URL, it should be respected (as an override to 1. user/session preferences, 2. the Accept-Language header). However, I don't buy it that having the default behavior being to include locale and encoding stuff in the URL is what users typically want. In the Wikipedia case, sure -- the content itself is localized and `en.` is explicitly different content than `es.` -- but this seems more about the app locale. Content or asset variants are another concern.

Stated more simply: I think most people want to click a link and see the app localized for them, not the sender.

I actually don't think this position is at odds with section 2.2 in general. Specifically 2.2.3 and 2.2.4 deal with authenticated users, language headers, and a deployment-wide default.

Thoughts?

mjgiarlo:
@botimer how about if this PR adds the ability to fall back to 2.2.4?

botimer:
Hey @mjgiarlo, I think it's great for the URL to be respected. I just saw @awead's concern that this gets a lot of treatment throughout the code and I read between the lines that URLs would always have them. I may have read too much. Either way, I do think a tiered fallback is right 👍  (URL, header, user pref, system).

A general principle to me is that it should not be a lot harder to develop because the app is internationalized. It's already enough mental overhead to deal with views that have little to none of the actual text. Having to be explicit about locale support in all the routes and link helpers -- OY!

I do acknowledge the trickiness of the "session locale" notion... Should visiting a page with an explicit locale carry over to any links you would click? Or should those come up in *your* preferred locale? What do real users want here? I'm also curious what dev/demo needs would be unmet by an "easy locale switcher" for real users. Maybe a predominantly multi-lingual partner could advise -- I'm sure they run into good and rough examples on the web.

app/views/_user_util_links.html.erb
```diff
@@ -1,4 +1,13 @@
 <ul id="user_utility_links" class="nav navbar-nav navbar-right">
+  <li>
+    <% available_translations.each do |lang, flag| %>
+      <% next if lang == I18n.locale.to_s %>
+      <span class="sr-only">Switch language:</span>
```
cbeer: This ought to be i18n'ed

app/helpers/sufia/sufia_helper_behavior.rb
```diff
@@ -16,6 +16,12 @@ def institution_name_full
       t('sufia.institution_name_full', default: institution_name)
     end
 
+    # Which translations are available for the user to select
+    # @return [Hash<String,String>] locale abbreviations as keys and flags as values
+    def available_translations
+      { 'en' => '🇬🇧', 'es' => '🇪🇸' }
```
cbeer: I wonder if it'd be easier to customize this if it was part of the `Sufia::Engine.config`?

app/controllers/concerns/sufia/controller.rb
```diff
@@ -14,4 +16,15 @@ def current_ability
   def after_sign_in_path_for(_resource)
     sufia.dashboard_index_path
   end
+
+  # Ensure that the locale choice is persistent across requests
+  def default_url_options
+    { locale: I18n.locale }
```
cbeer: Does this need to merge with `super` in case another module is trying to inject its own behavior?

app/helpers/sufia/sufia_helper_behavior.rb
cbeer: I'm also not a particular fan of `flags = language`, but my only suggestion would be to use the full language name.

mjgiarlo: I like the way the WDL site approaches this: https://www.wdl.org/en/item/1/# (See the flag dropdown.)

mjgiarlo:
@botimer OK. I would also be inclined to punt on user preference for now, since that's not a thing in Sufia and that can come later.

We've talked with @ejrivero about his Spanish translation but have yet to identify any other potential partners with which to do this I18n work. That would be helpful. In the meantime, might this PR as is, with the ability to use HTTP leaders as a fallback mechanism, be a good-enough advancement of our community's i18n support until we have more resources and experience?

botimer:
@mjgiarlo sure, I'm the last to stand in the way of good work, especially on something that naturally has to be incremental or iterative. I just wanted to raise a quick flag in case this was a proverbial fork in the road (a new feature vs. a modification sprinkled throughout vs. a constraint or concession).

mjgiarlo:
@botimer Oh, I didn't think you were being obstructionist or anything like that. Thanks for raising the flag!

mjgiarlo:
Replaced with https://github.com/projecthydra-labs/hyrax/pull/33. Thanks, @jcoyne!

PR #2767, opened 2016-10-05
Adding FileSet, Permission and Version builders

### LINKED ISSUE #2144: Migration::Import::FileSetBuilder
I would like to create an class that creates a Sufia PCDM Work based on Sufia 6.0 exported Metadata hash. See #2141 for calling service.

* Build - builds a FileSet based on the metadata passed in and attaches the content and versions history
  * parameter: options # metadata about the FileSet to be created

This is done when I can:

Migration::Import::FileSetBuilder.build produces a FileSet with the same descriptive Metadata as is passed in via options and has attached content and versions as specified in the options hash.

An example of an FileSet importer can be found in Sufia [import_s7](https://github.com/projecthydra/sufia/blob/import_s6/lib/sufia/import/import_service.rb#L35-L61)

cam156:
Fixes #2144 

Adds in FileSetImporter along with permission and version importers.

@hackmastera suggested maybe this should go into a separate gem so it does not need to be versioned along with sufia.  What do people think?  This code could really be anywhere with the exception that it depends on the Objects in CC & Sufia.

### DIFF
--- a/lib/sufia.rb
+++ b/lib/sufia.rb
@@ -19,6 +19,7 @@ require 'curation_concerns'
 require 'sufia/engine'
 require 'sufia/version'
 require 'sufia/inflections'
+require 'sufia/import'
 require 'kaminari_route_prefix'
 
 module Sufia
--- a/lib/sufia/import.rb
+++ b/lib/sufia/import.rb
@@ -1,0 +1,10 @@
+require 'sufia/import/builder'
+require 'sufia/import/permission_builder'
+require 'sufia/import/version_builder'
+require 'sufia/import/file_set_builder'
+
+module Sufia
+  module Permissions
+    VERSION = Sufia::VERSION
+  end
+end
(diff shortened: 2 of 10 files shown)


awead:
I think it needs to remain in Sufia. It's my understanding that that import features in Sufia 7 would take a set of exported data from Sufia 6 and migrate it into your new application. To me, that says that this chunk of code is definitely meant for version 7. Putting it a separate gem would get confusing and might make dependency management difficult.

When Sufia 8 comes along, we can revisit this and either remove it or update it to do any 7 to 8 migrations, if necessary.

cam156:
@awead and I discussed the gem off ticket.   The new plan would be to branch the migration work off of Sufia 7.2.  Finish that work in a Feature branch that can later be rebased and merged into master with all the new features that have been added. That would mean one larger PR later that includes all the work for Migration.  @hackmastera @jcoyne @mjgiarlo how does that sound?

mjgiarlo:
@cam156 I'm :cool: with that. Thank you for working on this!

hackmastera:
sounds like a good plan.

Issue #2143, opened 2016-06-06
Migration::Import::WorkBuilder


cam156:
I would like to create an class that creates a Sufia PCDM Work based on Sufia 6.0 exported Metadata hash. See #2141 for calling service.

This class should have the following methods:
* Build - builds a GenericWork based on the metadata passed in
  * parameter: options # metadata about the Work to be created

This is done when I can:

Migration::Import::WorkBuilder.build produces a GenericWork with the same descriptive Metadata as is passed in via options

An example of an Work importer can be found in Sufia [import_s7](https://github.com/projecthydra/sufia/blob/import_s6/lib/sufia/import/import_service.rb#L110)

PR #2777, opened 2016-10-11
Add work builder to sufia import, fixes projecthydra/sufia#2143

hackmastera:
@cam156

### DIFF
--- a/lib/sufia/import/work_builder.rb
+++ b/lib/sufia/import/work_builder.rb
@@ -1,0 +1,57 @@
+# Builder for generating a work incluing permissions
+#
+module Sufia::Import
+  class WorkBuilder < Builder
+    attr_reader :work, :permission_builder
+
+    # @param settings see Sufia::Import::Builder for settings
+    def initialize(settings)
+      super
+      @work = Sufia.primary_work_type.new
+      @permission_builder = PermissionBuilder.new(settings, work)
+    end
+
+    # Build a Work from GenericFile metadata
+    #
+    # @param OpenStruct data metadata from the generic_file in OpenStruct format
+    #    <OpenStruct id="44558d49x", label="my label", depositor="cam156@psu.edu", arkivo_checksum="arkivo checksum",
+    #                relative_path="relative path", import_url="import url", resource_type=["resource type"],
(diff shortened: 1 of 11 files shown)


lib/sufia/import/work_builder.rb
```diff
@@ -0,0 +1,57 @@
...
+    #                title=["My Great File"], creator=["cam156@psu.edu"], contributor=["contributor1", "contribnutor2"],
+    #                description=["description of the file"], tag=["tag1", "tag2"], rights=["Attribution 3.0"],
+    #                publisher=["publisher joe"], date_created=["a long time ago"], date_uploaded="2015-09-28T20:00:14.243+00:00",
+    #                date_modified="2015-10-28T20:00:14.243+00:00", subject=["subject 1", "subject 2"], language=["WA Language WA"],
+    #                identifier=["You ID ME"], based_near=["Kalamazoo"], related_url=["abc123.org"], bibliographic_citation=["cite me"],
+    #                source=["source of me"], batch_id="qn59q409q", visibility="restricted",
+    #                permissions=[#<OpenStruct id="b5911dfd-07b1-43ab-b11d-1bc0534d874c", agent="http://projecthydra.org/ns/auth/person#cam156@psu.edu", mode="http://www.w3.org/ns/auth/acl#Write", access_to="44558d49x">...]>
+    def build(data)
+      work.id = data.id
+      work.title << data.title
```
cam156: @hackmastera I wonder if all these direct assigns could be done with an update attributes call.  Have you tried it?  Not sure it would work, but it would clean up the code. This is not a blocker though...

lib/sufia/import/work_builder.rb
```diff
@@ -0,0 +1,57 @@
...
+      work.label = data.label
+      work.date_uploaded = data.date_uploaded
+      work.date_modified = data.date_modified
+      work.apply_depositor_metadata(data.depositor)
+      permission_builder.build(data.permissions)
```
cam156: @hackmastera I might move this to the bottom so it is more obvious it is doing more than assigning a value.  It's a bit hidden.

hackmastera:
@cam156 I refactored to use hashes as we discussed. I also changed the work builder to use update_attributes.

PR #2782, opened 2016-10-12
Add install template for quicker installation

mjgiarlo:
Using a template, and updating it within every release, allows us to simplify the instructions for developers who wish to get Sufia up and running in development mode very quickly. Also helps automate manual steps such as the inclusion of the `flipflop` gem from GitHub.

README and release management process have already been updated to reflect the new template.

PR #2783, opened 2016-10-12
Update README with simpler installation procedure

mjgiarlo:
Goes with #2782

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,15 +27,10 @@ Jump in: [![Slack Status](http://slack.projecthydra.org/badge.svg)](http://slack
     * [Environments](#environments)
     * [Ruby](#ruby)
   * [Creating a Sufia\-based app](#creating-a-sufia-based-app)
+    * [Redis](#redis)
     * [Rails](#rails)
-    * [Sufia's Ruby\-related dependencies](#sufias-ruby-related-dependencies)
-    * [Install Sufia](#install-sufia)
     * [Generate a primary work type](#generate-a-primary-work-type)
-    * [Database tables and indexes](#database-tables-and-indexes)
-    * [Start Redis](#start-redis)
-    * [Start Solr](#start-solr)
-    * [Start FCRepo](#start-fcrepo)
-    * [Spin up the web server](#spin-up-the-web-server)
+    * [Start servers](#start-servers)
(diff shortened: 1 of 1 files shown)


barmintor:
assigning myself while we get @cgalarza set up as a contributor

README.md
```diff
@@ -154,32 +149,28 @@ We recommend either Ruby 2.3 or the latest 2.2 version.
...
+## Rails
 
-Add the following lines to your application's Gemfile.
+Generate a new Rails application. We recommend the latest Rails 5.0 or 4.2 release.
 
 ```
-gem 'sufia', '7.2.0'
+# If you don't already have Rails at your disposal...
+gem install rails -v 5.0.0.1
+rails new my_app -m https://raw.githubusercontent.com/projecthydra/sufia/master/template.rb
```
cgalarza: Is there anyway to tie this url to the particular commit of this README? If users wanted to install a version of sufia that wasn't the version in `master` they would run into some trouble.

barmintor: that's a good question. I don't know if there's a latest release tag shortcut, which would be ideal.

barmintor: you can link to the latest release and be redirected to a tag, but I don't see a way for us to link to the tag as a tree.

mjgiarlo: @barmintor @cgalarza I dig the idea, but I can't think of a good way to handle this. If I point at, say, the 7.2.0 release, the README will be broken because `template.rb` does not exist in the 7.2.0 release. I'm not sure I want to cut a special release just to fix up the README, though I can be convinced otherwise. Pointing at a release branch could work, but we tend to create those at point of need (when someone wants something backported). :thinking:

What if we leave it pointing at `master` now, and use our release management process to edit this line when we cut our next release? I have already edited the release mgmt notes to reflect that cutting a new release should at least require reviewing, if not modifying, the README and `template.rb`.

mjgiarlo: Wrote this up as an issue: https://github.com/projecthydra/sufia/issues/2784

cgalarza: @mjgiarlo Editing the template line when a new release is cut sounds good to me.

mjgiarlo: @cgalarza Thanks! I set it as a blocker for 7.3.0.

Issue #2141, opened 2016-06-06
Migration::Import::GenericFileTranslator


cam156:
I would like to create an class that imports (and translates) a Sufia 6.0 exported Metadata file into a Sufia PCDM GernericWork and FileSet. See #2137 for executable.

This class should have the following methods:
* import - loops through the files found in the file_pattern and maps the GenericFile to a GenericWork and FileSet
  * parameter: file_pattern # an array of ids to process

This is done when I can:

Migration::Import::GenericFileTranslator.import produces a GenericWork and a FileSet for each file in the directory specified by the input pattern, and that each GenericWork contains the same descriptive Metadata as the GenericFile and has a FileSet with the same content as the GenericFile.

An example of an GenericFile Convertor can be found in Sufia [import_s7](https://github.com/projecthydra/sufia/blob/import_s6/lib/sufia/import/import_service.rb#L148)

mjgiarlo:
@cam156 I haven't read these tickets in depth, but are you thinking that the import process will record the GF's ID and make sure the new GW also uses the same ID?

cam156:
@mjgiarlo Short answer is yes. @hectorcorrea created some code to do the import export. I the export code exists in import_s6 branch.  It is old and a bit stale, but can be reused as a reference at the very least.

mjgiarlo:
:ok_hand: I thought it might also be good to document that as a requirement in the related tickets, but if y'all already have this one accounted for, np. Thanks!

cam156:
@mjgiarlo linked the code in the original descripotion. Not that we have to stick with what the code does, we can push the linking down if we believe that would be a better location...

mjgiarlo:
I'm a big :+1: on providing continuity by using GF IDs for the new GWs.

hackmastera:
> parameter: file_pattern # an array of ids to process

Not sure it's practical to loop through an array of ids, and I suspect that language is copied from the exporter design. I'd prefer to accept a directory containing files with names as output by the exporter (generic_file_[id].json). Only files in the given directory with this filename format will be processed.

objections?

PR #2786, opened 2016-10-13
Add Sufia::Import::GenericFileTranslator

hackmastera:
@cam156 or others; anything else you would like to see tested? Other comments so far?

### DIFF
--- a/lib/sufia/import/generic_file_translator.rb
+++ b/lib/sufia/import/generic_file_translator.rb
@@ -1,0 +1,43 @@
+module Sufia::Import
+  # Imports a Sufia 6.0-exported GenericFile into a Sufia PCDM GernericWork and FileSet
+  class GenericFileTranslator
+    # @param Hash settings needed to pass through to builder
+    #        @attr sufia6_user      - User name for accessing the sufia 6 fedora
+    #        @attr sufia6_password  - Password for accessing the sufia 6 fedora
+    #        @attr import_binary    - (default true) Import the binary content from fedora
+    def initialize(settings)
+      # TODO: work builder doesn't need settings; just file builder
+      @work_builder = WorkBuilder.new(settings)
+      @file_set_builder = FileSetBuilder.new(settings)
+    end
+
+    # loop through the files found in import_dir and map each GenericFile to a Sufia.primary_work and FileSet
+    # @param import_dir a directory containing export files, e.g.: generic_file_[id].json
+    # @return list of imported files
+    def import(import_dir, filename_prefix)
+      filename_prefix ||= "generic_file_"
+      raise "No such directory: '#{import_dir}'" unless Dir.exist?(import_dir)
+      # get filenames
+      files = Dir.glob(File.join(import_dir, '*')).select { |f| exported_gf?(f, filename_prefix) }
+      files.each { |f| import_file(f) }
+    end
+
+    private
+
+      def import_file(filename)
+        Rails.logger.debug "Importing #{File.basename(filename)}"
+        file_str = File.read(filename)
+        json = JSON.parse(file_str)
+        work = @work_builder.build(json)
+        fileset = @file_set_builder.build(json)
+        work.ordered_members << fileset
+        work.save
+      end
+
+      def exported_gf?(f, filename_prefix)
+        File.file?(f) &&
+          File.extname(f) == '.json' &&
+          File.basename(f).start_with?(filename_prefix)
+      end
+  end
+end
--- a/spec/lib/sufia/import/generic_file_translator_spec.rb
+++ b/spec/lib/sufia/import/generic_file_translator_spec.rb
@@ -1,0 +1,37 @@
+require 'spec_helper'
+
+describe Sufia::Import::GenericFileTranslator do
+  let(:sufia6_user) { "s6user" }
+  let(:sufia6_password) { "s6password" }
+  let(:translator) { described_class.new(sufia6_user: sufia6_user, sufia6_password: sufia6_password) }
+
+  let(:import_directory) { File.join(fixture_path, 'import') }
+  let(:gf_metadata) { JSON.parse(File.read(File.join(import_directory, "generic_file_#{work_id}.json")), symbolize_names: true) }
+  let(:work) { Sufia.primary_work_type.find(work_id) }
+  # used to retrieve the fixture and then test that the object was created with the same id
+  let(:work_id) { 'qr46r0963' }
+
+  let(:permission_builder) { instance_double(Sufia::Import::PermissionBuilder) }
+  let(:version_builder) { instance_double(Sufia::Import::VersionBuilder) }
+  before do
+    allow(Sufia::Import::PermissionBuilder).to receive(:new).and_return(permission_builder)
+    allow(Sufia::Import::VersionBuilder).to receive(:new).and_return(version_builder)
+    allow(permission_builder).to receive(:build).and_return([])
+    allow(version_builder).to receive(:build).and_return([])
+  end
+
+  describe '#import' do
+    it 'Creates a work with attached fileset' do
+      expect(Rails.logger).to receive(:debug).with("Importing generic_file_qr46r0963.json")
+      translator.import(import_directory, 'generic_file_')
+      expect(Sufia.primary_work_type.count).to eq 1
+      expect(FileSet.count).to eq 1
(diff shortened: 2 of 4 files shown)


lib/sufia/import/generic_file_translator.rb
```diff
@@ -0,0 +1,42 @@
...
+
+    # loop through the files found in import_dir and map each GenericFile to a Sufia.primary_work and FileSet
+    # @param import_dir a directory containing export files, e.g.: generic_file_[id].json
+    # @return list of imported files
+    def import(import_dir)
+      # TODO: error instead of return
+      return unless Dir.exist?(import_dir)
+      # turn import_dir into a list of actual filenames
+      files = Dir.glob(File.join(import_dir, '*')).select { |f| exported_gf?(f) }
+      files.each { |f| import_file(f) }
```
cam156: @hackmastera What are you planning to do with the files?  I would want the files or Works that were created, not necessarily the json file, but I am still not sure what the plan is for them.

lib/sufia/import/generic_file_translator.rb
```diff
@@ -0,0 +1,42 @@
...
+        json = JSON.parse(file_str)
+        work = @work_builder.build(json)
+        fileset = @file_set_builder.build(json)
+        work.ordered_members << fileset
+      end
+
+      def exported_gf?(f)
+        File.file?(f) &&
+          File.extname(f) == '.json' &&
+          File.basename(f) =~ /^generic_file.*/
```
cam156: We may want this to be customizable.  Not sure it is needed, since that is exactly the string that is created by https://github.com/projecthydra/sufia/blob/6.x-stable/lib/sufia/export/actor.rb#L81

I like the validation!

spec/lib/sufia/import/generic_file_translator_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+  let(:sufia6_password) { "s6password" }
+  let(:translator) { described_class.new(sufia6_user: sufia6_user, sufia6_password: sufia6_password) }
+
+  let(:import_directory) { File.join(fixture_path, 'import') }
+  let(:gf_metadata) { JSON.parse(File.read(File.join(import_directory, 'generic_file_qr46r0963.json')), symbolize_names: true) }
+  let(:work_id) { 'qr46r0963' }
+  let(:work) { Sufia.primary_work_type.find(work_id) }
+
+  let(:permission_builder) { instance_double(Sufia::Import::PermissionBuilder) }
+  let(:version_builder) { instance_double(Sufia::Import::PermissionBuilder) }
```
cam156: This should be a VersionBuilder

spec/lib/sufia/import/generic_file_translator_spec.rb
```diff
@@ -0,0 +1,31 @@
+require 'spec_helper'
+
+describe Sufia::Import::GenericFileTranslator do
+  let(:sufia6_user) { "s6user" }
+  let(:sufia6_password) { "s6password" }
+  let(:translator) { described_class.new(sufia6_user: sufia6_user, sufia6_password: sufia6_password) }
+
+  let(:import_directory) { File.join(fixture_path, 'import') }
+  let(:gf_metadata) { JSON.parse(File.read(File.join(import_directory, 'generic_file_qr46r0963.json')), symbolize_names: true) }
+  let(:work_id) { 'qr46r0963' }
```
cam156: This is slightly mysterious that is matches with the id in the fixture. May want a comment.

lib/sufia/import/generic_file_translator.rb
hackmastera: You mean as a return value? I guess I saw this as primarily being used by the exe, so maybe just to report that they were imported.

spec/lib/sufia/import/generic_file_translator_spec.rb
hackmastera: Ha, thanks

lib/sufia/import/generic_file_translator.rb
hackmastera: Or if you are asking about line 22, I'm not sure I understand the question.

lib/sufia/import/generic_file_translator.rb
hackmastera: I could move the regex to a constant as a first step

lib/sufia/import/generic_file_translator.rb
hackmastera: Okay I think I see where we're not seeing eye-to-eye; see my comment from a couple days ago: https://github.com/projecthydra/sufia/issues/2141#issuecomment-253327681

I am now thinking it sounds like some of the work I've done here should move to the exe, which will pull the file list and pass just that to the translator. will rearrange.

hackmastera:
@cam156 I messed with pulling logic out of here and into the exe, but it was less testable and I felt like I was accommodating a use case that no one had told me existed. So I ended up just beefing up the functionality a tiny bit here (adding the filename prefix to .import and adding a verbosity flag, which may not be standard really but I figured it's better than printing to a log considering how people will use this class). Let me know what you think! Related: the exe is in pr #2793

lib/sufia/import/generic_file_translator.rb
```diff
@@ -0,0 +1,45 @@
...
+      raise "No such directory: '#{import_dir}'" unless Dir.exist?(import_dir)
+      # get filenames
+      files = Dir.glob(File.join(import_dir, '*')).select { |f| exported_gf?(f, filename_prefix) }
+      files.each { |f| import_file(f) }
+    end
+
+    private
+
+      def import_file(filename)
+        puts "Importing #{File.basename(filename)}" if @verbose
```
cam156: @hackmastera Since we are in the context of a Rails app possibly using Rails.logger.debug might be better here.  That may help with the failing test.

spec/lib/sufia/import/generic_file_translator_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+    it 'Errors when passed a nonexistent directory' do
+      expect { translator.import('totallynotadirectory', 'whatever') }.to raise_error RuntimeError
+    end
+
+    describe 'in verbose mode' do
+      before do
+        translator.verbose = true
+      end
+      it 'Prints info' do
+        expect(STDOUT).to receive(:puts).with("Importing generic_file_qr46r0963.json")
```
cam156: @hackmastera This would change to  `expect(Rails.logger).to receive(:debug)`

lib/sufia/import/generic_file_translator.rb
hackmastera: Thanks for looking at it. You are probably right. I was hesitant, though, because the class is primarily intended to be used by the exe. maybe there is a way for the exe to print from the rails log if it's in verbose mode? then i can remove the 'verbose' stuff from here and just write to the log straight-up.

hackmastera:
@cam156 Okay I think this one is ready.

spec/lib/sufia/import/generic_file_translator_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+      expect(work.members.one?).to be true
+      expect(work.title).to include 'Myth'
+    end
+
+    it 'Errors when passed a nonexistent directory' do
+      expect { translator.import('totallynotadirectory', 'whatever') }.to raise_error RuntimeError
+    end
+
+    it 'logs import events' do
+      expect(Rails.logger).to receive(:debug).with("Importing generic_file_qr46r0963.json")
```
cam156: @hackmastera given this will happen each time, could this just be added after line 26?  That would be slightly more efficient and make the tests faster in the long run.

spec/lib/sufia/import/generic_file_translator_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+  end
+
+  describe '#import' do
+    it 'Creates a work with attached fileset' do
+      expect(Sufia.primary_work_type.count).to eq 0
+      expect(FileSet.count).to eq 0
+      translator.import(import_directory, 'generic_file_')
+      expect(Sufia.primary_work_type.count).to eq 1
+      expect(FileSet.count).to eq 1
+      expect(work.members.one?).to be true
```
cam156: @hackmastera in the code you are assigning to ordered members.  Should this test that is set?

PR #2793, opened 2016-10-17
Add executable to run imports from sufia 6

hackmastera:
@cam156

### DIFF
--- a/exe/sufia_import
+++ b/exe/sufia_import
@@ -1,0 +1,62 @@
+#!/usr/bin/env ruby
+require 'rails'
+require ::File.join(File.dirname(Bundler.default_gemfile), 'config/environment')
+
+options = OpenStruct.new
+options.verbose = false
+options.import_dir = "tmp/import"
+# TODO: add collection like options.mappings = ["Sufia::Import::CollectionTranslator=collection_", "Sufia::Import::GenericFileTranslator=generic_file_"]
+options.mappings = ["Sufia::Import::GenericFileTranslator=generic_file_"]
+options.import_binary = true
+options.user = 's6user'
+options.password = 's6password'
+
+args = OptionParser.new do |opts|
+  opts.banner = "imports model metdata json files in, contacting fedora to import binaries. Usage: sufia_import [options]"
+
+  opts.on("-v", "--[no-]verbose", "Run verbosely") do |v|
+    options.verbose = v
+  end
+
+  opts.on("-d", "--json_directory [path/to/files]", "Directory containing files for import.  Defaults to tmp/import") do |dir|
+    options.import_dir = dir
+  end
+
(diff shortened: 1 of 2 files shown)


exe/sufia_import
```diff
@@ -0,0 +1,43 @@
...
+  $stderr.puts args.help
+  exit 1
+end
+
+verbose = options[:verbose]
+$stdout.puts "running with #{options}" if verbose
+
+options.mappings.each do |pair|
+  class_name, prefix = pair.split('=')
+  translator = "Sufia::Import::#{class_name}Translator".constantize.new(sufia6_user: 'username', sufia6_password: 'password', import_binary: 'false')
```
cam156: I'm wondering how extensible this is.  I know it matches the ticket, but if I wanted to add my own ticket.  How about if we change it to just the translator class and the file pattern?  
`Sufia::Import::GenericFileTranslator=generic_file_,Sufia::Import::CollectionTranslator=collection_`
That way other people could namespace and name the translators as they please...

exe/sufia_import
```diff
@@ -0,0 +1,43 @@
...
+options.mappings = ["GenericFile=generic_file_"]
+
+args = OptionParser.new do |opts|
+  opts.banner = "imports model metdata json files in, contacting fedora to import binaries. Usage: sufia_import [options]"
+
+  opts.on("-v", "--[no-]verbose", "Run verbosely") do |v|
+    options.verbose = v
+  end
+
+  opts.on("-d", "--json_directory [path/to/files]", "Directory containing files for import.  Defaults to tmp/import") do |dir|
```
hackmastera: I assumed it would be moved, I guess because I am setting up a new box. I can change it if you have strong feelings, but considering people will probably have all kinds of different approaches, I like that at least the semantics of the location match those of the script.

exe/sufia_import
hackmastera: yes; will do this.

exe/sufia_import
```diff
@@ -0,0 +1,43 @@
...
+  $stderr.puts args.help
+  exit 1
+end
+
+verbose = options[:verbose]
+$stdout.puts "running with #{options}" if verbose
+
+options.mappings.each do |pair|
+  class_name, prefix = pair.split('=')
+  translator = "#{class_name}Translator".constantize.new(sufia6_user: 'username', sufia6_password: 'password', import_binary: 'false')
```
hackmastera: Geez, thank you for catching this. on it.

hackmastera:
@cam156 just a note that we probably shouldn't merge this until #2786 is ready. I'm 100% stuck on that; the test that is failing in travis is succeeding for me locally. Yesterday on travis it was succeeding with rails 5 but failing with rails 4. today in one run that switched and in the other it's failing on both.

hackmastera:
@cam156 I think this is ready.

PR #2795, opened 2016-10-18
handeling either has_model_ssim or active_fedora_model_ssi

cam156:
Since there is a chance that user will have a newer version of ActiveFedora that includes the has_model_ssim, or something might be a bit funky with thier repository lets handle records with no model and both of the possible keys.

### DIFF
--- a/lib/sufia/migration/survey/fedora_id_service.rb
+++ b/lib/sufia/migration/survey/fedora_id_service.rb
@@ -49,7 +49,12 @@ module Sufia
             query = 'id:"' + id + '"'
             matches = ActiveFedora::SolrService.query(query)
             return nil if matches.count == 0
-            model_str = matches.first["active_fedora_model_ssi"]
+            model_str = matches.first["has_model_ssim"]
+            model_str = model_str.first if model_str.is_a?(Array)
+            if model_str.blank? || !Object.const_defined?(model_str)
+              Rails.logger.error("Invalid model #{id} #{model_str}")
+              return nil
+            end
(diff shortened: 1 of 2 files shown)


lib/sufia/migration/survey/fedora_id_service.rb
```diff
@@ -49,7 +49,12 @@ def active_fedora_model(id)
             query = 'id:"' + id + '"'
             matches = ActiveFedora::SolrService.query(query)
             return nil if matches.count == 0
-            model_str = matches.first["active_fedora_model_ssi"]
+            model_str = matches.first["active_fedora_model_ssi"] || matches.first["has_model_ssim"]
```
jcoyne: All version of ActiveFedora (as far back as I can remember) have `has_model_ssim`. The new versions do not use `active_fedora_model_ssi`. I'd prefer that we just change to `has_model_ssim` as it should be completely compatible.

lib/sufia/migration/survey/fedora_id_service.rb
```diff
@@ -49,7 +49,12 @@ def active_fedora_model(id)
...
+            model_str = model_str.first if model_str.kind_of?(Array)
+            if (model_str.blank? || !Object.const_defined?(model_str))
+              Rails.logger.warn("Invalid model #{id} #{model_str}")
```
jcoyne: I'd make this more than a warning.  How about `error`?

cam156:
@jcoyne Do the updates address your concerns?

PR #2796, opened 2016-10-18
Fix import_binary so it can be set to false

hackmastera:
@cam156


lib/sufia/import/builder.rb
```diff
@@ -11,8 +11,7 @@ class Builder
       #        @attr import_binary    - (default true) Import the binaryt content from fedora
       def initialize(settings)
         @settings = settings
-        @import_binary = settings[:import_binary]
-        @import_binary ||= true
+        @import_binary = settings[:import_binary].nil? ? true : settings[:import_binary]
```
jcoyne: A more idiomatic way to write this is `settings.fetch(:import_binary, true)`

PR #2797, opened 2016-10-18
Don't give the depositor edit access if mediated deposit is configured.

jcoyne:
Extend the new actor in the curation_concerns branch.

### DIFF
--- a/app/actors/sufia/grant_edit_to_depositor_actor.rb
+++ b/app/actors/sufia/grant_edit_to_depositor_actor.rb
@@ -1,0 +1,15 @@
+module Sufia
+  class GrantEditToDepositorActor < CurationConcerns::Actors::GrantEditToDepositorActor
+    private
+
+      # Overriding the parent class so it doesn't grant access
+      # if mediated deposit is enabled
+      def grant_edit_access
+        super unless mediated_deposit?
+      end
+
+      def mediated_deposit?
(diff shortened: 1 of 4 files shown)


app/actors/sufia/grant_edit_to_depositor_actor.rb
```diff
@@ -0,0 +1,15 @@
+module Sufia
+  class GrantEditToDepositorActor < CurationConcerns::Actors::GrantEditToDepositorActor
+    private
+
+      # Overriding the parent class so it doesn't grant access
+      # if mediated deposit is enabled
+      def grant_edit_access
+        super unless mediated_deposit?
```
mjgiarlo: So if `mediated_deposit?` returns `false` will that halt the stack because `#grant_edit_access` returns `nil`?

jcoyne: No. The return value of this method is inconsequential https://github.com/projecthydra/curation_concerns/commit/7169243c5d201514dc6204e54369d16c2545b452#diff-e93e2181c149e5670c6afc7cf3f216c7R8

mjgiarlo: Because it's not one of the magick `#update` or `#create` methods that handle chaining? :ok_hand:

PR #2798, opened 2016-10-18
Translate "Last Modified"

jcoyne:


app/views/curation_concerns/base/show.html.erb
```diff
@@ -12,7 +12,7 @@
   </div>
   <div class="col-sm-4">
     <%= render "show_actions", presenter: @presenter %>
-    Last modified: <%= @presenter.date_modified %>
+    <%= t('.last_modified') %> <%= @presenter.date_modified %>
```
cbeer: I think `.last_modified` needs to accept the `@presenter.date_modified`, e.g. `t('.last_modified', value: @presenter.date_modified)`. In RTL languages, presumably you'd want to swap the order of the label and value.

PR #2799, opened 2016-10-18
Don't allow mediated deposit to grant edit access

jcoyne:


spec/forms/sufia/forms/work_form_spec.rb
```diff
@@ -1,4 +1,4 @@
-describe Sufia::Forms::WorkForm do
+describe Sufia::Forms::WorkForm, :no_clean do
```
cam156: @jcoyne is this no_clean needed?

jcoyne: It's not *needed*, but it allows you to run this test without having fedora or solr running.

spec/forms/sufia/forms/work_form_spec.rb
```diff
@@ -17,4 +17,34 @@
...
+      let(:attributes) { { permissions_attributes: [{ type: 'person', name: 'justin', access: 'edit' }] } }
+
+      context "without mediated deposit" do
+        it { is_expected.to eq ActionController::Parameters.new(permissions_attributes: [ActionController::Parameters.new(type: 'person', name: 'justin', access: 'edit')]).permit! }
+      end
+      context "with mediated deposit" do
+        before do
+          allow(Flipflop).to receive(:enable_mediated_deposit?).and_return(true)
+        end
+        it { is_expected.to eq ActionController::Parameters.new(permissions_attributes: []).permit! }
```
jcoyne: it's `permit!` is required because it sets a `permitted=true` flag within the `Parameters` object. This is part of the `eq` comparison.

PR #2800, opened 2016-10-19
Downloads controller should load uploaded files from Sufia's cache if Fedora is down

mjgiarlo:
This is a proof of concept for how uploaded files might be served to users even if Fedora is down. Haven't worked on specs yet. Could use feedback on this approach.

### DIFF
--- a/app/controllers/concerns/sufia/download_behavior.rb
+++ b/app/controllers/concerns/sufia/download_behavior.rb
@@ -1,0 +1,32 @@
+# Extends CurationConcerns::DownloadBehavior to allow loading cached uploaded files
+module Sufia
+  module DownloadBehavior
+    extend ActiveSupport::Concern
+    include CurationConcerns::DownloadBehavior
+
+    # Override to handle when file is a File
+    def show
+      return super unless file.is_a? File
+      # For derivatives stored on the local filesystem
+      response.headers['Accept-Ranges'] = 'bytes'
+      response.headers['Content-Length'] = File.size(file).to_s
+      send_file file, derivative_download_options
+    end
+
+    protected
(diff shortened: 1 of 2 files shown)


cam156:
@mjgiarlo Interesting.  If I am reading this correctly for original content we go to Fedora first and for derivatives we go to the filesystem first?

mjgiarlo:
@cam156 For derivatives, we *only* go to the filesystem, AFAIK. For uploaded content, we pull from Fedora first because it's the current default behavior; this PR only adds fallback behavior if Fedora is down.

awead:
@mjgiarlo @cam156 when files are first created, they're put into an uploads directory, which by default is `tmp/uploads`. For characterization and derivative creation, we use that local file for the rest of the creation process which creates a derivative and places it in `tmp/derivatives`. From then, all calls to the thumbnail are to the local filesystem and not Fedora.

If that file in the uploads directly sticks around, we could keep serving it as the original content if Fedora went down. However, I don't know how this would behave with updates and new versions. In other words, you could have the potential for downloading an older or outdated version of the file if Fedora was down.

hackmastera:
might be good to ask some sysadmins. I asked @greensun and he said many of these same things.

"If Fedora is down things are broken." Is the purpose to buy time in fixing fedora? More about the use case driving the pr?

This feels like it would encourage bad practice (filling up arbitrary directories on the filesystem) or worse, stress people out for making the right decision (deleting duplicate data when it's no longer in use).

mjgiarlo:
Thanks for entertaining this proof of concept, folks! It's given me some other ideas for how we might satisfy this functional requirement.

PR #2810, opened 2016-10-21
Link fileset to work as thumbnail and representative

hackmastera:
This makes the thumbnail show in search results and on work show page.


hackmastera:
@cam156

PR #2811, opened 2016-10-21
added showpage workflow information

### LINKED ISSUE #2802: Surface showpage workflow information
Showpage for an object should show the workflow actions for the object. This should be surfaced from Curation Concerns.

luisgreg99:
Fixes #2802 

This PR exposes the workflow status and workflow actions bar in the show-page from curation concerns.

### DIFF
--- a/app/assets/javascripts/sufia/workflow_actions_affix.js
+++ b/app/assets/javascripts/sufia/workflow_actions_affix.js
@@ -1,0 +1,14 @@
+Blacklight.onLoad(function() {
+  if ($('.workflow-actions').length) {
+    $(document).on('scroll', function() {
+      var workflowDiv = $('#workflow_controls');
+      var workflowDivPos = $('.workflow-actions').offset().top + $('#workflow_controls').height();
+      workflowDiv.removeClass('workflow-affix');
+      if(workflowDivPos > ($(window).scrollTop() + $(window).height())){
+        workflowDiv.addClass('workflow-affix');
+      } else {
+        workflowDiv.removeClass('workflow-affix');
+      }
+    });
+  }
+});
--- a/app/views/curation_concerns/base/_workflow_actions_widget.erb
+++ b/app/views/curation_concerns/base/_workflow_actions_widget.erb
@@ -1,0 +1,3 @@
+<div class="workflow-actions">
+  <%= render 'workflow_actions', presenter: @presenter if @presenter.workflow.actions.present? %>
+</div>
--- a/app/views/curation_concerns/base/show.html.erb
+++ b/app/views/curation_concerns/base/show.html.erb
@@ -4,6 +4,10 @@
     <header>
       <%= render 'work_title', presenter: @presenter %>
     </header>
+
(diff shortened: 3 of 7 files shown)


cam156:
@vantuyls What is this going to look like for the single operation workflows?

It would be easier for people to omit if the workflow display were it's own partial.

jcoyne:
@luisgreg99 can you add a check for `Flipflop.enable_mediated_deposit?` on the view so that the `div` doesn't show for people who haven't enabled it?

app/assets/javascripts/sufia/workflow_actions_affix.js
```diff
@@ -0,0 +1,12 @@
+Blacklight.onLoad(function() {
+  $(document).on('scroll', function() {
+    var workflowDiv = $('#workflow_controls');
+
+    var workflowDivPos = $('.workflow-actions').offset().top + $('.workflow-actions').height();
```
jcoyne: we need a way to gracefully exit out of here if $('.workflow-actions') isn't found.

app/views/curation_concerns/base/_workflow_actions_widget.erb
```diff
@@ -0,0 +1,5 @@
+<div class="workflow-actions">
+  <% if Flipflop.enable_mediated_deposit? %>
```
jcoyne: Can you put this outside the workflow-actions div?

app/assets/javascripts/sufia/workflow_actions_affix.js
```diff
@@ -0,0 +1,16 @@
+Blacklight.onLoad(function() {
+  $(document).on('scroll', function() {
+    var workflowDiv = $('#workflow_controls');
+
+    if ($('.workflow-actions').length) {
```
jcoyne: May I suggest you move this check outside of the `$(document).on('scroll',...`, that way, it doesn't check this on every scroll action.

app/views/sufia/admin/_sidebar.html.erb
```diff
@@ -21,7 +21,11 @@
                     </li>
                     <li><%= link_to curation_concerns.admin_workflow_path do %>
                       <span class="fa fa-code-fork"></span> <%= t('sufia.admin.sidebar.workflow') %>
-                        <% end %>
+                      <% end %>
```
jcoyne: The indent changed here such that it no longer matches up with the `link_to` call.  Can you fix it?

app/views/curation_concerns/base/show.html.erb
```diff
@@ -22,8 +26,11 @@
 <div class="row">
   <div class="col-sm-12">
     <%= render 'items', presenter: @presenter %>
+    <%= render 'workflow_actions_widget', presenter: @presenter %>
```
jcoyne: Can we move the `FlipFlop` check out of the `workflow_actions_widget` partial and into here?  This will keep the program for having to find a partial that doesn't do anything.

Issue #2145, opened 2016-06-06
Migration::Import::CollectionBuilder


cam156:
I would like to create an class that creates a Sufia PCDM Collection based on Sufia 6.0 exported Metadata hash. See #2141 for calling service.

This class should have the following methods:
* build - builds a Collection based on the metadata passed in 
  * parameter: options # metadata about the Collection to be created

This is done when I can:
Migration::Import::Collection.build produces a Collection with the same descriptive Metadata as is passed in via options.

* open Question are we expecting the builder or the translator to create the relationships to the works?

An example of an Collection importer can be found in Sufia [import_s7](https://github.com/projecthydra/sufia/blob/import_s6/lib/sufia/import/import_service.rb#L176)

cam156:
Run this after work import and create relationships to the works.

PR #2812, opened 2016-10-21
Add collection builder for import fixes projecthydra/sufia#2145

hackmastera:
@cam156

### DIFF
--- a/lib/sufia/import/collection_builder.rb
+++ b/lib/sufia/import/collection_builder.rb
@@ -1,0 +1,47 @@
+# Builder for generating a collection including permissions
+#
+module Sufia::Import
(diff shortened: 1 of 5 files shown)


lib/sufia/import/collection_builder.rb
```diff
@@ -0,0 +1,47 @@
...
+    end
+
+    def get_members(data)
+      members = []
+      data.each do |i|
+        members << Sufia.primary_work_type.find(i)
+      end
+      members
+    rescue ActiveFedora::ObjectNotFoundError
+      raise "GenericFiles must be imported before Collections"
```
cam156: ❤️  Better error messahe

PR #2815, opened 2016-10-24
Undo #2814 now that it is no longer needed.

mjgiarlo:
The default minter in AF::Noid 2.0.0.beta4 was changed to the database-backed one, which required #2814 to get the test suite working again. This is no longer required with AF::Noid 2.0.0.beta5.


mjgiarlo:
Dang, :red_circle:. Will come back to this once https://github.com/projecthydra/curation_concerns/pull/1061 is merged and a new CC is released.

PR #2817, opened 2016-10-25
Set the human readable name using the supplied name

jcoyne:

Otherwise all the works will display with the same names by default.

### DIFF
--- a/lib/generators/sufia/work_generator.rb
+++ b/lib/generators/sufia/work_generator.rb
@@ -23,9 +23,10 @@ module Sufia
     end
 
     def inject_sufia_work_behavior
-      insert_into_file "app/models/#{name.underscore}.rb", after: 'include ::CurationConcerns::BasicMetadata' do
+      underscored_name = name.underscore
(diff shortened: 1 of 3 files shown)


lib/generators/sufia/work_generator.rb
```diff
@@ -23,9 +23,10 @@ def register_work
     end
 
     def inject_sufia_work_behavior
-      insert_into_file "app/models/#{name.underscore}.rb", after: 'include ::CurationConcerns::BasicMetadata' do
+      underscored_name = name.underscore
+      insert_into_file "app/models/#{underscored_name}.rb", after: 'include ::CurationConcerns::BasicMetadata' do
         "\n  include Sufia::WorkBehavior" \
-        "\n  self.human_readable_type = 'Work'"
+        "\n  self.human_readable_type = '#{underscored_name.humanize}'"
```
mjgiarlo: IMO, this should use `#titleize` instead of `#humanize`. The text `'New Generic work'` below looks like it ought to be `'New Generic Work'` instead (to my :eyes: anyway).

mjgiarlo:
Thanks for the change, @jcoyne. :+1:

Issue #2142, opened 2016-06-06
Migration::Import::CollectionTranslator


cam156:
I would like to create an class that imports (and translates) a Sufia 6.0 exported Metadata file into a Sufia PCDM Collection. See #2137 for executable.

This class should have the following methods:
- import - loops through the files found in the file_pattern and maps the Collection to a Collection
  - parameter: file_pattern # an array of ids to process

This is done when I can:

Migration::Import::CollectionTranslator.import produces a Collection for each file in the directory specified by the input pattern, and that each Collection contains the same descriptive Metadata as the original Collection and has the same members as the original Collection (points to the new GenericWorks instead of the old GenericFiles, but the same ids).

An example of an Collection Convertor can be found in Sufia [import_s7](https://github.com/projecthydra/sufia/blob/import_s6/lib/sufia/import/import_service.rb#L176)

PR #2822, opened 2016-10-26
Add collection translator, fixes projecthydra/sufia#2142

hackmastera:
@cam156

### DIFF
--- a/lib/sufia/import/collection_translator.rb
+++ b/lib/sufia/import/collection_translator.rb
@@ -1,0 +1,35 @@
+module Sufia::Import
+  # Imports a Sufia 6.0-exported Collection into a Sufia PCDM Collection
+  class CollectionTranslator
+    def initialize(settings)
+      @collection_builder = CollectionBuilder.new(settings)
+    end
+
+    # loop through the files found in import_dir and map each Collection
+    # @param import_dir a directory containing export files, e.g.: collection_[id].json
+    # @return list of imported files
+    def import(import_dir, filename_prefix)
+      filename_prefix ||= "collection_"
+      raise "No such directory: '#{import_dir}'" unless Dir.exist?(import_dir)
+      # get filenames
+      files = Dir.glob(File.join(import_dir, '*')).select { |f| exported_json?(f, filename_prefix) }
+      files.each { |f| import_file(f) }
+    end
+
+    private
+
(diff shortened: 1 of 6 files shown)


lib/sufia/import/collection_translator.rb
```diff
@@ -0,0 +1,35 @@
...
+  # Imports a Sufia 6.0-exported Collection into a Sufia PCDM Collection
+  class CollectionTranslator
+    def initialize(settings)
+      @collection_builder = CollectionBuilder.new(settings)
+    end
+
+    # loop through the files found in import_dir and map each Collection
+    # @param import_dir a directory containing export files, e.g.: collection_[id].json
+    # @return list of imported files
+    def import(import_dir, filename_prefix)
```
cam156: @hackmastera This code for opening files seems quite simmilar to the GenericFileTranslator: https://github.com/projecthydra/sufia/blob/7.2-migration/lib/sufia/import/generic_file_translator.rb#L17-L23

May indicate that a base class would be nice.

lib/sufia/import/collection_translator.rb
```diff
@@ -0,0 +1,35 @@
...
+      files = Dir.glob(File.join(import_dir, '*')).select { |f| exported_json?(f, filename_prefix) }
+      files.each { |f| import_file(f) }
+    end
+
+    private
+
+      def import_file(filename)
+        Rails.logger.debug "Importing #{File.basename(filename)}"
+        file_str = File.read(filename)
+        json = JSON.parse(file_str)
```
cam156: This code is also strikingly similar to GenericFileTranslator.

lib/sufia/import/collection_translator.rb
```diff
@@ -0,0 +1,35 @@
...
+        collection = @collection_builder.build(json)
+        collection.save
+      end
+
+      def exported_json?(f, filename_prefix)
```
cam156: This method is the same as the one in GFT except for the name.

hackmastera:
@cam156 yes that's exactly the refactoring I was referring to earlier! I'm working on the config file bit first since it will get rid of the need for the username / password settings and simplify things, then will do this refactor. I think it would be easier for me to merge first if that's your only concern. But I defer to your judgment!

PR #2827, opened 2016-10-26
Move fedora username/password to config file instead of parameters

hackmastera:
@cam156


exe/sufia_import
```diff
@@ -7,9 +7,7 @@ options.verbose = false
 options.import_dir = "tmp/import"
 # TODO: add collection like options.mappings = ["Sufia::Import::CollectionTranslator=collection_", "Sufia::Import::GenericFileTranslator=generic_file_"]
 options.mappings = ["Sufia::Import::GenericFileTranslator=generic_file_"]
-options.import_binary = true
-options.user = 's6user'
-options.password = 's6password'
+options.import_binary = false
```
cam156: Did you intend to not import binaries by default?

hackmastera:
yeah you're right i will change the default back. i was thinking of it as a test run and default off was safer but it's creating the objects either way so it's not really a test run!

hackmastera:
@cam156 this is ready for another look if you please!

PR #2829, opened 2016-10-26
Hides inactive work from search results and adds functionality to ask if work is suppressed.

### LINKED ISSUE #2828: Hide inactive work from search results if mediated deposit it active
If mediated deposit is active, then we want to make sure inactive objects in a workflow don't show up in the search results. This can be done via the solr document and a simple solr query.

Relates #2640 

This is simply the first step in hiding inactive objects throughout the system.

straleyb:
Fixes #2828 Relates #2640 

This introduces a new module called `Sufa::Publishable`, which extends a module in curation concerns called `CurationConcerns::Publishable`. This is added on as a behavior on a "GenericWork" which allows for the asking of whether or not the work is suppressed or not. Suppression depends on the state of the Work. If the work is inactive, then the work is "suppressed". Otherwise, it is not suppressed and can be accessed by users other than its creator.

This also adds the functionality of hiding an inactive work from the search results.

### DIFF
--- a/spec/models/generic_work_spec.rb
+++ b/spec/models/generic_work_spec.rb
@@ -13,6 +13,26 @@ describe GenericWork do
     end
   end
 
+  describe "suppressed?" do
+    let(:work) { described_class.new(title: ['demoname']) { |gw| gw.apply_depositor_metadata("user") } }
+    before do
+      allow(work).to receive(:state).and_return(::RDF::URI(activity_uri))
+    end
+    context "When a work's state is inactive" do
+      let(:activity_uri) { 'http://fedora.info/definitions/1/0/access/ObjState#inactive' }
+      it "is suppressed" do
+        expect(work).to be_suppressed
+      end
+    end
+
+    context "When a work's state is active" do
+      let(:activity_uri) { 'http://fedora.info/definitions/1/0/access/ObjState#active' }
+      it "is not suppressed" do
+        expect(work).not_to be_suppressed
+      end
+    end
+  end
+
   describe "created for someone (proxy)" do
     let(:work) { described_class.new(title: ['demoname']) { |gw| gw.apply_depositor_metadata("user") } }
     let(:transfer_to) { create(:user) }
--- a/app/indexers/sufia/work_indexer.rb
+++ b/app/indexers/sufia/work_indexer.rb
@@ -12,6 +12,7 @@ module Sufia
         admin_set_label = object.admin_set.to_s
         solr_doc[Solrizer.solr_name('admin_set', :facetable)] = admin_set_label
         solr_doc[Solrizer.solr_name('admin_set', :stored_searchable)] = admin_set_label
+        solr_doc[Solrizer.solr_name('suppressed', STORED_BOOL)] = object.suppressed?
       end
     end
   end
--- a/app/models/concerns/sufia/publishable.rb
+++ b/app/models/concerns/sufia/publishable.rb
@@ -1,0 +1,15 @@
+module Sufia
+  module Publishable
+    extend CurationConcerns::Publishable
+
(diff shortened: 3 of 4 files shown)


spec/models/generic_work_spec.rb
```diff
@@ -13,6 +13,26 @@
...
 
+  describe "suppressed?" do
+    let(:work) { described_class.new(title: ['demoname']) { |gw| gw.apply_depositor_metadata("user") } }
+    before do
+      allow(work).to receive(:state).and_return(::RDF::URI(activity_uri))
+    end
+    context "When a work's state is inactive" do
+      let(:activity_uri) { 'http://fedora.info/definitions/1/0/access/ObjState#inactive' }
+      it "returns a truthy statement" do
+        expect(work.suppressed?).to be_truthy
```
mjgiarlo: Can you change this to:

```ruby
expect(work).to be_suppressed
```
:question:

spec/models/generic_work_spec.rb
```diff
@@ -13,6 +13,26 @@
...
+      end
+    end
+
+    context "When a work's state is active" do
+      let(:activity_uri) { 'http://fedora.info/definitions/1/0/access/ObjState#active' }
+      it "returns a falsey statement" do
+        expect(work.suppressed?).to be_falsey
```
mjgiarlo: Can you change this to:

```ruby
expect(work).not_to be_suppressed
```
:question:

spec/models/generic_work_spec.rb
straleyb: Sure! I didn't know RSpec knew how to translate a truthy or falsey statements into #be_(blank) statements. Ill get that changed pushed up now.

spec/models/generic_work_spec.rb
```diff
@@ -13,6 +13,26 @@
...
   end
 
+  describe "suppressed?" do
+    let(:work) { described_class.new(title: ['demoname']) { |gw| gw.apply_depositor_metadata("user") } }
+    before do
+      allow(work).to receive(:state).and_return(::RDF::URI(activity_uri))
+    end
+    context "When a work's state is inactive" do
+      let(:activity_uri) { 'http://fedora.info/definitions/1/0/access/ObjState#inactive' }
+      it "returns a truthy statement" do
```
mjgiarlo: Can you also change the label here to match what the test is testing now? Like `it 'is suppressed' do`

spec/models/generic_work_spec.rb
```diff
@@ -13,6 +13,26 @@
...
+        expect(work).to be_suppressed
+      end
+    end
+
+    context "When a work's state is active" do
+      let(:activity_uri) { 'http://fedora.info/definitions/1/0/access/ObjState#active' }
+      it "returns a falsey statement" do
```
mjgiarlo: Can you also change the label here to match what the test is testing now? Like `it 'is not suppressed' do`

straleyb:
@mjgiarlo Is there an easy way to drop the picked up commit? I don't feel like simply deleting the code is the right way to do it.....

straleyb:
@jcoyne Ill put this wherever you all want it 👍  Should we take a pass through the Sufia mediated deposit tickets and find out which you prefer to go in CC instead? So we know where do put future work.

mjgiarlo:
@straleyb Cool. Just so you know, @vantuyls and I are sweeping through all the mediated deposit issues in Sufia to clearly mark any tickets that are ambiguous about where they should be implemented. If you want to do a quick pass through those once too at your leisure next week, @jcoyne, that'd be good. Thanks!

mjgiarlo:
Closing, since @straleyb is taking this over to CC. :clap:

PR #2830, opened 2016-10-26
Add Managers, Depositors and Viewers to a permission template

jcoyne:

screen shot 2016-10-26 at 4 34 58 pm

### DIFF
--- a/app/forms/sufia/forms/admin_set_form.rb
+++ b/app/forms/sufia/forms/admin_set_form.rb
@@ -4,10 +4,21 @@ module Sufia
       self.model_class = AdminSet
       self.terms = [:title, :description, :thumbnail_id]
 
+      # @param model [AdminSet]
+      # @param permission_template [PermissionTemplate]
+      def initialize(model, permission_template)
+        super(model)
+        @permission_template = permission_template
+      end
+
       # Cast any array values on the model to scalars.
       def [](key)
         return super if key == :thumbnail_id
         super.first
+      end
+
+      def permission_template
+        PermissionTemplateForm.new(@permission_template)
       end
 
       class << self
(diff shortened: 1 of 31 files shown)


app/forms/sufia/forms/admin_set_form.rb
```diff
@@ -4,12 +4,23 @@ class AdminSetForm < CurationConcerns::Forms::CollectionEditForm
       self.model_class = AdminSet
       self.terms = [:title, :description, :thumbnail_id]
 
+      # @param model [AdminSet]
+      # @param permission_template [PermissionTemplate]
+      def initialize(model, permission_template)
+        super(model)
+        @permission_template = permission_template
```
mjgiarlo: Does this use the param or the method? Why give them the same name?

config/locales/sufia.en.yml
```diff
@@ -443,3 +474,8 @@ en:
       admin_set:
         create: 'Save'
         update: 'Save'
+      sufia_permission_template:
```
mjgiarlo: Why are this translation and `sufia_permission_template_access` prefixed with `sufia_`? Because of how nested attrs work?

jcoyne: because the model is in the Sufia namespace. This is how rails finds a i18n key for a model in a namespace.

app/forms/sufia/forms/admin_set_form.rb
jcoyne: In ruby a local variable always has precedence over a method call.

mjgiarlo:
@jcoyne rebase, pls.

mjgiarlo:
I'd like one non-Stanford approval to this PR before merging it. @awead @cam156 @hackmastera @jeremyf @projecthydra/sufia-code-reviewers ?

awead:
I've gave it a quick look and it looks good to me. However, Penn State won't be actively using any of the admin set code until we get our 7.2 version out the door, which is some months away. If others who are more actively involved in this code care to weigh-in, I'd choose their advice over mine.

mjgiarlo:
@awead Well, it's been 4 days, so I'll take it. Thanks!

PR #2834, opened 2016-10-27
Consolidate unique layout content

cbeer:


jcoyne:
My one concern is that this limits our ability for the admin page to have styles that apply only in the admin context.  As we're talking consolidation, probably not a big deal in the long run.

mjgiarlo:
Kicked the build to get it going again.

PR #2839, opened 2016-10-27
i18n for form_progress partial

### LINKED ISSUE #2836: Change text from "Enter required metadata" to "Enter required descriptions"
The text on the right sidebar (under "requirements") says "Enter required metadata".  The tab for the metadata is titled "Descriptions".  So we should harmonize these terms so that the user will know to use the "descriptions" tab.

screen shot 2016-10-27 at 10 35 17 am

jcoyne:
Also changes "Enter required metadata" to "Describe your work" fixes #2836

### DIFF
--- a/app/views/curation_concerns/base/_form_progress.html.erb
+++ b/app/views/curation_concerns/base/_form_progress.html.erb
@@ -5,11 +5,11 @@
   <div class="list-group">
     <div class="list-group-item">
       <fieldset>
-        <legend class="legend-save-work">Requirements</legend>
+        <legend class="legend-save-work"><%= t('.requirements') %></legend>
         <ul class="requirements">
-          <li class="incomplete" id="required-metadata">Enter required metadata</li>
+          <li class="incomplete" id="required-metadata"><%= t('.required_descriptions') %></li>
           <% if Sufia.config.work_requires_files %>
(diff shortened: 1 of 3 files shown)


app/views/curation_concerns/base/_form_progress.html.erb
```diff
@@ -43,6 +43,6 @@
     <%= link_to t(:'helpers.action.cancel'),
                 sufia.dashboard_index_path,
                 class: 'btn btn-default' %>
-    <%= f.submit 'Save', class: 'btn btn-primary', onclick: "confirmation_needed = false;", id: "with_files_submit", name: "save_with_files" %>
+    <%= f.submit class: 'btn btn-primary', onclick: "confirmation_needed = false;", id: "with_files_submit", name: "save_with_files" %>
```
mjgiarlo: Should you be using `t('sufia.helpers.submit.create)` (or some such) here with the new `create` translation below or is that automagic?

jcoyne: it uses the rails defaults. in this case `helpers.submit.create`

mjgiarlo: :sparkles:

PR #2847, opened 2016-10-31
Extract translator logic into a base class

hackmastera:
@cam156

### DIFF
--- a/lib/sufia/import/generic_file_translator.rb
+++ b/lib/sufia/import/generic_file_translator.rb
@@ -1,35 +1,26 @@
 module Sufia::Import
   # Imports a Sufia 6.0-exported GenericFile into a Sufia PCDM GernericWork and FileSet
-  class GenericFileTranslator
-    # @param Hash settings
+  class GenericFileTranslator < Translator
+    # @param Hash settings (see super for more)
     #   @attr import_binary    - (default true) Import the binary content from sufia6 fedora instance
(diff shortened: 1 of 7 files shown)


lib/sufia/import/generic_file_translator.rb
```diff
@@ -1,32 +1,20 @@
...
   # Imports a Sufia 6.0-exported GenericFile into a Sufia PCDM GernericWork and FileSet
-  class GenericFileTranslator
-    # @param Hash settings
+  class GenericFileTranslator < Translator
+    # @param Hash settings (see super for more)
     #   @attr import_binary    - (default true) Import the binary content from sufia6 fedora instance
     #     If true, sufia6_user and sufia6_password must be set in the appropriate sections of fedora.yml
     def initialize(settings)
+      super
+      @filename_prefix = settings.fetch(:filename_prefix, "generic_file_")
```
cam156: @hackmastera I'm wondering if creating a method or an attribute that each class can override might make more sense than copying the fetch into each class.

In Translator
```
def default_prefix
  raise "Must override"
end
```

In this class
```
def default_prefix
  "generic_file_"
end
```

In Translator constructor:
```
@filename_prefix = settings.fetch(:filename_prefix, default_prefix)
```

hackmastera: I like this idea! Will give it a try.

having trouble with rubocop in the meantime; for some reason I'm not able to pull the new version locally.

PR #2848, opened 2016-10-31
Allow depositing in admin sets you have deposit access to

jcoyne:
~~Merge #2830 first~~ (Merged.)

Fixes projecthydra-labs/lerna#427

### DIFF
--- a/app/services/sufia/admin_set_service.rb
+++ b/app/services/sufia/admin_set_service.rb
@@ -1,0 +1,5 @@
+module Sufia
+  # Returns AdminSets that the current user has permission to use.
+  class AdminSetService < CurationConcerns::AdminSetService
+  end
+end
--- a/lib/sufia/engine.rb
+++ b/lib/sufia/engine.rb
@@ -53,6 +53,10 @@ module Sufia
       end
 
       CurationConcerns::CurationConcern.actor_factory = Sufia::ActorFactory
+      # Don't try to load this class until the application has been generated
+      if defined? ::SearchBuilder
+        CurationConcerns::AdminSetService.default_search_builder = Sufia::AdminSetSearchBuilder
+      end
     end
 
     initializer 'sufia.assets.precompile' do |app|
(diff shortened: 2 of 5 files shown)


app/services/sufia/admin_set_service.rb
```diff
@@ -0,0 +1,5 @@
+module Sufia
```
cam156: Why is this class needed?  It seems like a straight override.

lib/sufia/engine.rb
```diff
@@ -53,6 +53,10 @@ class Engine < ::Rails::Engine
       end
 
       CurationConcerns::CurationConcern.actor_factory = Sufia::ActorFactory
+      # Don't try to load this class until the application has been generated
+      if defined? ::SearchBuilder
+        CurationConcerns::AdminSetService.default_search_builder = Sufia::AdminSetSearchBuilder
+      end
```
cam156: Should there be a warning about not using the correct search builder if this is not done?  What generator should have been run?

app/services/sufia/admin_set_service.rb
jcoyne: For the `initialize` and `models` methods

lib/sufia/engine.rb
jcoyne: It breaks when you try to load sufia, before you run `rails g sufia:install`, this just protects against that breakage.

app/services/sufia/admin_set_service.rb
cam156: @jcoyne when I search this change set only CurationConcerns::AdminSetService is being used.  I do not see Sufia::AdminSetService anywhere.  If it is meant to be used elsewhere some yard docs are needed.

jcoyne: Good catch! I've changed `app/views/curation_concerns/base/_form_relationships.html.erb` to reference this new class.

lib/sufia/engine.rb
cam156: It would be nice to give a warning in an else.  Could make someone's upgrade path easier
```
else
  logger.warn(' You must run "rails g sufia:install" for the Admin Set Search to run correctly')
end
```

PR #2855, opened 2016-11-03
Change builders to instantiate objects in #build instead of initialize

hackmastera:
@cam156 the builders were all instantiating objects on initialize which meant if they were used to build multiple objects they never actually created a new object to use.

I think this problem was hidden by the way mocking was done in the tests but it wasn't obvious to me how to modify the tests to reveal the bug. If you have an idea there I'd be glad to hear it.

The solution is to either instantiate a new builder every time you want to build something (which I think is contrary to the pattern implied by the name) or move this object creation out of the initializer and into #build, as I've done here.

### DIFF
--- a/lib/sufia/import/permission_builder.rb
+++ b/lib/sufia/import/permission_builder.rb
@@ -1,31 +1,24 @@
 # Create permissions on a Work or FileSet
 module Sufia::Import
   class PermissionBuilder
-    attr_reader :object
-
-    # @param PCDMObject object   Object for the permissions to be applied to
-    #
-    def initialize(object)
-      @object = object
-    end
-
     # Build permissions on a FileSet or a work based on the metadata from GenericFile
     #
-    #  @param Array[hash] generic_file_perms An array of hashes with the below keys
-    #  @option :agent    - Agent that has permisisons; example: "http://projecthydra.org/ns/auth/person#user1@example.com"
-    #  @option :mode     - access acl; example: "http://www.w3.org/ns/auth/acl#Write"
-    #  @option :acces_to - not used - Permissions are added to the object passed in the initializer
-    #  @option :id       - not used - Id for the permissions is generated
-    def build(generic_file_perms)
+    # @param PCDMObject object   Object for the permissions to be applied to
+    # @param Array[hash] generic_file_perms An array of hashes with the below keys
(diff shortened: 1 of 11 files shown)


lib/sufia/import/permission_builder.rb
```diff
@@ -3,20 +3,16 @@ module Sufia::Import
...
-    #  @option :id       - not used - Id for the permissions is generated
-    def build(generic_file_perms)
+    # @param PCDMObject object   Object for the permissions to be applied to
+    # @param Array[hash] generic_file_perms An array of hashes with the below keys
+    #   @option :agent    - Agent that has permisisons; example: "http://projecthydra.org/ns/auth/person#user1@example.com"
+    #   @option :mode     - access acl; example: "http://www.w3.org/ns/auth/acl#Write"
+    #   @option :acces_to - not used - Permissions are added to the object passed in the initializer
+    #   @option :id       - not used - Id for the permissions is generated
+    def build(object, generic_file_perms)
+      @object = object
```
cam156: @hackmastera  know this is the smallest change, but I think it would be more obvious to just pass the object through to create on line 21 instead of assigning it to an instance variable.

lib/sufia/import/version_builder.rb
```diff
@@ -3,19 +3,16 @@ module Sufia::Import
...
-    # @option :uri Link to content in Sufia 6 repository
-    # @option :label Version label
-    # @option :created date the version was created
+    #   @option :uri Link to content in Sufia 6 repository
+    #   @option :label Version label
+    #   @option :created date the version was created
     #
-    def build(generic_file_versions)
+    def build(file_set, generic_file_versions)
+      @file_set = file_set
```
cam156: @hackmastera this one is a bit more complex, but I still think passing it would make it more obvious later where the information is coming from.

spec/lib/sufia/import/collection_builder_spec.rb
```diff
@@ -15,7 +15,7 @@
   end
 
   it "creates a collection with metadata and permissions" do
```
cam156: I wonder if you created multiple collections within the same test if you would see the error?

hackmastera:
@cam156 ready for another look!

PR #2856, opened 2016-11-03
[WIP] Mike could you check this code

blancoj:
I can't seem to get sufia workflow to connect to the generic work.

### DIFF
--- a/app/actors/sufia/apply_permission_template_actor.rb
+++ b/app/actors/sufia/apply_permission_template_actor.rb
@@ -8,7 +8,8 @@ module Sufia
     protected
 
       def add_edit_users(attributes)
-        return unless attributes[:admin_set_id]
+        byebug
+        return unless attributes[:admin_set_id].present?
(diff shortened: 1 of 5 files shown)


mjgiarlo:
:speech_balloon: @blancoj 
> I have put my code in a branch as we talked about at the meeting. What I'm seeing now is that I'm able to deposit an item but the workflow I created in Sufia is is not being liked to the work - I don't see the area where I can traverse through the workflow on the show page.  Basically the way I got my workflow to show up in the .internal_test_app is to add this code:
> [snip]
> can you see this page https://github.com/projecthydra/sufia/compare/mediated_deposit_epic...blancoj:mde_notify?expand=1
> it has the diff between what's in the mediated_deposit_epic branch and my working branch
> Here is the where the new workflow is defined and the one line that makes it diff https://github.com/blancoj/sufia/blob/mde_notify/lib/generators/sufia/templates/workflow.json.erb#L32
> Now that I think about it perhaps I should do a pull request and you can see it better that way, no?
> here is the pull request:  https://github.com/projecthydra/sufia/pull/2856

PR #2859, opened 2016-11-07
[WIP] Test that file set builds distinct objects on repeated use

hackmastera:
@cam156 follow-up for discussion on #2855 

This demonstrates that the problem is with the file_set_builder (as well as work and collection builders, though I haven't demonstrated that here yet) and not so much the version and permission builders. However, I lean toward changing those also (as shown in the other PR) for consistency. What do you think?

This means I should write failing tests for work and collection builders, but tests should be fine as-is for version and permissions builders.

### DIFF
--- a/lib/sufia/import/file_set_builder.rb
+++ b/lib/sufia/import/file_set_builder.rb
@@ -2,15 +2,12 @@
 #
 module Sufia::Import
   class FileSetBuilder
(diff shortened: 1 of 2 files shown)


cam156:
@hackmastera I agree with keeping things consistent.  This is looking good!

PR #2864, opened 2016-11-07
send notification after item is marked as complete.

blancoj:
This is the code so that notifications are sent out when user has completed and deposit.

### DIFF
--- a/app/services/sufia/workflow/complete_notification.rb
+++ b/app/services/sufia/workflow/complete_notification.rb
@@ -1,0 +1,25 @@
+module Sufia
+  module Workflow
+    class CompleteNotification
+      def self.send_notification(entity:,
+                                 comment:,
+                                 user:,
+                                 recipients:)
+        id = entity.proxy_for_global_id.sub(/.*\//, '')
+        title = entity.proxy_for.title.join("; ")
+        comment = comment.comment.to_s || ''
+
+        subject = "Deposit has been approved"
+        message = "#{title} (#{id}) was approved by #{user.user_key}. #{comment}"
+
+        users_to_notify = recipients["to"] + recipients["cc"] + [depositor(work_id: id)]
+        users_to_notify.uniq.each { |u| u.send_message u, message, subject }
+      end
+
+      def self.depositor(work_id:)
+        user_key = ActiveFedora::Base.find(work_id).depositor
+        ::User.find_by(email: user_key)
+      end
+    end
+  end
+end
--- a/lib/generators/sufia/templates/workflow.json.erb
+++ b/lib/generators/sufia/templates/workflow.json.erb
@@ -1,24 +1,40 @@
 {
-  "workflows": [
-    {
-      "name": "one_step_mediated_deposit",
-      "label": "One-step mediated deposit workflow",
-      "description": "A single-step workflow for mediated deposit",
-      "actions": [
-         {
-           "name": "deposit",
-           "from_states": [],
-           "transition_to": "pending_review"
-         },
-         {
-           "name": "approve",
-           "from_states": [{"names": ["pending_review"], "roles": ["approving"]}],
-           "transition_to": "complete",
-           "methods": [
-             "CurationConcerns::Workflow::ActivateObject"
-           ]
-         }
-      ]
-    }
-  ]
+    "workflows": [
+        {
+            "name": "one_step_mediated_deposit",
+            "label": "One-step mediated deposit workflow",
+            "description": "A single-step workflow for mediated deposit",
+            "actions": [
+                {
+                    "name": "deposit",
+                    "from_states": [],
+                    "transition_to": "pending_review",
+                    "notifications": [
+                        {
+                            "notification_type": "email",
+                            "name": "Sufia::Workflow::PendingReviewNotification",
+                            "to": ["approving"],
+                            "cc": ["creating_user"]
+                        }
+                    ]
+                },
+                {
+                    "name": "approve",
+                    "from_states": [{"names": ["pending_review"], "roles": ["approving"]}],
+                    "transition_to": "complete",
+                    "notifications": [
+                        {
+                            "notification_type": "email",
+                            "name": "Sufia::Workflow::CompleteNotification",
+                            "to": ["creating_user"],
+                            "cc": ["approving"]
+                        }
+                    ],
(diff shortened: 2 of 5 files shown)


app/services/curation_concerns/workflow/confirm_submission.rb
```diff
@@ -0,0 +1,30 @@
+module CurationConcerns
+  module Workflow
+    class ConfirmSubmission
+      # params consists of
```
jcoyne: Please use the YARD doc style for this. See http://www.rubydoc.info/gems/yard/file/docs/Tags.md#option

app/services/curation_concerns/workflow/confirm_submission.rb
```diff
@@ -0,0 +1,30 @@
...
+
+        notify_users = params[:recipients]["to"] + params[:recipients]["cc"]
+        notify_users = notify_users.uniq
+
+        updated_at = params[:entity].updated_at
+        user_email = params[:user].email
+
+        # title and id
+        id = params[:entity].proxy_for_global_id.sub(/.*\//, '')
+        g = GenericWork.find id
```
jcoyne: Don't hard code GenericWork here. Not all works are GenericWork. Use `Sipity::Entity#proxy_for` instead.

app/services/curation_concerns/workflow/confirm_submission.rb
```diff
@@ -0,0 +1,30 @@
+module CurationConcerns
+  module Workflow
+    class ConfirmSubmission
+      # params consists of
+      #   entity: entity
+      #   comment: comment
+      #   user: user
+      #   recipients: recipients(notification)
+      def self.send_notification(params = {})
```
jcoyne: Can you use keyword arguments here rather than a Hash?

lib/generators/sufia/install_generator.rb
```diff
@@ -145,5 +145,9 @@ def install_blacklight_gallery
     def datatables
       generate 'jquery:datatables:install bootstrap3'
     end
+
+    def create_workflow
+      template('workflow.json.erb', "config/workflows/default_workflow.json")
```
jcoyne: Shouldn't this be something other than default?  We already have a default created by CurationConcerns.  Should this be `meditated_deposit_workflow.json`?

lib/generators/sufia/templates/workflow.json.erb
```diff
@@ -0,0 +1,60 @@
+{
+  "workflows": [
+    {
+      "name": "default",
```
jcoyne: Probably should change the name to `mediated_deposit`.

lib/generators/sufia/templates/workflow.json.erb
```diff
@@ -0,0 +1,60 @@
...
+      "label": "Digital collections workflow",
```
jcoyne: This is not a digital collections workflow, it is a "self-deposit workflow"

lib/generators/sufia/templates/workflow.json.erb
```diff
@@ -0,0 +1,60 @@
...
+      "description": "A multi-step workflow for digital collections",
```
jcoyne: Again, replace the digital collections.

lib/generators/sufia/templates/workflow.json.erb
```diff
@@ -0,0 +1,60 @@
...
+      "actions": [
+         {
+           "name": "finalize_digitization",
```
jcoyne: Why do we have all these steps for a self deposit workflow?

app/services/curation_concerns/workflow/confirm_submission.rb
```diff
@@ -0,0 +1,30 @@
+module CurationConcerns
```
jcoyne: You should not use the CurationConcerns namespace in Sufia. Use `Sufia` instead.

lib/generators/sufia/templates/workflow.json.erb
mjgiarlo: Switching the workflow to be MD-centered is a separate issue. This is just to wire up notifications.

jcoyne: @mjgiarlo so we should not generate a new workflow at all then?

blancoj: Perhaps the thing to do here is to make a change in CC to have this in the default workflow:
           "notifications": [
             {
               "notification_type": "email",
               "name": "Sufia::Workflow::ConfirmSubmission",
               "to": ["creating_user"],
               "cc": ["advising"]
             }

Then to test this issue, I could load the CC default work. If this makes sense, I'll write an issue and put a pull request for it. Let me know.

mjgiarlo: @jcoyne We generate the new workflow only to add a notification which lives in the Sufia namespace Do you think this is the wrong approach?

jcoyne: I think we ought to either:
1. create the mediated deposit workflow first
1. don't bother putting the notification service into a workflow.

If we don't pick one of these, we going to have to have another PR to undo this stuff.

mjgiarlo:
Fixes #2627

app/services/sufia/workflow/complete_notification.rb
```diff
@@ -0,0 +1,25 @@
+module Sufia
+  module Workflow
+    class CompleteNotification
+      def self.send_notification(entity:,
```
straleyb: Seems like in these two classes there is a bit of repeated code. Do we maybe want to make an object that builds a notification? @jcoyne @mjgiarlo Just a thought really.

mjgiarlo: This is indeed ripe for refactoring. Thanks, @straleyb.

app/services/sufia/workflow/complete_notification.rb
```diff
@@ -0,0 +1,25 @@
...
+                                 comment:,
+                                 user:,
+                                 recipients:)
+        id = entity.proxy_for_global_id.sub(/.*\//, '')
```
mjgiarlo: Dig that idea, too.

mjgiarlo:
Superseded by #2911.

PR #2865, opened 2016-11-08
Add a modal when there are many types of works to select from

### LINKED ISSUE #2434: Multiple work types: 2 'create work' links on dashboard
There are 2 'Create work' links on the dashboard. They share a hover animation but can be clicked on separately. They link to different routes (/concern/generic_works/new and /concern/snowflakes/new).
### Expected behavior

Unclear. Should there be a link for each work type? Or should there be some other way to say what kind of work you're creating?

First generate at least two work types:

```
rails generate sufia:work GenericWork
rails generate sufia:work Image
```

Then when you go to `/dashboard` you see
screen shot 2016-10-26 at 12 31 51 pm
### Related work

umbrella ticket is https://github.com/projecthydra/sufia/issues/2199

jcoyne:
screen shot 2016-11-08 at 3 47 34 pm
Fixes #2434

### DIFF
--- a/app/presenters/sufia/select_type_presenter.rb
+++ b/app/presenters/sufia/select_type_presenter.rb
@@ -1,0 +1,35 @@
+module Sufia
+  class SelectTypePresenter
+    def initialize(concern)
+      @concern = concern
+    end
+
+    attr_reader :concern
+
+    def icon_class
+      translate('icon_class')
+    end
+
+    def description
+      translate('description')
+    end
+
+    def name
+      translate('name')
+    end
+
+    private
+
+      def object_name
+        @object_name ||= concern.model_name.i18n_key
+      end
+
+      def translate(key)
+        defaults = []
+        defaults << :"sufia.select_type.#{object_name}.#{key}"
+        defaults << :"sufia.select_type.#{key}"
+        defaults << ''
+        I18n.t(defaults.shift, default: defaults)
+      end
+  end
+end
--- a/app/views/dashboard/_select_work_type.html.erb
+++ b/app/views/dashboard/_select_work_type.html.erb
@@ -1,0 +1,46 @@
+<div class="modal fade" id="worktypes-to-create" tabindex="-1" role="dialog" aria-labelledby="select-worktype-label">
+  <div class="modal-dialog" role="document">
+    <div class="modal-content">
+      <div class="modal-header">
+        <button type="button" class="close" data-dismiss="modal" aria-label="Close"><span aria-hidden="true">&times;</span></button>
+        <h4 class="modal-title" id="select-worktype-label"><%= t('sufia.dashboard.heading_actions.select_type_of_work') %></h4>
+      </div>
+      <div class="modal-body">
+        <table class="table">
+          <% presenter.each do |row_presenter| %>
+            <tr>
+              <td class="select-work-icon">
+                <%= link_to(new_polymorphic_path([main_app, row_presenter.concern]),
+                            class: "item-option contextual-quick-classify #{dom_class(row_presenter.concern, 'new').gsub('_', '-')}",
+                            role: 'menuitem'
+                    ) do %>
+                    <span class="<%= row_presenter.icon_class %>"></span>
+                <% end %>
+              </td>
+              <td class="select-work-description">
+                <%= link_to(new_polymorphic_path([main_app, row_presenter.concern]),
+                            class: "item-option contextual-quick-classify #{dom_class(row_presenter.concern, 'new').gsub('_', '-')}",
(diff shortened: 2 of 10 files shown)


app/presenters/sufia/select_type_presenter.rb
```diff
@@ -0,0 +1,35 @@
...
+      def object_name
+        @object_name ||= concern.model_name.i18n_key
+      end
+
+      def translate(key)
+        defaults = []
+        defaults << :"sufia.select_type.#{object_name}.#{key}"
+        defaults << :"sufia.select_type.#{key}"
+        defaults << ''
+        I18n.t(defaults.shift, default: defaults)
```
cam156: @jcoyne does this code really work to get the generic description if the specific one is not present?  I do not see any test for this.  I would think the code would need to be ` I18n.t(defaults.shift,default: I18n.t(defaults.shift, default: defaults))`

app/views/dashboard/_select_work_type.html.erb
```diff
@@ -0,0 +1,46 @@
+<div class="modal fade" id="worktypes-to-create" tabindex="-1" role="dialog" aria-labelledby="select-worktype-label">
+  <div class="modal-dialog" role="document">
```
cam156: @mtribone does this modal have the appropriate aria labels for accessibility?

lib/generators/sufia/work/templates/locale.en.yml.erb
```diff
@@ -1,4 +1,10 @@
 en:
+  sufia:
+    select_type:
+      <%= file_name %>:
+        name:               "<%= human_name.titleize %>"
+        description:        "<%= human_name %> works"
```
cam156: This default seems a bit unuseful.  How about `A work to store a <%= human_name %>`  The you would get 'A work to store a Book' vs 'Book works' 'A work to store a Generic Work' vs 'Generic Work works'

app/presenters/sufia/select_type_presenter.rb
```diff
@@ -0,0 +1,35 @@
+module Sufia
+  class SelectTypePresenter
+    def initialize(concern)
+      @concern = concern
+    end
+
+    attr_reader :concern
+
+    def icon_class
+      translate('icon_class')
```
cam156: @jcoyne I wonder if this should be the default thumbnail for the work and not a translated icon.

lib/generators/sufia/work/templates/locale.en.yml.erb
jcoyne: The idea is that people will change it.  It only really looks awful with `GenericWork`, but I assume most people will generate more specific work types and certainly will change these default values. That's why we're generating them right into the host app.

rotated8: Is there a place this can be documented? I fear expectations like this that aren't collected may become barriers to entry for new users.

app/presenters/sufia/select_type_presenter.rb
jcoyne: This works. This is the same pattern as used by `ActionView::Helpers::FormHelper` https://github.com/rails/rails/blob/master/actionview/lib/action_view/helpers/form_helper.rb#L1881-L1886.

app/presenters/sufia/select_type_presenter.rb
jcoyne: We need to have a different icon for each worktype. See: https://cloud.githubusercontent.com/assets/101482/19989183/0ca5f6a4-a1e2-11e6-8f7f-44845724f209.png

lib/generators/sufia/work/templates/locale.en.yml.erb
jcoyne: @rotated8 what exactly should be documented?  That you'll find i18n templates generated into your `config/locales` and that you should customize them?

jcoyne: @rotated8 is this sufficient? https://github.com/projecthydra/sufia/wiki/Customizing-Metadata

app/presenters/sufia/select_type_presenter.rb
cam156: This PR will create another location to consolidate with #2867

app/presenters/sufia/select_type_presenter.rb
cam156: @jcoyne I would still love to see a test...

jcoyne: I don't think we should test the behavior of another gem.  The line is covered and we are testing that all the public methods work correctly

PR #2907, opened 2016-11-16
making sure we do not use the incompatible google-client-api version

cam156:
Fixes the build for sufia 6.x-stable on Ruby 2.1


jcoyne:
Also, there is a typo in your commit message `rwt` ▶️ `jwt`

cam156:
@jcoyne Updated to pin google-client-api instead of jwt.  I'm not really tied to any option.  I just want the build to start passing again.

PR #2911, opened 2016-11-17
Notifications should be sent for workflow steps

### LINKED ISSUE #2627: Send notification to user when item in workflow needs their attention

mjgiarlo:
Can now see notifications sent to depositor and approver both on deposit and on approval. Fixes #2627, #2864.

# TODO 

- [x] Use FactoryGirl vs. real AR instances
- [x] Specs need to pass and need to be complete
- [x] Look for refactoring opportunities (DRY it out)
- [x] Comments not coming through in messages
- [x] `creating_user` role is not used (remove it from workflow json?)

### DIFF
--- a/app/services/sufia/workflow/abstract_notification.rb
+++ b/app/services/sufia/workflow/abstract_notification.rb
@@ -1,0 +1,39 @@
+module Sufia
+  module Workflow
+    class AbstractNotification
+      def self.send_notification(entity:, comment:, user:, recipients:)
+        new(entity, comment, user, recipients).call
+      end
+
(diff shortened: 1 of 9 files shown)


mjgiarlo:
@narogers does that workflow require any approval before things go live?

straleyb:
Gets a +1 in my book. Being able to extract that notification into an abstract class was nice. And then overriding the subject and message methods makes it easy to add any other notifications in the future.

narogers:
@mjgiarlo Review is a spot check of the images by a person to make sure derivatives look good before the originals get deleted from the server.

PR #2912, opened 2016-11-17
Implementing the validation service

### LINKED ISSUE #2134: Migration::Validate::Service
Create a class that validates a Fedora 4/Sufia 6 survey in a Sufia PCDM application by looping through each Migration::Survey::Item #2131  created by the surveyor #2133 and updating its status. See #2129 for full use case.

- register_model_mapping  # Default mapping { Collection: Collection.class, GenericFile: GenericWork.class}
  - parameters:
    - original_class   # class found by the Sufia 6.x survey
    - target_class     # expected class after migration
- call - loops through survey in the database and verifies each item's existence and type

This is done when I can:
- Migration::Survey::Surveyor.new  creates an instance of the surveyor
- instance.register_model_mapping adds additional model mappings to the registry
- instance.call updates each Migration::Survey::Item with a status:
  - :valid           # it exists and is the correct type
  - :type_error  # it exists but is not of the expected type
  - :missing      # it does not exists

cam156:
Fixes #2134 

This PR pulls in the Survey::Item from the sufia 6.x-stable branch along with implementing the validation service.


hackmastera:
oh, sweet, just saw this. will take a look in the morning!

PR #2929, opened 2016-11-22
Adding error logger and rescue to log errors instead of ending import

### LINKED ISSUE #2926: Log errors instead of failing the entire import
### Descriptive summary

In Sufia 7.2-migration. When running import on an item log errors instead of crashing the entire import.

This would mean adding a rescue around each item import and logging the error to an error log.

cam156:
Fixes #2926

Creates a logger that can be tested and reused.  Not using system log since we may want to process the migration log in some fashion separate from any other information.

The log is then used to rescue and note any error that occurs with each file being imported.

Additionally added a more detailed error message when processing collection members since an error on import on a Work or the Collections being process before the works could result in the error.

### DIFF
--- a/lib/sufia/import.rb
+++ b/lib/sufia/import.rb
@@ -7,6 +7,7 @@ require 'sufia/import/work_builder'
 require 'sufia/import/collection_builder'
 require 'sufia/import/generic_file_translator'
 require 'sufia/import/collection_translator'
+require 'sufia/import/log'
 
 module Sufia
   module Permissions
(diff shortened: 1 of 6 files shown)


hackmastera:
i really like the idea of having a different file for this log. Do you want to also move this over? https://github.com/psu-stewardship/sufia/blob/c61450d36c9a14d725d081af1f071389578a6b1b/lib/sufia/import/translator.rb#L30

cam156:
@hackmastera I was thinking about it more of an error log, but we certainly could add processing and completion status into the log file.  I was thinking it would be an easy way of reprocessing failed imports to have a file that only contained those files.  I guess the log file name should include error though.

hackmastera:
@cam156 I'm happy to merge; do you want me to wait for a filename change or do that later?

PR #2940, opened 2016-11-23
Increase test coverage

mjgiarlo:


.rubocop_todo.yml
```diff
@@ -91,6 +91,7 @@ RSpec/VerifiedDoubles:
 RSpec/AnyInstance:
   Exclude:
     - 'spec/controllers/sufia/admin/admin_sets_controller_spec.rb'
+    - 'spec/controllers/sufia/trophies_controller_spec.rb'
```
jcoyne: I'm not seeing why this is required.  I don't see any instances of this in the `trophies_controller_spec.rb`

PR #2942, opened 2016-11-28
create linked facets in index results

jenlindner:
A helper method to create links to facets in index results.

This is something I did for the Art Institute but wonder if it might be useful for Sufia. I looked at the available helper methods and none of them were quite what I needed so I essentially pulled all the bits I needed into this one. It lets me index and create links from the catalog_controller, putting the facet term into the options that BL passes, and using the > separator to show their hierarchy. 

```config.add_index_field solr_name("document_types", :stored_searchable),  label: AIC.documentType.label, helper_method: :link_to_each_facet_field, helper_facet: solr_name("document_types", :facetable).to_sym ```

example

I understand this might not be the best way to accomplish this, and would be happy to re-work it (and the tests) until it meets Sufia standards. But maybe it'll be useful and helpful to others.

* Add a helper method to create linked facets in index results  
* Add unit helper test of method

### DIFF
--- a/app/helpers/sufia/sufia_helper_behavior.rb
+++ b/app/helpers/sufia/sufia_helper_behavior.rb
@@ -123,6 +123,33 @@ module Sufia
       safe_join(links, ", ")
     end
 
+    # A Blacklight helper_method
+    #
+    # @example
+    #   link_to_each_facet_field({ value: "Imaging > Object Photography", config: { helper_facet: :document_types_sim }})
+    #   ```html
+    #   <a href=\"/catalog?f%5Bdocument_types_sim%5D%5B%5D=Imaging\">Imaging</a> &gt; <a href=\"/catalog?f%5Bdocument_types_sim%5D%5B%5D=Object+Photography\">Object Photography</a>
+    #   ```
+    #
+    # @param options [Hash] from blacklight invocation of helper_method
+    # @option options [Array<#strip>] :value
+    # @option options [Hash>] :config Example: { separator: '>', helper_facet: :document_types_sim, output_separator: '::' }
+    # @return the html_safe facet links separated by the given separator (default: ' > ') to indicate facet hierarchy
+    #
+    # @raise KeyError when the options are note properly configured
+    def link_to_each_facet_field(options)
+      config = options.fetch(:config)
+      separator = config.fetch(:separator, ' > ')
+      output_separator = config.fetch(:output_separator, separator)
+      facet_search = config.fetch(:helper_facet)
+      facet_fields = Array.wrap(options.fetch(:value)).first.split(separator).map(&:strip)
+
+      facet_links = facet_fields.map do |type|
+        link_to(type, main_app.search_catalog_path(f: { facet_search => [type] }))
+      end
+      safe_join(facet_links, output_separator)
+    end
+
     # Uses Rails auto_link to add links to fields
     #
     # @param field [String,Hash] string to format and escape, or a hash as per helper_method
--- a/spec/helpers/sufia_helper_spec.rb
+++ b/spec/helpers/sufia_helper_spec.rb
@@ -68,6 +68,29 @@ describe SufiaHelper, type: :helper do
     end
   end
 
+  describe "#link_to_each_facet_field" do
+    subject { helper.link_to_each_facet_field(options) }
+    context "with helper_facet and default separator" do
+      let(:options) { { config: { helper_facet: Solrizer.solr_name("document_types", :facetable).to_sym }, value: ["Imaging > Object Photography"] } }
+      it { is_expected.to eq("<a href=\"/catalog?f%5Bdocument_types_sim%5D%5B%5D=Imaging\">Imaging</a> &gt; <a href=\"/catalog?f%5Bdocument_types_sim%5D%5B%5D=Object+Photography\">Object Photography</a>") }
+    end
+
+    context "with helper_facet and optional separator" do
+      let(:options) { { config: { helper_facet: Solrizer.solr_name("document_types", :facetable).to_sym, separator: " : " }, value: ["Imaging : Object Photography"] } }
+      it { is_expected.to eq("<a href=\"/catalog?f%5Bdocument_types_sim%5D%5B%5D=Imaging\">Imaging</a> : <a href=\"/catalog?f%5Bdocument_types_sim%5D%5B%5D=Object+Photography\">Object Photography</a>") }
+    end
+
+    context "with :output_separator" do
+      let(:options) { { config: { helper_facet: Solrizer.solr_name("document_types", :facetable).to_sym, output_separator: ' ~ ', separator: ":" }, value: ["Imaging : Object Photography"] } }
+      it { is_expected.to eq("<a href=\"/catalog?f%5Bdocument_types_sim%5D%5B%5D=Imaging\">Imaging</a> ~ <a href=\"/catalog?f%5Bdocument_types_sim%5D%5B%5D=Object+Photography\">Object Photography</a>") }
+    end
+
+    context "with :no_spaces_around_separator" do
+      let(:options) { { config: { helper_facet: Solrizer.solr_name("document_types", :facetable).to_sym, output_separator: '~', separator: ":" }, value: ["Imaging : Object Photography"] } }
+      it { is_expected.to eq("<a href=\"/catalog?f%5Bdocument_types_sim%5D%5B%5D=Imaging\">Imaging</a>~<a href=\"/catalog?f%5Bdocument_types_sim%5D%5B%5D=Object+Photography\">Object Photography</a>") }
+    end
+  end
+
   describe "has_collection_search_parameters?" do
     subject { helper }
     context "when cq is set" do

### CODE REVIEW

File: app/helpers/sufia/sufia_helper_behavior.rb [hunk 1]
```diff
@@ -123,6 +123,20 @@ def index_field_link(options)
...
     end
 
+    # A Blacklight helper_method
+    # @param [Hash] options from blacklight invocation of helper_method
+    # @return the html_safe facet links separated by " > " to indicate facet hierarchy
+
+    def link_to_each_facet_field(options)
+      raise ArgumentError unless options[:config] && options[:value] && options[:config][:helper_facet]
+      facet_search = options[:config][:helper_facet]
+      facet_fields = options[:value].first.split(">").each(&:strip!)
```
jcoyne: @cbeer is there a way of configuring the hierarchical separator that we could use here?

jcoyne: @jenlindner can you use `map(&:strip)` instead? This is a more functional approach which I think makes it clearer what is going on.

jenlindner: Yes, I'll work on it.

jenlindner:
I made the separator default to ">" but take an option also just now, running tests.

app/helpers/sufia/sufia_helper_behavior.rb
```diff
@@ -123,6 +123,22 @@ def index_field_link(options)
...
     end
 
+    # A Blacklight helper_method
+    # @param [Hash] options from blacklight invocation of helper_method
+    # @return the html_safe facet links separated by " > " to indicate facet hierarchy
+
+    def link_to_each_facet_field(options)
+      raise ArgumentError unless options[:config] && options[:value] && options[:config][:helper_facet]
+
+      separator = options[:config][:separator] ? options[:config][:separator] : ">"
```
jcoyne: How about `separator = options.fetch(:config).fetch(:separator, '>')`  ?

app/helpers/sufia/sufia_helper_behavior.rb
```diff
@@ -123,6 +123,22 @@ def index_field_link(options)
...
+      facet_search = options[:config][:helper_facet]
```
jcoyne: This seems to be required, so let's do `options.fetch(:config).fetch(:helper_facet)`

app/helpers/sufia/sufia_helper_behavior.rb
```diff
@@ -123,6 +123,22 @@ def index_field_link(options)
...
+      facet_fields = options[:value].first.split(separator).map(&:strip!)
+      facet_links = facet_fields.map do |type|
+        link_to(type, main_app.search_catalog_path(f: { facet_search => [type] }))
+      end
+      safe_join(facet_links, " " + separator + " ")
```
jcoyne: Perhaps we need an `join_with` too?   That option could include the spaces if they are desired.

app/helpers/sufia/sufia_helper_behavior.rb
```diff
@@ -123,6 +123,22 @@ def index_field_link(options)
...
+    # A Blacklight helper_method
+    # @param [Hash] options from blacklight invocation of helper_method
+    # @return the html_safe facet links separated by " > " to indicate facet hierarchy
+
+    def link_to_each_facet_field(options)
+      raise ArgumentError unless options[:config] && options[:value] && options[:config][:helper_facet]
+
+      separator = options[:config][:separator] ? options[:config][:separator] : ">"
+      facet_search = options[:config][:helper_facet]
+      facet_fields = options[:value].first.split(separator).map(&:strip!)
```
jcoyne: switch `strip!` to `strip` (no `!`)

app/helpers/sufia/sufia_helper_behavior.rb
```diff
@@ -123,6 +123,22 @@ def index_field_link(options)
       safe_join(links, ", ")
     end
 
+    # A Blacklight helper_method
+    # @param [Hash] options from blacklight invocation of helper_method
+    # @return the html_safe facet links separated by " > " to indicate facet hierarchy
```
jcoyne: Add an example here of what you expect the solr document to look like, and the example configuration for using this method.

jcoyne:
@jenlindner are you using blacklight-hierarchy? https://github.com/sul-dlss/blacklight-hierarchy

app/helpers/sufia/sufia_helper_behavior.rb
jenlindner: I'm not having luck with this yet, but the other changes are in, and thanks.

app/helpers/sufia/sufia_helper_behavior.rb
```diff
@@ -123,6 +123,23 @@ def index_field_link(options)
...
+
+    def link_to_each_facet_field(options)
+      raise ArgumentError unless options[:config] && options[:value] && options[:config][:helper_facet]
```
jeremyf: As part of the ArgumentError, can you add the expectations of the options? (e.g. "Expected options to have keys :value and :config. Expected options[:config] to have key :helper_facet")

app/helpers/sufia/sufia_helper_behavior.rb
```diff
@@ -123,6 +123,23 @@ def index_field_link(options)
...
+
+      separator = options.fetch(:config).fetch(:separator, '>')
+      facet_search = options.fetch(:config).fetch(:helper_facet)
+      facet_fields = options[:value].first.split(separator).map(&:strip)
```
jeremyf: Wondering about using `options.fetch(:value)` instead. This might mean the above guard clause need not worry about testing if `options[:value]`. Also, it looks as though `options[:value]` assumes an array. I'm not aware of whether it could be a string instead. Would the following work?

```ruby
facet_fields = Array.wrap(options.fetch(:value)).first.split(separator).map(&:strip)
```

jeremyf:
I added commits addressing my questions. @jcoyne did my commit address your concern on documentation.

app/helpers/sufia/sufia_helper_behavior.rb
```diff
@@ -123,6 +123,32 @@ def index_field_link(options)
...
+    def link_to_each_facet_field(options)
+      config = options.fetch(:config)
+      separator = config.fetch(:separator, '>')
+      facet_search = config.fetch(:helper_facet)
+      facet_fields = Array.wrap(options.fetch(:value)).first.split(separator).map(&:strip)
+
+      facet_links = facet_fields.map do |type|
+        link_to(type, main_app.search_catalog_path(f: { facet_search => [type] }))
+      end
+      safe_join(facet_links, " " + separator + " ")
```
jcoyne: can we make this second argument be a config option? `:output_separator` perhaps?

jeremyf: @jcoyne Sounds good. Will add.

jenlindner: I thought that the problem was line 149, adding spaces to the separator in the safe_join, and couldn't find a better way to get them there. Is it okay as it is?

app/helpers/sufia/sufia_helper_behavior.rb
```diff
@@ -123,6 +123,33 @@ def index_field_link(options)
...
+    # @param options [Hash] from blacklight invocation of helper_method
+    # @option options [Array<#strip>] :value
+    # @option options [Hash>] :config Example: { separator: '>', helper_facet: :document_types_sim, output_separator: '::' }
+    # @return the html_safe facet links separated by the given separator (default: ' > ') to indicate facet hierarchy
+    #
+    # @raise KeyError when the options are note properly configured
+    def link_to_each_facet_field(options)
+      config = options.fetch(:config)
+      separator = config.fetch(:separator, '>')
+      output_separator = config.fetch(:output_separator, separator).to_s.strip
```
jcoyne: Put the spaces in the output separator

jenlindner: like this? 
``` output_separator = " " + output_separator + " "```

jcoyne: No, as the second argument to the fetch. Maybe I don't want spaces as part of my separator.

jenlindner: Spaces get stripped from the option by default and have to be explicitly added back in. I think the dominant use case has spaces, not having any would severely hinder readability, be a usability problem. I made them the default, with a config that can override this. Will that work?

jcoyne: Why would we need to strip spaces from the output separator?

jenlindner:
I think this will satisfy all the requirements - a default separator with spaces, a configurable output  separator (that no longer strips spaces or explicitly casts to a string) that can accept and pass along a character with or without spaces around it.

I'd been looking for a way to do a left and right pad around the separator string before passing it to the join at the end, instead of looking for a more creative way to allow for spaces/no spaces, and just could not find a simple way to do it, like a join_with(" "). What I found looked like overkill.

Let me know if this works.

jenlindner:
🎊 🎊

PR #2944, opened 2016-11-29
add notion of 'depositor review' to one step mediated deposit workflow

### LINKED ISSUE #2879: A reviewer in a one-step workflow should be able to send a work back to the depositor

### LINKED ISSUE #2880: Deposited works should be marked as inactive in one-step workflow

revgum:
fixes #2880 
fixes #2879 

The approach here would allow for a notion of an admin/reviewer type person to move a work back to `depositor_review` at any point in time (during `pending_review` or `complete`). The work is set as inactive and is hidden from search. After the depositor submits the updated work to `pending_review`, the admin/reviewer can direct the work back to `depositor_review` for even more review or to be `complete` and the work activated (visible in search).

### DIFF
--- a/lib/generators/sufia/templates/workflow.json.erb
+++ b/lib/generators/sufia/templates/workflow.json.erb
@@ -3,7 +3,7 @@
         {
             "name": "one_step_mediated_deposit",
             "label": "One-step mediated deposit workflow",
-            "description": "A single-step workflow for mediated deposit",
+            "description": "A single-step workflow for mediated deposit in which all deposits must be approved by a reviewer. Reviewer may also send deposits back to the depositor.",
             "actions": [
                 {
                     "name": "deposit",
@@ -15,9 +15,25 @@
                             "name": "Sufia::Workflow::PendingReviewNotification",
                             "to": ["approving"]
                         }
+                    ],
+                    "methods": [
+                        "CurationConcerns::Workflow::DeactivateObject"
                     ]
-                },
-                {
+                }, {
+                    "name": "request_changes",
+                    "from_states": [{"names": ["complete", "pending_review"], "roles": ["approving"]}],
+                    "transition_to": "changes_required",
+                    "notifications": [
+                        {
+                            "notification_type": "email",
+                            "name": "Sufia::Workflow::ChangesRequiredNotification",
+                            "to": ["approving"]
+                        }
+                    ],
+                    "methods": [
+                        "CurationConcerns::Workflow::DeactivateObject"
+                    ]
+                }, {
                     "name": "approve",
                     "from_states": [{"names": ["pending_review"], "roles": ["approving"]}],
                     "transition_to": "complete",
@@ -30,6 +46,17 @@
                     ],
                     "methods": [
                         "CurationConcerns::Workflow::ActivateObject"
+                    ]
+                }, {
+                    "name": "request_review",
+                    "from_states": [{"names": ["changes_required"], "roles": ["depositing"]}],
+                    "transition_to": "pending_review",
+                    "notifications": [
+                        {
+                            "notification_type": "email",
+                            "name": "Sufia::Workflow::PendingReviewNotification",
+                            "to": ["approving"]
+                        }
                     ]
                 }
             ]
--- a/spec/services/sufia/workflow/changes_required_notification_spec.rb
+++ b/spec/services/sufia/workflow/changes_required_notification_spec.rb
@@ -1,0 +1,32 @@
+require 'spec_helper'
+
+RSpec.describe Sufia::Workflow::ChangesRequiredNotification do
+  let(:approver) { create(:user) }
+  let(:depositor) { create(:user) }
+  let(:to_user) { create(:user) }
+  let(:cc_user) { create(:user) }
+  let(:work) { create(:generic_work, user: depositor) }
+  let(:entity) { create(:sipity_entity, proxy_for_global_id: work.to_global_id.to_s) }
+  let(:comment) { double("comment", comment: 'A pleasant read') }
+  let(:recipients) { { 'to' => [to_user], 'cc' => [cc_user] } }
+
+  describe ".send_notification" do
+    it 'sends a message to all users' do
+      expect(approver).to receive(:send_message).once.and_call_original
+
+      expect { described_class.send_notification(entity: entity, user: approver, comment: comment, recipients: recipients) }
+        .to change { depositor.mailbox.inbox.count }.by(1)
+        .and change { to_user.mailbox.inbox.count }.by(1)
+        .and change { cc_user.mailbox.inbox.count }.by(1)
+    end
+    context 'without carbon-copied users' do
+      let(:recipients) { { 'to' => [to_user] } }
+      it 'sends a message to the to user(s)' do
+        expect(approver).to receive(:send_message).once.and_call_original
+        expect { described_class.send_notification(entity: entity, user: approver, comment: comment, recipients: recipients) }
+          .to change { depositor.mailbox.inbox.count }.by(1)
+          .and change { to_user.mailbox.inbox.count }.by(1)
+      end
+    end
+  end
+end
--- a/app/services/sufia/workflow/changes_required_notification.rb
+++ b/app/services/sufia/workflow/changes_required_notification.rb
@@ -1,0 +1,22 @@
+module Sufia
+  module Workflow
+    class ChangesRequiredNotification < AbstractNotification
+      protected
+
+        def subject
+          'Your deposit requires changes'
+        end
+
+        def message
+          "#{title} (#{work_id}) requires additional changes before approval.\n\n '#{comment}'"
+        end
+
+      private
+
+        def users_to_notify
+          user_key = ActiveFedora::Base.find(work_id).depositor
+          super << ::User.find_by(email: user_key)
+        end
+    end
+  end
+end


lib/generators/sufia/templates/workflow.json.erb
```diff
@@ -1,40 +1,70 @@
 {
-    "workflows": [
+  "workflows": [
+    {
+      "name": "one_step_mediated_deposit",
+      "label": "One-step mediated deposit workflow",
+      "description": "A single-step workflow for mediated deposit",
+      "actions": [
         {
-            "name": "one_step_mediated_deposit",
```
jcoyne: why did this file change?

revgum: The issues were seeking a "depositor review" concept be added to the single step mediated deposit.. so I sketched out an approach to the generator that builds the workflow..

revgum: Also to introduce `DeactivateObject` on deposit.

jcoyne: NM, I see you added a state.

lib/generators/sufia/templates/workflow.json.erb
```diff
@@ -9,19 +9,45 @@
                     "name": "deposit",
```
mjgiarlo: Can you expand the description above now that the workflow? Something like `A single-step workflow for mediated deposit in which all deposits must be approved by a reviewer. Reviewer may also send deposits back to the depositor.`?

lib/generators/sufia/templates/workflow.json.erb
```diff
@@ -9,19 +9,45 @@
...
                     "from_states": [],
                     "transition_to": "pending_review",
+                    "attributes": {
```
mjgiarlo: I don't think these are used. Can you remove the `attributes` hashes from this file?

lib/generators/sufia/templates/workflow.json.erb
```diff
@@ -9,19 +9,45 @@
...
                             "cc": []
                         }
+                    ],
+                    "methods": [
+                        "CurationConcerns::Workflow::DeactivateObject"
                     ]
-                },
-                {
+                }, {
+                    "name": "depositor_review",
```
mjgiarlo: Use a verb or verb phrase here, like `request_changes`.

lib/generators/sufia/templates/workflow.json.erb
```diff
@@ -9,19 +9,45 @@
...
+                    "from_states": [{"names": ["complete", "pending_review"], "roles": ["depositor_reviewing"]}],
```
mjgiarlo: I believe you should be able to use the same role here as for the `approve` action, since the same users will be either approving or requesting changes for a deposit. That role is `approving`.

lib/generators/sufia/templates/workflow.json.erb
```diff
@@ -9,19 +9,45 @@
...
+                    "transition_to": "depositor_review",
```
mjgiarlo: I'd like to see this change to `changes_required` to reflect that this is a state that a person has requested. (Told you I was picky about names -- hope I didn't disappoint!)

lib/generators/sufia/templates/workflow.json.erb
```diff
@@ -9,19 +9,45 @@
...
+                    "attributes": {
```
mjgiarlo: Pls drop the `attributes` hash.

lib/generators/sufia/templates/workflow.json.erb
```diff
@@ -9,19 +9,45 @@
...
+                        "presentation_sequence": 2
+                    },
+                    "notifications": [
+                        {
+                            "notification_type": "email",
+                            "name": "Sufia::Workflow::PendingDepositorReviewNotification",
```
mjgiarlo: Rename to match `transition_to` value: `ChangesRequiredNotification`.

lib/generators/sufia/templates/workflow.json.erb
```diff
@@ -9,19 +9,45 @@
...
+                            "to": ["depositor_reviewing"],
```
mjgiarlo: Change to `approving` to match the role change above.

lib/generators/sufia/templates/workflow.json.erb
```diff
@@ -9,19 +9,45 @@
...
+                        }
+                    ],
+                    "methods": [
+                        "CurationConcerns::Workflow::DeactivateObject"
+                    ]
+                }, {
                     "name": "approve",
                     "from_states": [{"names": ["pending_review"], "roles": ["approving"]}],
                     "transition_to": "complete",
+                    "attributes": {
```
mjgiarlo: Pls remove.

lib/generators/sufia/templates/workflow.json.erb
```diff
@@ -9,19 +9,45 @@
                     "name": "deposit",
                     "from_states": [],
```
mjgiarlo: Do we need to add a `from_state` here for `changes_requested`? 

And, if we do, will that break the ability to execute this action when not currently in a state, since new deposits aren't yet in a state? That also implies we'll need to add a role here to match the role in the final action declared at the bottom of this file, which implies that we need code that automatically adds a depositing role to all registered users, else they won't be able to participate in this workflow.

Tagging @jcoyne and @jeremyf too for feedback on the above.

lib/generators/sufia/templates/workflow.json.erb
```diff
@@ -33,6 +59,10 @@
                     "methods": [
                         "CurationConcerns::Workflow::ActivateObject"
                     ]
+                }, {
+                    "name": "depositor_updated",
```
mjgiarlo: How about `request_review`?

lib/generators/sufia/templates/workflow.json.erb
```diff
@@ -33,6 +59,10 @@
...
+                    "from_states": [{"names": ["depositor_review"], "roles": ["depositor_updated"]}],
+                    "transition_to": "pending_review"
                 }
```
mjgiarlo: This action should probably call the `PendingReviewNotification` too.

lib/generators/sufia/templates/workflow.json.erb
```diff
@@ -33,6 +59,10 @@
                     "methods": [
                         "CurationConcerns::Workflow::ActivateObject"
                     ]
+                }, {
+                    "name": "depositor_updated",
+                    "from_states": [{"names": ["depositor_review"], "roles": ["depositor_updated"]}],
```
mjgiarlo: Change `from_state` name to `changes_requested`.

Change role to `depositing`. See above question re: making sure we add this role to all registered users (or all users who can deposit to an admin set, if this workflow is used in the context of an admin set).

lib/generators/sufia/templates/workflow.json.erb
revgum: My understanding is that the first action without a state becomes the "deposit" action.. https://github.com/projecthydra/curation_concerns/blob/master/app/services/curation_concerns/workflow/workflow_factory.rb#L54

So I don't think we have a `from_state` here and just have other actions transition to `pending_review` as a sort of state that admin/reviewers interact with for mediating works?

We briefly discussed that there would need to be some mechanism that sets a default role for all registered users.. Definitely interested in some ideas on how to best wire this up.

mjgiarlo: @revgum You may be right; the `from_state` may not be required here, and I don't want to break `find_deposit_action`. :)

mjgiarlo: @revgum I'm on the fence as to how best to set up this role. We could do it super early, like when a user is created, and/or as a background job when mediated deposit is enabled. We could also do it on demand, like: when the reviewer sends a submission back to the depositor, we can call an action that ensures the depositor has the right role to resubmit their deposit?

Thoughts, @jcoyne @jeremyf?

revgum: @mjgiarlo @jcoyne @jeremyf I wonder if leveraging the `work.depositor` would be adequate for this type of feature. Let's say, when the depositor of a work is on `#show` or `#edit`, they are exposed the necessary bits of UI that aids in the awareness of a required review and submitting the work for approval?

lib/generators/sufia/templates/workflow.json.erb
```diff
@@ -31,6 +48,18 @@
                     "methods": [
                         "CurationConcerns::Workflow::ActivateObject"
                     ]
+                }, {
+                    "name": "request_review",
+                    "from_states": [{"names": ["changes_requested"], "roles": ["depositing"]}],
```
mjgiarlo: The state above is called `changes_required` and here it's `changes_requested`. Change this one to `changes_required`?

lib/generators/sufia/templates/workflow.json.erb
```diff
@@ -15,9 +15,26 @@
...
+                }, {
+                    "name": "request_changes",
+                    "from_states": [{"names": ["complete", "pending_review"], "roles": ["approving"]}],
+                    "transition_to": "changes_required",
+                    "notifications": [
+                        {
+                            "notification_type": "email",
+                            "name": "Sufia::Workflow::ChangesRequiredNotification",
+                            "to": ["approving"],
+                            "cc": []
```
mjgiarlo: I don't think the empty `cc` is adding anything. Remove?

lib/generators/sufia/templates/workflow.json.erb
```diff
@@ -31,6 +48,18 @@
...
+                }, {
+                    "name": "request_review",
+                    "from_states": [{"names": ["changes_requested"], "roles": ["depositing"]}],
+                    "transition_to": "pending_review",
+                    "notifications": [
+                        {
+                            "notification_type": "email",
+                            "name": "Sufia::Workflow::PendingReviewNotification",
+                            "to": ["approving"],
+                            "cc": []
```
mjgiarlo: Remove, pls.

spec/services/sufia/workflow/changes_required_notification_spec.rb
```diff
@@ -0,0 +1,28 @@
+require 'spec_helper'
+
+RSpec.describe Sufia::Workflow::ChangesRequiredNotification do
+  let(:depositor) { create(:user) }
+  let(:cc_user) { create(:user) }
+  let(:work) { create(:generic_work, user: depositor) }
+  let(:entity) { create(:sipity_entity, proxy_for_global_id: work.to_global_id.to_s) }
+  let(:comment) { double("comment", comment: 'A pleasant read') }
+  let(:recipients) { { 'to' => [depositor], 'cc' => [cc_user] } }
```
mjgiarlo: The `to` user is the one with the role `approving`, per the above workflow, not the depositor.

spec/services/sufia/workflow/changes_required_notification_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+
+  describe ".send_notification" do
+    it 'sends a message to all users' do
+      expect(depositor).to receive(:send_message).once.and_call_original
```
mjgiarlo: I would expect the approving user instance to receive the `:send_message` call as the user who initiates the state change.

spec/services/sufia/workflow/changes_required_notification_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+
+      expect { described_class.send_notification(entity: entity, user: depositor, comment: comment, recipients: recipients) }
```
mjgiarlo: I believe the `user` value here should be the approving user.

spec/services/sufia/workflow/changes_required_notification_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+        .to change { depositor.mailbox.inbox.count }.by(1)
+        .and change { cc_user.mailbox.inbox.count }.by(1)
```
mjgiarlo: `cc_user` should be the `to` user.

spec/services/sufia/workflow/changes_required_notification_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+    end
+    context 'without carbon-copied users' do
+      let(:recipients) { { 'to' => [depositor] } }
```
mjgiarlo: As above, the `to` user should be the one with the `approving` role per the declared workflow.

spec/services/sufia/workflow/changes_required_notification_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+      it 'sends a message to the to user(s)' do
+        expect(depositor).to receive(:send_message).once.and_call_original
+        expect { described_class.send_notification(entity: entity, user: depositor, comment: comment, recipients: recipients) }
```
mjgiarlo: Same comments on this block as the above.

PR #2947, opened 2016-11-29
Use strings instead of symbols for workflow notification recipients

### LINKED ISSUE #2943: undefined method `+' for nil:NilClass for AbstractNotification
A workflow json with notification configurations setup fails to email notifications due to invalid recipient keys (symbols vs string). 

```
def users_to_notify
  recipients[:to] + recipients[:cc]
end
```

In testing using stringified keys, another failure occurs unless both arrays contain at least a single user. 

Something like the following code would fix the issue and cases described above:

```
def users_to_notify
  recipients["to"] ||= []
  recipients["cc"] ||= []
  recipients["to"] + recipients["cc"]
end
```

mjgiarlo:
Because the underlying implementation apparently expects strings.

Fixes #2943
Refs #2944

### DIFF
--- a/app/services/sufia/workflow/abstract_notification.rb
+++ b/app/services/sufia/workflow/abstract_notification.rb
@@ -32,7 +32,7 @@ module Sufia
       private
 
         def users_to_notify
-          recipients[:to] + recipients[:cc]
+          recipients['to'] + recipients['cc']
         end
     end
   end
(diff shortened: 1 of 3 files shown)


app/services/sufia/workflow/abstract_notification.rb
```diff
@@ -32,7 +32,7 @@ def message
       private
 
         def users_to_notify
-          recipients[:to] + recipients[:cc]
+          recipients['to'] + recipients['cc']
```
revgum: My experience was that `recipients['cc']` being empty causes this to fail with...an exception I can't recall of the top of my head.

mjgiarlo: I don't think recipients['cc'] should ever be `nil`, @revgum.

mjgiarlo: @revgum To expound on that... the `cc` field is optional, and the JSON validator assumes `to` and `cc` (if present) are arrays: https://github.com/projecthydra/curation_concerns/blob/master/app/services/curation_concerns/workflow/workflow_schema.rb#L48

PR #2951, opened 2016-11-30
Do not use Admin Set-specified workflow if mediated deposit is disabled

### LINKED ISSUE #2949: Do not use Admin Set-specified workflow if mediated deposit is disabled
As a depositor, when I deposit a work and attach it to an admin set with a workflow configured, the work should not enter workflow unless `Flipflop.enable_mediated_deposit?` is `true`.

mjgiarlo:
Fixes #2949.

### DIFF
--- a/app/services/sufia/workflow/workflow_by_admin_set_strategy.rb
+++ b/app/services/sufia/workflow/workflow_by_admin_set_strategy.rb
@@ -7,7 +7,7 @@ module Sufia
 
       # @return [String] The name of the workflow by admin_set to use
       def workflow_name
-        return 'default' unless @admin_set_id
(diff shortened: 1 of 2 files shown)


app/services/sufia/workflow/workflow_by_admin_set_strategy.rb
```diff
@@ -7,7 +7,7 @@ def initialize(_work, attributes)
 
       # @return [String] The name of the workflow by admin_set to use
       def workflow_name
-        return 'default' unless @admin_set_id
+        return 'default' unless @admin_set_id && Flipflop.enable_mediated_deposit?
```
jcoyne: Do we need this feature flipper any more?  Can we say that the way to disable mediated deposit is to set the admin set to use the default (0-step) workflow?

PR #2953, opened 2016-11-30
Remove enable_mediated_deposit feature flipper.

### LINKED ISSUE #2952: Remove flippable option for Mediated Deposit
We no longer need Mediated Deposit to be a flippable option in Sufia, as the most recent thinking and work points towards workflows being "always on".

mjgiarlo:
This flipper is no longer needed now that Sufia ships with two usable workflows, and the ability to assign workflows to admin sets.

While doing so, remove or reduce (perhaps) unnecessary sleep statements in specs to prevent Travis hangs.

Fixes #2952


spec/features/collection_spec.rb
```diff
@@ -84,6 +84,10 @@
       visit '/dashboard/collections'
     end
 
+    it "has creation date for collections" do
```
jcoyne: why is this part of this PR?

PR #2955, opened 2016-12-01
Speed up feature specs and nuke the travis timeout

mjgiarlo:


mjgiarlo:
I've run twelve builds so far and all of them have passed, taking about the same amount of time as past builds. None of them have timed out. How about we give this a try for a while and see if this fixes the timeout issues on Travis?

spec/spec_helper.rb
```diff
@@ -48,6 +48,21 @@ def coverage_needed?
 Capybara.default_driver = :rack_test      # This is a faster driver
 Capybara.javascript_driver = :poltergeist # This is slower
 Capybara.default_max_wait_time = ENV['TRAVIS'] ? 30 : 15
+# Adding the below to deal with random Capybara-related timeouts in CI.
+# Found in this thread: https://github.com/teampoltergeist/poltergeist/issues/375
+poltergeist_options = {
+  js_errors: false,
```
awead: While this will squelch the false-positives, some of the js errors are legitimate. This has bitten us several times and we've now set it to `true` because there were errors we weren't seeing until we had deployed the app to our QA environment.

awead:
I tweaked our batch_edit spec test recently:

https://github.com/psu-stewardship/scholarsphere/blob/develop/spec/features/batch_edit_spec.rb

One thing I did was enable a specific js bit that prevents animations for happening. This was causing a hangup because the test was waiting for something to open:

https://github.com/psu-stewardship/scholarsphere/blob/develop/spec/features/support/feature_sessions.rb#L21

mjgiarlo:
@awead I'd be happy to make those changes as part of this branch.

mjgiarlo:
@awead ok, I made the `js_errors` change, but I don't think I want to pull in the batch_edit feature spec changes. If we still run into timeout issues, we should revisit your work in ScholarSphere, though! Thanks for sharing that.

awead:
@mjgiarlo np. hope it helps!

mjgiarlo:
(Self-merging because already approved.)

PR #2959, opened 2016-12-02
Give a file its original filename instead of the interstitial one we use

### LINKED ISSUE #2948: filename of download is taken from the tmp filename, not the previous value
After migrating if you download the file, it has a different filename (_version1) than it did previously. This is either the filename or label property being set by characterization I guess? instead of by metadata.

hackmastera:
during import, fixes #2948


hackmastera:
okay that test failure doesn't look right but there's a mistake in this PR anyway so don't merge yet!!

hackmastera:
this is fixed / ready

PR #2964, opened 2016-12-07
Admin abilities

### LINKED ISSUE #2946: Admins should be able to edit Admin Set w/o ACL
As an administrative user, I should be able to edit any Admin Set in the system regardless of its *other* `role_map.yml` ACLs.  

Split from #2902.

atz:
Grant admin abilities on AdminSets (and related models).

Fixes #2946

### DIFF
--- a/app/models/concerns/sufia/ability.rb
+++ b/app/models/concerns/sufia/ability.rb
@@ -63,7 +63,7 @@ module Sufia
     end
 
     def admin_set_abilities
-      can :create, AdminSet if admin?
+      can :manage, [AdminSet, Sufia::PermissionTemplate, Sufia::PermissionTemplateAccess] if admin?
 
       can [:create, :edit, :update, :destroy], Sufia::PermissionTemplate do |template|
         test_edit(template.admin_set_id)
--- a/spec/models/sufia/ability_spec.rb
+++ b/spec/models/sufia/ability_spec.rb
@@ -38,16 +38,17 @@ describe Sufia::Ability, type: :model do
     it { is_expected.to be_able_to(:read, ContentBlock) }
     it { is_expected.to be_able_to(:read, Sufia::Statistics) }
     it { is_expected.to be_able_to(:read, :admin_dashboard) }
-    it { is_expected.not_to be_able_to(:manage, AdminSet) }
+    it { is_expected.to be_able_to(:manage, AdminSet) }
     it { is_expected.to be_able_to(:create, AdminSet) }
   end
 
   describe "AdminSets and PermissionTemplates" do
     let(:permission_template) { build(:permission_template, admin_set_id: admin_set.id) }
     let(:permission_template_access) { build(:permission_template_access, permission_template: permission_template) }
-    describe "a user with edit access" do
-      let(:user) { create(:user) }
-      let(:admin_set) { create(:admin_set, edit_users: [user]) }
+    let(:user) { create(:user) }
+    let(:admin_set) { create(:admin_set) }
+
+    RSpec.shared_examples 'A user with additional access' do
       it { is_expected.to be_able_to(:edit, admin_set) }
       it { is_expected.to be_able_to(:update, admin_set) }
       it { is_expected.to be_able_to(:destroy, admin_set) }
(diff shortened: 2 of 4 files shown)


spec/models/sufia/ability_spec.rb
```diff
@@ -38,16 +38,35 @@
...
   describe "AdminSets and PermissionTemplates" do
     let(:permission_template) { build(:permission_template, admin_set_id: admin_set.id) }
     let(:permission_template_access) { build(:permission_template_access, permission_template: permission_template) }
-    describe "a user with edit access" do
-      let(:user) { create(:user) }
+    let(:user) { create(:user) }
+    let(:admin_set) { create(:admin_set) }
+
+    describe 'as admin' do
+      before { expect(ability).to receive(:admin?).and_return(true).at_least(1).times }
```
jcoyne: This is probably causing your problem. The call to `ability` is invoking `Ability.new` (and all the methods that grant access) before the mock can be created.  Instead, try mocking on `RoleMap`

spec/models/sufia/ability_spec.rb
```diff
@@ -38,16 +38,35 @@
...
+      it 'is an admin' do
+        expect(ability.admin?).to be_truthy
+        expect(subject.admin?).to be_truthy
```
jcoyne: The subject is the `ability`, right?  These are duplicates?

atz: Yes, just sanity checks.

spec/models/sufia/ability_spec.rb
atz: I see one oblique reference to `RoleMapper` in a spec.  Looks like it used to be Sufia, but currently is not:
https://docs.omniref.com/ruby/gems/sufia/3.6.0/symbols/RoleMapper

Also not in CurationConcerns.  What is it?

mjgiarlo: @atz it's a hydra-access-controls thing. See [here](https://github.com/projecthydra/hydra-head/blob/7a911e54a7cc0c49ca0e9cc8da1c6d47f9a467bb/hydra-access-controls/app/models/role_mapper.rb) (particularly for the comments) and [here](https://github.com/projecthydra/hydra-head/blob/7a911e54a7cc0c49ca0e9cc8da1c6d47f9a467bb/hydra-access-controls/lib/hydra/role_mapper_behavior.rb) (for the code).

app/models/concerns/sufia/ability.rb
```diff
@@ -63,7 +63,9 @@ def feature_abilities
     end
 
     def admin_set_abilities
-      can :create, AdminSet if admin?
+      can :manage, AdminSet if admin?
```
mjgiarlo: What would you think about batching these up? Like:

```ruby
can :manage, [AdminSet, Sufia::PermissionTemplate, Sufia::PermissionTemplateAccess] if admin?
```

Or put all three abilities in an `if admin?` block?

spec/models/sufia/ability_spec.rb
mjgiarlo: Could you use RSpec's predicate matchers here? `expect(ability).to be_admin` We may have a smattering of `truthy`s in the codebase but I have a preference for predicate matchers where they are usable.

spec/models/sufia/ability_spec.rb
```diff
@@ -38,16 +38,35 @@
...
+      it { is_expected.to be_able_to(:update, admin_set) }
+      it { is_expected.to be_able_to(:destroy, admin_set) }
+      it { is_expected.to be_able_to(:create, permission_template) }
+      it { is_expected.to be_able_to(:create, permission_template_access) }
+    end
+
+    describe 'via AdminSet-specific edit_users' do
       let(:admin_set) { create(:admin_set, edit_users: [user]) }
+      it 'is not admin' do
+        expect(ability.admin?).to be_falsey
```
mjgiarlo: Same as above re: predicate matchers.

tasks/sufia-dev.rake
```diff
@@ -1,7 +1,6 @@
-require 'rspec/core'
 require 'rspec/core/rake_task'
-require 'solr_wrapper'
-require 'fcrepo_wrapper'
+require 'solr_wrapper'   # necessary for rake_support to work
```
mjgiarlo: Good comments!

app/models/concerns/sufia/ability.rb
atz: sure, can combine.

tasks/sufia-dev.rake
mjgiarlo: @atz :beers:

mjgiarlo:
@jcoyne I believe @atz has addressed your changes above. Re-review?

PR #2965, opened 2016-12-07
Add a default admin set

jcoyne:

This reduced conditionals, as now every newly created or updated work
will have an admin set.  This is required because the AdminSet
determines which workflow to use.  The Default admin set specifies the
default (direct deposit) workflow.

Ref projecthydra-labs/hyku#481 (needs port to Hyrax)

### DIFF
--- a/spec/actors/sufia/default_admin_set_actor_spec.rb
+++ b/spec/actors/sufia/default_admin_set_actor_spec.rb
@@ -1,0 +1,43 @@
+require 'spec_helper'
+
+RSpec.describe Sufia::DefaultAdminSetActor do
+  let(:next_actor) do
+    double('next actor', create: true,
+                         curation_concern: work,
+                         update: true,
+                         user: depositor)
+  end
+  let(:actor) do
+    CurationConcerns::Actors::ActorStack.new(work, depositor, [described_class])
+  end
+  let(:depositor) { create(:user) }
+  let(:work) { build(:generic_work) }
+  let(:admin_set) { create(:admin_set) }
+  let(:permission_template) { create(:permission_template, admin_set_id: admin_set.id) }
+
+  describe "create" do
+    before do
+      allow(CurationConcerns::Actors::RootActor).to receive(:new).and_return(next_actor)
+    end
+
+    context "when admin_set_id is blank" do
+      let(:attributes) { { admin_set_id: '' } }
+      let(:default_id) { described_class::DEFAULT_ID }
+
+      it "creates the default AdminSet with a PermissionTemplate and calls the next actor with the default admin set id" do
+        expect(next_actor).to receive(:create).with(admin_set_id: default_id).and_return(true)
+        expect { actor.create(attributes) }.to change { AdminSet.count }.by(1)
+          .and change { Sufia::PermissionTemplate.count }.by(1)
+      end
+    end
+
+    context "when admin_set_id is provided" do
+      let(:attributes) { { admin_set_id: admin_set.id } }
+
+      it "uses the provided id and returns true" do
+        expect(next_actor).to receive(:create).with(attributes).and_return(true)
+        expect(actor.create(attributes)).to be true
+      end
+    end
+  end
+end
--- a/app/actors/sufia/default_admin_set_actor.rb
+++ b/app/actors/sufia/default_admin_set_actor.rb
@@ -1,0 +1,42 @@
+module Sufia
+  # Ensures that the default AdminSet id is set if this form doesn't have
+  # an admin_set_id provided. This should come before the
+  # CurationConcerns::Actors::InitializeWorkflowActor, so that the correct
+  # workflow can be kicked off.
+  class DefaultAdminSetActor < CurationConcerns::Actors::AbstractActor
+    def create(attributes)
+      ensure_admin_set_attribute!(attributes)
+      next_actor.create(attributes)
+    end
+
+    def update(attributes)
+      ensure_admin_set_attribute!(attributes)
+      next_actor.update(attributes)
+    end
+
+    private
+
+      def ensure_admin_set_attribute!(attributes)
+        return if attributes[:admin_set_id].present?
+        attributes[:admin_set_id] = default_admin_set_id
+      end
+
+      DEFAULT_ID = 'admin_set/default'.freeze
+
+      def default_admin_set_id
+        create_default_admin_set unless default_exists?
+        DEFAULT_ID
+      end
+
+      def default_exists?
+        AdminSet.exists?(DEFAULT_ID)
+      end
+
+      # Creates the default AdminSet and an associated PermissionTemplate with workflow
+      def create_default_admin_set
+        AdminSet.create!(id: DEFAULT_ID, title: ['Default Admin Set']).tap do |_as|
+          PermissionTemplate.create!(admin_set_id: DEFAULT_ID, workflow_name: 'default')
+        end
+      end
+  end
+end
--- a/app/services/sufia/actor_factory.rb
+++ b/app/services/sufia/actor_factory.rb
@@ -9,6 +9,7 @@ module Sufia
        CurationConcerns::Actors::AttachFilesActor,
        CurationConcerns::Actors::ApplyOrderActor,
        CurationConcerns::Actors::InterpretVisibilityActor,
+       DefaultAdminSetActor,
        CurationConcerns::Actors::InitializeWorkflowActor,
        ApplyPermissionTemplateActor,
        model_actor(curation_concern)]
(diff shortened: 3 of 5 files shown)


mjgiarlo:
@jcoyne We should probably discuss this on a Hydra Tech call or, at least, tag a bunch of folks to comment on this change.

Calling all @projecthydra/sufia-code-reviewers! :arrow_up:

mjgiarlo:
FWIW, I am :+1: on this PR but am hesitant to approve it until we've had a chance to hear from folks. Tagging @cam156 @awead @escowles @tpendragon @grosscol @hackmastera @jeremyf @newmanld @njaffer @elrayle

newmanld:
Having a default workflow – which implies a default admin set – does not seem onerous.  Seems to make explicit what is implicit.

mjgiarlo:
@newmanld Yes, right now our default workflow strategy goes like this:

* Does the work belong to an admin set? If so, use its workflow; else, use the default admin set.

So instead of checking if a work belongs to an admin set, we will know it belongs to one (always, without having to check across the whole codebase) and will merely say "use the workflow configured for this work's admin set." Thanks, Linda!

atz:
The arguments against a default AdminSet came during PCDM discussions.  Specifically, the problem was that because the expression of that relationship would have been **LDP Contains**, when a user went to adopt a non-default AdminSet, it would require changing the (URI) identity of all objects to move them into the new AdminSet.  

We can happily say that problem is not part of the Sufia implementation (i.e. isn't LDP Contains).

jeremyf:
On the surface, an explicit "null" AdminSet makes a lot of sense. But @atz's perspective is good to consider: the "null" set should not be persisted. When we reify from Fedora into Ruby we may want to consider the "null" AdminSet.

atz:
Persistence is not really important either way to my point.  Since we have not defined the relationship as **LDP Contains**, it is just a relationship to members based on `RDF::Vocab::DC.isPartOf`.

It is fine to have that object be in Fedora and base Fedora queries on it.

mjgiarlo:
It's been five days and we've heard from Cincinnati and Notre Dame so I am inclined to merge this PR soon. Any comments, @cam156 @awead @escowles @tpendragon @grosscol @hackmastera @njaffer @elrayle ?

cam156:
@mjgiarlo My only concern is how do we update all existing works to include the default.  It would be nice to have a rake task included to do that work.  That should not block this PR.  The other question I have is how does this work when admin sets are turned off via flipflop?

jcoyne:
@cam156 why would you want to update all existing works?  The admin set is of little use after the work is deposited.

cam156:
@jcoyne The way I was reading this is that you want a default admin set so that other code in the system does not need to check if the item is part of an admin set.  My assumption is that there are displays that need to know the admin set and would fail if it is not set.  For that reason I would think having some of the items in your repository part of an admin set and some of them not would defeat the purpose of the default admin set.

jcoyne:
@cam156 Currently the only thing that needs an admin set is the workflow.  The workflow has been completely re-tooled in CC 1.7.0, such that the workflow you use depends on the admin set.  As none of the existing objects need workflow, they don't need an admin set either.

spec/actors/sufia/default_admin_set_actor_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+  describe "create" do
+    before do
+      allow(CurationConcerns::Actors::RootActor).to receive(:new).and_return(create_actor)
+    end
+
+    context "when admin_set_id is blank" do
+      let(:attributes) { { admin_set_id: '' } }
+      let(:default_id) { described_class::DEFAULT_ID }
+
+      it "creates the default AdminSet with a PermissionTemplate and calls the next actor with the default admin set id" do
```
jeremyf: The variable name is `create_actor`, but the spec says `next actor`. Can you reconcile the language? I find the "next actor" to be a bit more intuitive (and it reconciles with the variable defined in the above class `Sufia:: DefaultAdminSetActor`).

spec/actors/sufia/default_admin_set_actor_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+        expect(create_actor).to receive(:create).with(admin_set_id: default_id).and_return(true)
+        expect { actor.create(attributes) }.to change { AdminSet.count }.by(1)
+          .and change { Sufia::PermissionTemplate.count }.by(1)
+      end
+    end
+
+    context "when admin_set_id is provided" do
+      let(:attributes) { { admin_set_id: admin_set.id } }
+
+      it "uses the provide id and returns true" do
```
jeremyf: I'm not certain of the intention of the specification's description. The description `"uses the provide id and returns true"` does not appear to be asserted in the spec's body.

hackmastera:
Isn't admin set facetable, also?  That would seem to imply that it's otherwise an organizing mechanism in some way.

I think if a rake task is required to assign existing objects to a default admin set that would be an appropriate reason to hold up the PR.

spec/actors/sufia/default_admin_set_actor_spec.rb
jcoyne: It's the "next actor used for creating"

spec/actors/sufia/default_admin_set_actor_spec.rb
jcoyne: @jeremyf the id in question (`admin_set_id`) is in the `attributes` hash.

jcoyne:
@hackmastera you *could* facet on it, but it isn't going to unless you change your Blacklight config.

jcoyne:
@hackmastera we never said that every work should have an admin set. We said every newly created or updated work should have an admin set.  This is only important for workflow.

hackmastera:
How are existing objects exempt from workflow? We have plenty of partially-completed objects already in the repository; probably about 40% of our stuff.

hackmastera:
You know what this PR is really small and is 100% additions. How did it reduce conditionals?

jcoyne:
> How are existing objects exempt from workflow? We have plenty of partially-completed objects already in the repository; probably about 40% of our stuff.

You don't want to have to mediate all the existing objects right?  Only the ones someone updates.

jcoyne:
> You know what this PR is really small and is 100% additions. How did it reduce conditionals?

It reduces conditionals that we would have to build.  Prior to this pull request, if there was no workflow on the form (Flipflop option), we would have to trap for that condition, and then hard code the workflow to use in this situation (e.g. "direct deposit").  Of course some people would want the workflow in this situation to be "mediated deposit".  Using the default admin set, we just store the workflow to use on the admin set and avoid the need for a special variable like "which workflow to use if they don't have an admin set selected"

cam156:
@jcoyne look at your description for this PR.  That is where I got the idea you were trying to reduce conditionals. All I am saying is if a person writing code sees the default admin set they are going to expect that all Works will either be in that or have their own.  As someone who will have to migrate to this code eventually I can see this opening up a Can O' Worms if we do not take that stance.  What would make me happy is if you create a ticket that needs to be closed before release, or create the rake task in this PR, so that the rake task can be referenced when people migrate to the new code.

jcoyne:
I'd be fine creating a new ticket. But I don't think that is productive work.  The admin set is only useful when you are creating or updating a work (at which time the default comes into play).

jcoyne:
I created #2977, but left a disclaimer on it about needing to determine the business value first.

hackmastera:
Reporting seems like a pretty obvious place where a conditional will still be in place without migration here. How many objects are in each admin set? What is the percentage for each admin set that is completely processed / published?

jcoyne:
@hackmastera we don't have any reports by admin set at this juncture.

hackmastera:
My opinion is that the correct time to provide an upgrade path is when adding the default. We clearly disagree. I'd be glad for a 3rd party to step in and decide at this point. @mjgiarlo?

jcoyne:
@hackmastera every task should be backed up with a story.  In this case:

> As a repository manager, I want all my objects to be added to an admin set so that ______________.

What goes in the blank?

hackmastera:
@jcoyne i added a use case above. Sorry I wasn't clear before. I am done now and request mediation. thanks.

jeremyf:
I'm in a meeting right now, but will review. However, @mjgiarlo could you review @hackmastera and @jcoyne's conversation to break out steps to move forward.

mjgiarlo:
@hackmastera @jcoyne @jeremyf Happy to outline a plan. Give me a few minutes to work it out.

mjgiarlo:
Do folks think this will suffice? https://github.com/projecthydra/sufia/pull/2978

@jcoyne wrote that up when I was asking him questions. ;)

hackmastera:
Only if @jcoyne also says "@hackmastera was right"

Just kidding; looks great. Thanks, y'all.

mjgiarlo:
Thank you all!

PR #2968, opened 2016-12-12
Pin curation_concerns to 1.6.0

jcoyne:
CurationConcerns 1.7.0 causes the build to fail.


jcoyne:
Sufia requires some of the workflow stuff in CC 1.7.0, so closing this an adding: #2969

PR #2970, opened 2016-12-13
Make select_work modal batch-aware

### LINKED ISSUE #2913: When you start a batch create you should be prompted for work type
When you click "Works" -> "Batch Create"  a modal should pop up offering you a choice of which type of work to create.  This should be similar to when you choose "Works" -> "New Work"

atz:
Advances, but does not resolve #2913 

Several tweaks to the modal:
- Translate the aria-label ("Close").
- Remove the useless button at the bottom.
- Add "Batch Create" buttons per type.

It also removes separate Batch Create dropdown in the toolbar, since now
the same flow (via modal) is used for single or batch uploads, when more than one object model is present.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -100,3 +100,6 @@ RSpec/NestedGroups:
 
 RSpec/LeadingSubject:
   Enabled: false
+
+RSpec/MessageSpies:
+  Enabled: false
--- a/app/controllers/concerns/sufia/batch_uploads_controller_behavior.rb
+++ b/app/controllers/concerns/sufia/batch_uploads_controller_behavior.rb
@@ -6,24 +6,33 @@ module Sufia
 
     included do
       self.work_form_service = BatchUploadFormService
-      self.curation_concern_type = work_form_service.form_class.model_class
+      self.curation_concern_type = work_form_service.form_class.model_class # includes CanCan side-effects
+      # We use BatchUploadItem as a null stand-in curation_concern_type.
+      # The actual permission is checked dynamically via `authorize!` during #create.
     end
 
+    # The permissions to create a batch are not as important as the permissions for the concern being batched.
     def create
       authenticate_user!
+      authorize! :create, params[:batch_upload_item][:payload_concern].constantize
       create_update_job
       flash[:notice] = t('sufia.works.new.after_create_html', application_name: view_context.application_name)
       redirect_after_update
     end
 
-    # Gives the class of the form.
     class BatchUploadFormService < CurationConcerns::WorkFormService
-      def self.form_class(_ = nil)
+      # Gives the class of the form.
+      def self.form_class(_curation_concern = nil)
         ::Sufia::Forms::BatchUploadForm
       end
     end
 
     protected
+
+      def build_form
+        super
+        @form.payload_concern = params[:payload_concern]
+      end
 
       def redirect_after_update
         if uploading_on_behalf_of?
--- a/app/controllers/concerns/sufia/batch_uploads_controller_behavior.rb
+++ b/app/controllers/concerns/sufia/batch_uploads_controller_behavior.rb
@@ -8,14 +8,19 @@ module Sufia
       self.work_form_service = BatchUploadFormService
       self.curation_concern_type = work_form_service.form_class.model_class # includes CanCan side-effects
       # We use BatchUploadItem as a null stand-in curation_concern_type.
-      # The actual permission is checked dynamically via `authorize!` during #create.
+      # The actual permission is checked dynamically during #create.
     end
 
     # The permissions to create a batch are not as important as the permissions for the concern being batched.
+    # @note we don't call `authorize!` directly, since `authorized_models` already checks `user.can? :create, ...`
     def create
       authenticate_user!
-      authorize! :create, params[:batch_upload_item][:payload_concern].constantize
-      create_update_job
+      unsafe_pc = params[:batch_upload_item][:payload_concern]
+      # Calling constantize on user params is disfavored (per brakeman), so we sanitize by matching it against an authorized model.
+      safe_pc = Sufia::SelectTypeListPresenter.new(current_user).authorized_models.map(&:to_s).find { |x| x == unsafe_pc }
+      raise CanCan::AccessDenied, "Cannot create an object of class '#{unsafe_pc}'" unless safe_pc
+      authorize! :create, safe_pc
+      create_update_job(safe_pc.constantize)
       flash[:notice] = t('sufia.works.new.after_create_html', application_name: view_context.application_name)
       redirect_after_update
     end
@@ -42,7 +47,8 @@ module Sufia
         end
       end
 
-      def create_update_job
+      # @param [Class] klass the Sufia Work Class being created by the batch
+      def create_update_job(klass)
         log = BatchCreateOperation.create!(user: current_user,
                                            operation_type: "Batch Create")
         # ActionController::Parameters are not serializable, so cast to a hash
@@ -50,7 +56,7 @@ module Sufia
                                      params[:title].permit!.to_h,
                                      params[:resource_type].permit!.to_h,
                                      params[:uploaded_files],
-                                     attributes_for_actor.to_h,
+                                     attributes_for_actor.to_h.merge!(model: klass),
                                      log)
       end
 
--- a/app/controllers/concerns/sufia/batch_uploads_controller_behavior.rb
+++ b/app/controllers/concerns/sufia/batch_uploads_controller_behavior.rb
@@ -15,12 +15,12 @@ module Sufia
     # @note we don't call `authorize!` directly, since `authorized_models` already checks `user.can? :create, ...`
     def create
       authenticate_user!
-      unsafe_pc = params[:batch_upload_item][:payload_concern]
+      unsafe_pc = params.fetch(:batch_upload_item, {})[:payload_concern]
       # Calling constantize on user params is disfavored (per brakeman), so we sanitize by matching it against an authorized model.
       safe_pc = Sufia::SelectTypeListPresenter.new(current_user).authorized_models.map(&:to_s).find { |x| x == unsafe_pc }
       raise CanCan::AccessDenied, "Cannot create an object of class '#{unsafe_pc}'" unless safe_pc
-      authorize! :create, safe_pc
-      create_update_job(safe_pc.constantize)
+      # authorize! :create, safe_pc
+      create_update_job(safe_pc)
       flash[:notice] = t('sufia.works.new.after_create_html', application_name: view_context.application_name)
       redirect_after_update
     end
@@ -47,7 +47,8 @@ module Sufia
         end
       end
 
-      # @param [Class] klass the Sufia Work Class being created by the batch
+      # @param [String] klass the name of the Sufia Work Class being created by the batch
+      # @note Cannot use a proper Class here because it won't serialize
       def create_update_job(klass)
         log = BatchCreateOperation.create!(user: current_user,
                                            operation_type: "Batch Create")
--- a/app/controllers/concerns/sufia/works_controller_behavior.rb
+++ b/app/controllers/concerns/sufia/works_controller_behavior.rb
@@ -9,8 +9,8 @@ module Sufia
     end
 
     module ClassMethods
-      # We don't want the breadcrumb action to occur until after the concern has
-      # been loaded and authorized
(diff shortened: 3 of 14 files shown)


jcoyne:
If this is the UI for this PR, I have a couple of concerns:

screen shot 2016-12-12 at 3 46 01 pm

First of all, there's a big call out for "Batch works" (which we sort of want to discourage) and no call out for "Create single work with multiple files."

Secondly, you've removed the "Batch create" from the drop down, which is something the UI working group worked on.  

Is there some reason you can't select "Batch create" and then see this modal for picking a type? This would look just like how the single create work is implemented now.

atz:
I don't understand what you think we want to discourage.  Use of batch features?  "Callouts"?  Buttons?

There is now literally no difference between the two drop-down menu options' behavior.  That seems useless and bad design.  The design originated from when the expectation was that the behaviors would be different.  That is no longer true.

Also, see the ticket where I already call out the link vs. button-ness.  I would be fine doing two buttons, if you want.  It was obvious to me that further attention to the modal would be needed from design, considering there were: useless buttons, defaulty icons, defaulty description text; New button(s) could be vertically centered; etc.  I am not a bootstrap designer, and I'm unlikely to become one for the purposes of this PR.

atz:
@jcoyne: see above.  You want 2 buttons or no buttons or something else?

mjgiarlo:
2 buttons++

jcoyne:
@atz I defer to @ggeisler

ggeisler:
Is this the same Select Work Type modal that was designed in https://github.com/projecthydra/sufia/issues/2434?

(I was not involved in any design related to batch upload so want to make sure I understand what is going on before commenting further.)

jcoyne:
@ggeisler yes, we need that modal available for both single item and batch upload.

ggeisler:
Okay, then I have a several concerns:

* I don't think the issues @hannahfrost and I had with the implementation of the single item upload work type selection have been addressed. Ticket https://github.com/projecthydra/sufia/issues/2434 is the original design. Ticket https://github.com/projecthydra-labs/hyku/issues/496 documents the QA feedback.

* The requirement for work type selection in a batch upload scenario was never indicated as "design needed" and was never on my radar as something that needed design (let me know if I just missed it, but I definitely don't recall anyone telling me this needed design).

Because of the points above, it feels like we are trying to tack on undesigned functionality to an existing feature, which itself was not completed to match the intended design. So it's difficult for me to weigh in on the discussion above, since I'm not sure I'm on board with the premise that bulk upload should be tacked onto the single item upload UI. The select work type modal approach was definitely not designed with the idea that we'd just add a bulk upload button to it (maybe that will end up being the way we go, but that wasn't part of the requirements I had in front of me when designing the solution in #2434).

hannahfrost:
@ggeisler You are right that requirements for designing a batch upload modal have not been provided to you. 

My two cents at this point:

- Keep Batch Create in the menu
- When it selected, display a modal that looks like the one in #2434, but the blue button is labeled "Create batch"
- Oh and let's address #496 (which is now in the ready column)

atz:
@ggeisler, @hannahfrost: you seem to be missing the context from the related issue #2913. I have been calling for design attention since last week.  There are technical reasons that the existing pattern (dropdown-specific always-rendered hidden modal) is bad, but duplicating it would be worse.  We want to minimize the number of components we render hidden on every page.  This is not intended to implement Gary's design and indeed seems unable to reuse much of it, since that implementation is unfinished.  

I proposed a handful of alternatives.  Consensus was to reuse the modal for both single and batch.  That conversation was prerequisite to my starting work.  

To restate the question i posed to @jcoyne: 
> I don't understand what... we want to discourage. Use of batch features? "Callouts"? Buttons?

mjgiarlo:
@ggeisler @jcoyne @hannahfrost @atz I think this PR should go forward without significant UI tweaks, and then we should tackle #496 in a way that pulls the UI changes @ggeisler mocked up into both the single item and the batch modal. How does that sound?

atz:
There will only be one modal when this is done.  That is the *less* significant "UI tweak".  The ticket I am working from never had #2434 included as scope or context, so I am happy to defer that revision to a subsequent pass. 

Note: projecthydra-labs/hyku#496 is not #496 (a PR that has been closed for years).

mjgiarlo:
That sounds right to me, @atz.

ggeisler:
Just for the record, @atz's statement above that:

> There will only be one modal when this is done.

is a design decision. Not a code design decision, but a UX design decision. Since @mjgiarlo seems to agree with it, I won't argue, but I did not design the Select work type modal to support both the single- and batch-upload scenarios. (If we are making quick, on-the-spot decisions about this design, I'd say independent modals--even if they share the same underlying view--is a better solution than sticking buttons into the original design.)

(Also for the record: neither @hannahfrost nor I were tagged on this ticket until 20 hours ago, there has never been a "design needed" label attached to this ticket, and no one has ever asked me at standup to look this issue.)

mjgiarlo:
@ggeisler I was thinking that @atz was on the right track because it will allow batch creation to work with multiple model types, and it uses the current UI we have. It also gives us time to come up with a better UX design for the batch creation modal -- which, as you pointed out, fell between the cracks (sorry!) -- that aligns with the already slick single item creation modal ticketed in projecthydra-labs/hyku#496. Sounds like we have a ticket to write up?

Does this sound like a practical approach that balances dev needs with UX needs? @ggeisler @hannahfrost @atz ?

atz:
I agree it was a design decision, which is why I continually raised it until I got enough feedback to go on.  Gary, I don't understand your differentiation of "independent modals, same view."  Maybe we are saying the same thing.  I agree the view (template) is the part I care about not duplicating.  If different contexts dynamically alter that or present it differently, no problem.  

The preceding ticket is #2913.  Many questions have come up there; some have received answers.  

Believe it or not, the UI questions are the easier part of this problem.

ggeisler:
@mjgiarlo Sure, if it is a temporary solution just to get batch upload working with multiple works types, then no problem. We should probably be very clear that it is a temporary solution, though (in demos or whatnot), because we sought and received input from a handful of Sufia people on the design; I don't want those people to think we asked for their feedback on a feature and then changed it on them (which is what **I think** when I see the WIP screenshot).

@atz Again, this is without dedicating time to fully look at the issue (wasn't on my radar and I'm reluctant to push aside other things I'm behind on to deal with it), but what I was suggesting was that a user would select one menu item to trigger the Select work type modal for a single-item upload, and that there is a different menu item that the user would select when that user wanted to do a batch upload. The batch upload menu item might trigger the display of a modal very much like, if not identical to, the one designed for single item upload, but it is not the same modal (in that it doesn't have extra buttons where the user has to distinguish between single- and batch-upload; they've already done that with the menu item they choose).

So, two distinct menu items to add works (single and multiple), and then a modal from which to select the work type. It doesn't matter to me whether that modal is exactly the same, code-wise. Just that UI-wise, it looks like it does in the original design, without extra buttons.

atz:
Yeah, that makes sense.  If the modal worked the way you intended, then it would be more adaptable to this purpose.  Since it is currently single-click link based, that functionality isn't there to reuse.

jcoyne:
> The requirement for work type selection in a batch upload scenario was never indicated as "design needed" and was never on my radar as something that needed design (let me know if I just missed it

@ggeisler https://github.com/projecthydra-labs/hyku/issues/165 😉

ggeisler:
@jcoyne Are you seriously suggesting that a ticket titled "Design the whole process," which has no description aside from a reference to a ticket about new account creation, has something to do with selecting a work type in a batch upload scenario?

mjgiarlo:
@ggeisler I'm sure he was joking!

jcoyne:
Yep joking, that's why I put the wink in there.

mjgiarlo:
:speech_balloon: @ggeisler 
> Sure, if it is a temporary solution just to get batch upload working with multiple works types, then no problem. We should probably be very clear that it is a temporary solution, though (in demos or whatnot), because we sought and received input from a handful of Sufia people on the design

:+1:

hannahfrost:
@atz @jcoyne @mjgiarlo I have been chatting with @ggeisler about this business with the modal (mostly with respect to https://github.com/projecthydra-labs/hyku/issues/496). We are rethinking the modal at this point, as the effort to get it to work as designed is greater than the value. Gary is suggesting an alternate design with an intermediary page for selecting the work type before getting to the create work page. Wanted to alert @atz right away.

atz:
The modal is truly not the meat of this PR: it is the easiest piece and it has been done for weeks.  Rethink whatever you please, it will not affect this work at all.

atz:
Testing is blocked by #2999

mjgiarlo:
@atz Apologies for the twists and turns on this one, Joe. Here's a note about arriving at the final design for this feature: https://github.com/projecthydra-labs/hyku/issues/580#issuecomment-272035060 Let me know if you have any questions about that -- happy to chat about this.

atz:
Obviously, I might do some squashing on what is only a 250-line delta, but I'd also be happy just to get this over with.

mjgiarlo:
@atz Now that we have consensus on a UI for this interaction, could you tweak your PR to use the following UI? https://github.com/projecthydra-labs/hyku/issues/580#issuecomment-272035060 In short, we will:

1. Retain the "Batch Create" menu link
2. (With *n* > 1 concerns generated) Pop up a modal when clicked, just like when "New Work" is clicked.
3. Display a list of concern labels and descriptions, each with a radio button.
4. Display a "Cancel" button and a "Create Batch" button

Like I said above, sorry for the gyrations on this issue.

atz:
Absolutely not.  This gets in first, and *then* we do the other interface as a follow-on.

mjgiarlo:
@atz OK, thanks for writing up your thoughts. Let's discuss in our stand-up today, after everything else.

atz:
@jcoyne Added the renaming of BatchUploadItem.  Otherwise the same.

jcoyne:
I don't agree with the last PR.  The BatchUpload (which is not persisted and is purely conceptual) is the set of BatchUploadItems.

atz:
@jcoyne The renaming was [previously discussed](https://github.com/projecthydra/sufia/issues/2913#issuecomment-266927897).  It is not persisted, but the naming convention is still applicable.  I'm saying the BatchUpload is the set of future objects.  That is how we were already using the class.  When we post `:create` to the BatchUploadsController, we are instantiating one such set from `uploaded_files`.  AFAICT, we never were creating a set of `BatchUploadItem`s, one from each of uploaded file.

jcoyne:
@atz if you need to change it, you should change it the other way (e.g. `BatchUploadItemsController`) because the model represents one part (an item) of the batch upload (a collection).

mjgiarlo:
@atz Can we drop the last commit so we can include this work in Sufia 7.3.0? I'd like to avoid API breaking changes in Sufia master. If you do that, we can get this merged this morning. I'd be happy to write up a new issue about fixing this in Hyrax 1.0.0.

atz:
@jcoyne I don't find that interpretation reflected in the code.  The `BatchUploadsController` needs is a thing to control: a batch upload.  In practice, we have called that thing a "BatchUploadItem".  AFAICT, no part of the code needs a (separate, temporary) model for each individual `uploaded_files` file.  Correct me if I'm missing something.  

The lifecycle of such an object has two parts:
- before object creation the `uploaded_files` file file data is handed off to an async job with the model to be created.  
- after creation, the object uses the already existing controller for that model

I understand your impression that the model would map to a single item, but technically I don't think that is now or ever has been the case.

atz:
While it is a class rename, I also fundamentally disagree that this is a serious API breaking change, mainly because I can't see how anybody would have done anything useful with this null model class in the first place, unless it was to unbreak batch uploads with multiple concerns defined, which this PR accomplishes.  The point of the class was to **do nothing**.  The idea that somebody has extended it seems... far fetched.  

To be clear, I included it in this work, @mjgiarlo, because you explicitly told me to include it.  That being said, yes, I can split off a separate branch for that one commit, obviously.

mjgiarlo:
Thank you, @atz.

hackmastera:
> While it is a class rename, I also fundamentally disagree that this is a serious API breaking change, mainly because I can't see how anybody would have done anything useful with this null model class in the first place, unless it was to unbreak batch uploads with multiple concerns defined, which this PR accomplishes. The point of the class was to do nothing. The idea that somebody has extended it seems... far fetched.

I've extended it to use my custom metadata. The batch upload form was broken for me in sufia 7.2 because of differences in metadata fields. still sorting out the implications of this pr... but seemed worth mentioning.

PR #2975, opened 2016-12-15
Add float to fieldset to prevent off-screen display in AdminSet edit view

mjgiarlo:
Before:

screenshot from 2016-12-15 14-08-44

After:

screenshot from 2016-12-15 14-08-25

### DIFF
--- a/app/assets/stylesheets/sufia/admin.scss
+++ b/app/assets/stylesheets/sufia/admin.scss
@@ -33,8 +33,12 @@ body {
   background-color: $admin-sidebar-background-color;
 }
 
+/* This is needed in Chrome for the admin/admin_sets edit view to render the participants tab
+ * Without this the contents of that tab flow off the right-hand side of the screen
+ * (The other containers on that page have `float: left` except for the fieldset.)
+ */
 fieldset {
-  width: 100%;
+  float: left;
 }
 
 #content-wrapper {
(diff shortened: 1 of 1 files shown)


jcoyne:
Can you add before/after screenshots about this?

mjgiarlo:
@jcoyne See the description above. Screenshots now included.

jcoyne:
Ah, was this the bug that was only on Chrome?

app/assets/stylesheets/sufia/admin.scss
```diff
@@ -33,6 +33,10 @@ body {
   background-color: $admin-sidebar-background-color;
 }
 
+fieldset {
```
jcoyne: I suspect that all the containers on that page have `float: left` except for that fieldset.  Would be up for experimenting to see if that fixes it for you? Otherwise add a comment about this being needed for the admin/admin_set_controller on Chrome.

mjgiarlo:
@jcoyne Is it only in Chrome? Huh, OK. I only use Chrome, so that is possible. ;)

app/assets/stylesheets/sufia/admin.scss
mjgiarlo: @jcoyne So I should remove the `width: 100%` and try `float: left` instead? K, I can give that a shot.

mjgiarlo: @jcoyne Yep, `float: left` works.

mjgiarlo:
@jcoyne Changes made.

PR #2978, opened 2016-12-16
Rake task to move all works to the default admin set

### LINKED ISSUE #2977: Create a Rake task to give all works an admin set
Sufia 7.3 will place all newly created or updated works in an admin set.  In order that our existing works may also be in an admin set, create a rake task that moves all works without an admin set to the default admin set.

Before starting, we ought to explore the business value of this.

jcoyne:
Fixes #2977


hackmastera:
+1

PR #2983, opened 2016-12-20
Use curation_concerns 1.7.1 or greater

### LINKED ISSUE #2979: Update to use CC1.7.0
Sufia should use CC 1.7.0 but there were breaking changes. So, make sufia master work atop CC master, then you can update the gemspec.

There is at least one known breaking change in CC:
https://github.com/projecthydra/curation_concerns/pull/1105#discussion_r91798384

This should fix a workflow-related bug in master that can be reproduced as follows:
- generate/run test app
- load workflows
- create an account
- deposit a work
- Upon redirection to the work show page, you're unauthorized to view, because depositor was never given edit access.

jcoyne:
Fixes #2979 


hackmastera:
:clap: I'm testing this on the workflow bug mentioned in that issue.

hackmastera:
update: the bug is gone.

PR #2999, opened 2017-01-07
Generate second concern in test app

### LINKED ISSUE #2998: Generate a second concern in CI
Far too much logic in code and tests relies on the fact there is only one possible CurationConcern at runtime.  This is more subtle than just looking for hardcoded `GenericWork` strings.  CI should generate a second concern before execution to reveal any such shortcomings.

atz:
fixes #2998

At this point we aren't even *doing* anything with the second concern, but it is still enough to blow up some of our ad hoc logic.

### DIFF
--- a/lib/generators/sufia/work/work_generator.rb
+++ b/lib/generators/sufia/work/work_generator.rb
@@ -25,10 +25,19 @@ module Sufia
       template 'locale.en.yml.erb', "config/locales/#{file_name}.en.yml"
     end
 
+    # Inserts after the last registering work, or at the top of the config block
     def register_work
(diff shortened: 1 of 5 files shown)


atz:
One simple change that shouldn't affect application logic at all, the mere presence of additional classes in a namspace...

  **17 Failures**: https://travis-ci.org/projecthydra/sufia/jobs/189698862#L2217

atz:
I'm going to break the groups of errors into semantic areas and ticket them separately:
- Zotero
- Arkivio
- Batch
- other

atz:
@jcoyne: i'm following one of the paths we discussed yesterday morning.  Updating the generator was not particularly difficult.

PR #3014, opened 2017-01-12
Redirect back to My Works after delete

### LINKED ISSUE #2447: When I delete a work from My Work Listing I get forwarded to the catalog
### Descriptive summary

In Sufia 7 when I delete a work from my work listing instead of staying on the work listing I get forwarded to the catalog.  This also occurs from my highlights, my collections and files shared with me.
### Expected behavior

I would expect that if I delete a work from a My Listing (Works, Highlights shares) that I would return to the My Listing after the deletion occurs.
### Actual behavior

What happens currently is that I get sent to the catalog index.

screen shot 2016-08-10 at 7 31 42 am

When I have works in the system 
And I go to My Work Listing
And I select delete from the Action Menu 
And I confirm the delete
Then I should return to the work listing
### Related work

Link to related tickets or prior related work here.

hortongn:
Fixes #2447

### DIFF
--- a/app/controllers/concerns/sufia/works_controller_behavior.rb
+++ b/app/controllers/concerns/sufia/works_controller_behavior.rb
@@ -72,6 +72,13 @@ module Sufia
         end
       end
 
+      def after_destroy_response(title)
+        respond_to do |wants|
+          wants.html { redirect_to sufia.dashboard_works_path, notice: "Deleted #{title}" }
+          wants.json { render_json_response(response_type: :deleted, message: "Deleted #{curation_concern.id}") }
+        end
+      end
+
       # Called by CurationConcerns::CurationConcernController#show
       def additional_response_formats(format)
         format.endnote do
--- a/spec/controllers/curation_concerns/generic_works_controller_spec.rb
+++ b/spec/controllers/curation_concerns/generic_works_controller_spec.rb
@@ -186,4 +186,23 @@ describe CurationConcerns::GenericWorksController do
       end
     end
   end
+
+  describe "#delete" do
+    let!(:work) { create(:work, title: ['test title'], user: user) }
+
+    context "after deletion" do
+      it "redirects to My Works" do
+        delete :destroy, params: { id: work }
+        expect(response).to redirect_to(Sufia::Engine.routes.url_helpers.dashboard_works_path)
+        expect(flash[:notice]).to eq "Deleted test title"
+      end
+
+      it "returns json" do
+        delete :destroy, params: { format: :json, id: work }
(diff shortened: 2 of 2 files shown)


app/controllers/concerns/sufia/works_controller_behavior.rb
```diff
@@ -72,6 +72,14 @@ def after_update_response
         end
       end
 
+      def after_destroy_response(title)
+        flash[:notice] = "Deleted #{title}"
+        respond_to do |wants|
+          wants.html { redirect_to sufia.dashboard_works_path }
```
jeremyf: As written, the `flash[:notice]` will be set even for JSON responses. You can move the `flash[:notice]` to `redirect_to` command via the following:

```ruby
wants.html { redirect_to sufia.dashboard_works_path, notice: "Deleted #{title}" }
```

spec/controllers/curation_concerns/generic_works_controller_spec.rb
```diff
@@ -186,4 +186,15 @@
       end
     end
   end
+
+  describe "#delete" do
+    let!(:work) { create(:work, title: ['test title'], user: user) }
+
+    context "after deletion" do
+      it "redirects to My Works" do
```
jeremyf: Please add a spec for JSON format.

spec/controllers/curation_concerns/generic_works_controller_spec.rb
```diff
@@ -186,4 +186,15 @@
...
+        delete :destroy, params: { id: work }
+        expect(response).to redirect_to(Sufia::Engine.routes.url_helpers.dashboard_works_path)
```
jeremyf: Do you want to verify that the flash has been set?

app/controllers/concerns/sufia/works_controller_behavior.rb
hortongn: 👍 fixed

spec/controllers/curation_concerns/generic_works_controller_spec.rb
hortongn: 👍 fixed

spec/controllers/curation_concerns/generic_works_controller_spec.rb
hortongn: 👍 fixed

jcoyne:
@cam156 this looks good to me. Does it meet your expectations as the author of the ticket?

hortongn:
Do we still need a 👍  from @cam156 on this before merging?

PR #3015, opened 2017-01-13
Autocomplete for child and parent relationships on work edit form.

### LINKED ISSUE #1563: UI Create _find_work_widget partial

andjsmit:
Fixes #1563

Creates autocomplete fields for child and parent relationship fields on work edit form using Question Your Authority and a new search builder.

### DIFF
--- a/app/forms/sufia/forms/work_form.rb
+++ b/app/forms/sufia/forms/work_form.rb
@@ -49,6 +49,12 @@ module Sufia::Forms
       model.ordered_members.to_a.select { |m| m.model_name.singular.to_sym != :file_set }
     end
 
+    # The in_work items
+    # @return [Array] All of the works that this work is a member of
+    def in_work_members
+      model.in_works.to_a
+    end
+
     def self.multiple?(term)
       return true if [:rights, :collection_ids].include? term
       super
--- a/app/search_builders/sufia/find_works_search_builder.rb
+++ b/app/search_builders/sufia/find_works_search_builder.rb
@@ -1,0 +1,44 @@
+# Search for possible works that user can edit and could be a work's child or parent.
+class Sufia::FindWorksSearchBuilder < Sufia::SearchBuilder
+  include Sufia::MySearchBuilderBehavior
+  include CurationConcerns::FilterByType
+
+  self.default_processor_chain += [:filter_on_title, :show_only_resources_deposited_by_current_user, :show_only_other_works, :show_only_works_not_child, :show_only_works_not_parent]
+
+  def initialize(context)
+    super(context)
+    @id = context.params[:id]
+    @q = context.params[:q]
+  end
+
+  def filter_on_title(solr_parameters)
+    solr_parameters[:fq] ||= []
+    solr_parameters[:fq] += [ActiveFedora::SolrQueryBuilder.construct_query(title_tesim: @q)]
+  end
+
+  def show_only_other_works(solr_parameters)
+    solr_parameters[:fq] ||= []
+    solr_parameters[:fq] += [
+      "-" + ActiveFedora::SolrQueryBuilder.construct_query_for_ids([@id])
+    ]
+  end
+
+  def show_only_works_not_child(solr_parameters)
+    ids = ActiveFedora::SolrService.query("{!field f=id}#{@id}", fl: "member_ids_ssim", rows: 10_000).flat_map { |x| x.fetch("member_ids_ssim", []) }
+    solr_parameters[:fq] ||= []
+    solr_parameters[:fq]  += [
+      "-" + ActiveFedora::SolrQueryBuilder.construct_query_for_ids(ids)
+    ]
+  end
+
+  def show_only_works_not_parent(solr_parameters)
+    solr_parameters[:fq] ||= []
+    solr_parameters[:fq]  += [
+      "-" + ActiveFedora::SolrQueryBuilder.construct_query(member_ids_ssim: @id)
+    ]
+  end
+
+  def only_works?
+    true
+  end
+end
--- a/app/assets/javascripts/sufia.js
+++ b/app/assets/javascripts/sufia.js
@@ -62,6 +62,7 @@
 //= require sufia/autocomplete/location
 //= require sufia/autocomplete/subject
 //= require sufia/autocomplete/language
+//= require sufia/autocomplete/work
 //= require sufia/relationships
 //= require curation_concerns/collections
 //= require hydra-editor/hydra-editor
--- a/app/assets/javascripts/sufia/autocomplete.es6
+++ b/app/assets/javascripts/sufia/autocomplete.es6
@@ -16,6 +16,11 @@ export class Autocomplete {
             case "based_near":
               this.autocompleteLocation(selector);
               break;
+            case "work":
+              var user = selector.data('user');
+              var id = selector.data('id');
+              this.autocompleteWork(selector, user, id);
+              break;
           }
       });
   }
@@ -47,5 +52,15 @@ export class Autocomplete {
   autocompleteLanguage(field) {
       var lang = require('sufia/autocomplete/language');
       new lang.Language(field, field.data('autocomplete-url'))
+  }
+
+  autocompleteWork(field, user, id) {
+    var work = require('sufia/autocomplete/work')
+    new work.Work(
+      field,
+      field.data('autocomplete-url'),
+      user,
+      id
+    )
   }
 }
--- a/app/assets/javascripts/sufia/autocomplete/work.es6
+++ b/app/assets/javascripts/sufia/autocomplete/work.es6
@@ -1,0 +1,29 @@
+export class Work {
+  // Autocomplete for finding possible related works (child and parent).
+  constructor(element, url, user, id) {
+    this.url = url;
+    this.user = user;
+    this.work_id = id;
+    element.autocomplete(this.options());
+  }
+
+  options() {
+    return {
+      minLength: 2,
+      source: ( request, response ) => {
+        $.getJSON(this.url, {
+          q: request.term,
+          id: this.work_id,
+          user: this.user
+        }, response );
+      },
+      focus: function() {
+        // prevent value inserted on focus
+        return false;
+      },
+      complete: function(event) {
+        $('.ui-autocomplete-loading').removeClass("ui-autocomplete-loading");
+      }
+    };
+  }
+}
--- a/app/authorities/qa/authorities/find_works.rb
+++ b/app/authorities/qa/authorities/find_works.rb
@@ -1,0 +1,15 @@
+module Qa::Authorities
+  class FindWorks < Qa::Authorities::Base
+    def search(_q, controller)
+      repo = CatalogController.new.repository
+      builder = Sufia::FindWorksSearchBuilder.new(controller)
+      response = repo.search(builder)
+      docs = response.documents
+      docs.map do |doc|
+        id = doc.id
+        title = doc.title
+        { id: id, label: title, value: id }
+      end
+    end
+  end
+end
--- a/app/views/curation_concerns/base/_find_work_widget.html.erb
+++ b/app/views/curation_concerns/base/_find_work_widget.html.erb
@@ -1,0 +1,16 @@
+<%= f.input id_type,
+      label: false,
+      type: :text,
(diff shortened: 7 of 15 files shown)


app/models/qa/authorities/local/find_works.rb
```diff
@@ -0,0 +1,23 @@
+module Qa::Authorities
+  class Local::FindWorks < ActiveFedora::Base
```
jcoyne: Why does this extend ActiveFedora::Base?  Shouldn't it extend `Qa::Authorities::Base`?

app/models/qa/authorities/local/find_works.rb
```diff
@@ -0,0 +1,23 @@
...
+      @q = q
+    end
+
+    def search(_q, controller)
+      repo = CatalogController.new.repository
+      builder = Sufia::FindWorksSearchBuilder.new(controller)
+      response = repo.search(builder)
+      docs = response.documents
+      results = []
+      docs.each do |doc|
```
jcoyne: Instead of doing `each` could you use `map` instead? Then you can also remove line 14 and 20

app/models/qa/authorities/local/find_works.rb
```diff
@@ -0,0 +1,23 @@
+module Qa::Authorities
+  class Local::FindWorks < ActiveFedora::Base
+    attr_reader :q
+
+    def initialize(q)
+      @q = q
```
jcoyne: Why are you assigning `@q`, or even defining `initialize`? This should be handled by the base.

app/search_builders/sufia/find_works_search_builder.rb
```diff
@@ -0,0 +1,44 @@
...
+  include CurationConcerns::FilterByType
+
+  def initialize(context)
+    super(context)
+    @id = context.params[:id]
+    @q = context.params[:q]
+  end
+
+  self.default_processor_chain += [:filter_on_title, :show_only_resources_deposited_by_current_user]
+  self.default_processor_chain += [:show_only_other_works, :show_only_works_not_child, :show_only_works_not_parent]
```
jcoyne: please combine lines 11 & 12 and move them above the initializer.

app/search_builders/sufia/find_works_search_builder.rb
```diff
@@ -0,0 +1,44 @@
+class Sufia::FindWorksSearchBuilder < Sufia::SearchBuilder
+  include Sufia::MySearchBuilderBehavior
```
jcoyne: Add a comment here about why you're including this.  Something like `# Only search for works that we have edit access to`

app/models/qa/authorities/local/find_works.rb
```diff
@@ -0,0 +1,23 @@
+module Qa::Authorities
```
jcoyne: I don't think this belongs in `app/models`, how about `app/authorities` instead? I also don't think we want it in the `Qa` namespace, how about `Sufia`?

spec/models/qa/authorities/local/find_works_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+
+  subject { service.search(q, controller) }
+
+  describe '#search' do
+    let!(:work1) { create(:generic_work, :public, title: ['foo'], user: user1) }
+    let!(:work2) { create(:generic_work, :public, title: ['foo foo'], user: user1) }
+    let!(:work3) { create(:generic_work, :public, title: ['foo foo foo'], user: user2) }
+    let!(:work4) { create(:generic_work, :public, title: ['bar'], user: user1) }
+
+    it 'shows a list of other works deposited by current user' do
```
jcoyne: Can you add a test that shows parent works/child works are properly excluded?

app/models/qa/authorities/local/find_works.rb
andjsmit: Would `app/authorities/sufia/local/find_works.rb` be correct ?

jcoyne: I think that would be fine.

app/models/qa/authorities/local/find_works.rb
andjsmit: Without this init, I get an error : `ArgumentError (wrong number of arguments (given 1, expected 0))` even after extending `Qa::Authorities::Base` as requested above.

app/models/qa/authorities/local/find_works.rb
jcoyne: Or  `app/authorities/sufia/authorities/local/find_works.rb` if you care going to put it in the `Sufia::Authoritities::Local` namespace.

app/search_builders/sufia/find_works_search_builder.rb
```diff
@@ -0,0 +1,44 @@
...
+    solr_parameters[:fq] += [
+      "-" + ActiveFedora::SolrQueryBuilder.construct_query_for_ids([@id])
+    ]
+  end
+
+  def show_only_works_not_child(solr_parameters)
+    ids = ActiveFedora::SolrService.query("{!field f=id}#{@id}", fl: "member_ids_ssim").flat_map { |x| x.fetch("member_ids_ssim", []) }
+    solr_parameters[:fq] ||= []
+    solr_parameters[:fq]  += [
+      "-" + ActiveFedora::SolrQueryBuilder.construct_query_for_ids([ids])
```
jeremyf: If I'm reading this correctly, the `ids` from line 27 should be an array. Then you are wrapping the `ids` in an array.

I believe you may want `ActiveFedora::SolrQueryBuilder.construct_query_for_ids(Array.wrap(ids))`

jcoyne: I think it's safe to say ids is an array (output of `#map`), there's no need to wrap it.

app/search_builders/sufia/find_works_search_builder.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  def show_only_other_works(solr_parameters)
+    solr_parameters[:fq] ||= []
+    solr_parameters[:fq] += [
+      "-" + ActiveFedora::SolrQueryBuilder.construct_query_for_ids([@id])
+    ]
+  end
+
+  def show_only_works_not_child(solr_parameters)
+    ids = ActiveFedora::SolrService.query("{!field f=id}#{@id}", fl: "member_ids_ssim").flat_map { |x| x.fetch("member_ids_ssim", []) }
```
jcoyne: make sure you pass `:rows` to `ActiveFedora::SolrService.query` or it'll only return the first 10.

andjsmit: I set `rows: 0` . Is this the correct ay to get all results?

andjsmit: @jcoyne Never mind, that breaks it. So, just set it to a large number? `rows: 10000` ?

jcoyne: @andjsmit yep. 10000 ought to be good.

app/authorities/sufia/authorities/local/find_works.rb
```diff
@@ -0,0 +1,21 @@
+module Sufia::Authorities
+  class Local::FindWorks < Qa::Authorities::Base
+    attr_reader :q
+
+    def initialize(q)
```
jeremyf: What is `q`. The terse variable name is confusing and not obvious.

andjsmit: I assumed it was the required query variable for the QA.  @jcoyne had mentioned removing it. But in doing so, the code crashes.

jeremyf: @andjsmit But what does the `q` variable mean?

andjsmit: Ok. I did a little digging. It is not the actual query as I thought. It appears to be the type? When I check the value of "q" in the init, I get `works`. So should I be calling it "type"?

app/authorities/sufia/authorities/local/find_works.rb
```diff
@@ -0,0 +1,21 @@
+module Sufia::Authorities
+  class Local::FindWorks < Qa::Authorities::Base
+    attr_reader :q
```
jcoyne: I don't see this attribute ever accessed. Is it needed?

andjsmit: @jcoyne Without it, a wrong number of arguments error occurs (given 1, expected 0)

```
 def instance
    klass.new(@subauthority)
```
in questioning_authority/lib/qa/authorities/local/registry.rb.

jcoyne: You are talking about the parameter to initialize.  I am talking about `attr_reader :q` that is never used. Additionally, Is there some reason you need to override initialize? Can  it be completely inherited?

andjsmit: @jcoyne Ok. The attr_reader can be removed. However, if I remove the `intitialize()`, the error above occurs.

andjsmit: @jcoyne @jeremyf Tracking further back, the init is pulling in the subauthority.

jcoyne: @andjsmit This is because you are attempting to https://github.com/projecthydra/sufia/pull/3015/files#diff-70cc36be61e57f96ec44887d4d62aa0fR120 register this as a subauthority. However it is not a subauthority. It is an authority in its own right.

jcoyne: @andjsmit Apparently you are doing it the recommended way.   This still seems like a bug in QA.  I see that this has been worked around in the  past like this: https://github.com/curationexperts/alexandria/blob/7e6dfafefb2cd9bc17aa956208e8a86495e5b1eb/app/authorities/local_names.rb#L16-L17

jcoyne: @andjsmit Okay, I looked more closely. Do not call `register_subauthority` and do not define an `initialize` method. This should just work because it is not a subauthority. It is equivalent to the built in tgnlang authority: https://github.com/projecthydra-labs/questioning_authority/blob/master/lib/qa/authorities/tgnlang.rb

app/forms/sufia/forms/work_form.rb
```diff
@@ -49,6 +49,12 @@ def ordered_work_members
       model.ordered_members.to_a.select { |m| m.model_name.singular.to_sym != :file_set }
     end
 
+    # The in_work items
+    # @return [Array] All of the works in the in_works
```
jcoyne: Can you rephrase "in the in_works" to "that this work is a member of"?

mjgiarlo:
Re-review, @jeremyf?

app/forms/sufia/forms/work_form.rb
andjsmit: Description changed.

mjgiarlo:
Well done, all!

PR #3017, opened 2017-01-17
[WIP] Hide files when I don't have permission

### LINKED ISSUE #2394: An open Access Work with restricted FileSets creates broken links
### Descriptive summary

In Sufia 7 master, when I create a work and set it to Institution, and then I later set the work access level to open access, I get broken thumbnail links, and visibility to FileSets I do not have access to.
### Expected behavior

If I am not logged in and a view an open access work with restricted files, I would not expect to see the files, and I would not expect broken links to the thumbnails if I can see the files.
### Actual behavior

I see the restricted files, but get an error when I click on them, and the thumbnail is a broken link
### Anonymous user

screen shot 2016-08-04 at 12 23 18 pm
### Logged in user

screen shot 2016-08-04 at 12 23 09 pm
1. Create a work that is restricted to institution
2. Then edit the work and change the permissions to public
3. Click only the work for the change (so that the files are left restricted)

hortongn:
Fixes #2394

### DIFF
--- a/app/views/curation_concerns/base/_member.html.erb
+++ b/app/views/curation_concerns/base/_member.html.erb
@@ -1,0 +1,16 @@
+<tr class="<%= dom_class(member) %> attributes">
+  <td class="thumbnail">
+    <% if member.reader? %>
+      <%= render_thumbnail_tag(member) %>
+    <% else %>
+      <%= link_to(image_tag(member.default_thumbnail), contextual_path(member, @presenter)) %> 
+    <% end %>
+  </td>
+  <td class="attribute filename"><%= link_to(member.link_name, contextual_path(member, @presenter)) %></td>
+  <td class="attribute date_uploaded"><%= member.try(:date_uploaded) %></td>
+  <td class="attribute permission"><%= member.permission_badge %></td>
+  <td>
+    <%= render 'actions', member: member %>
+  </td>
+</tr>
+
--- a/app/presenters/sufia/file_set_presenter.rb
+++ b/app/presenters/sufia/file_set_presenter.rb
@@ -9,6 +9,14 @@ module Sufia
       title.first
     end
 
+    def default_thumbnail
+      "default.png"
+    end
+
+    def reader?
+      current_ability.can?(:read, id)
+    end
+
     def link_name
       current_ability.can?(:read, id) ? page_title : 'File'
     end
--- a/spec/presenters/sufia/file_set_presenter_spec.rb
+++ b/spec/presenters/sufia/file_set_presenter_spec.rb
@@ -23,6 +23,32 @@ describe Sufia::FileSetPresenter do
   it { is_expected.to delegate_method(:date_modified).to(:solr_document) }
   it { is_expected.to delegate_method(:itemtype).to(:solr_document) }
 
+  describe '#default_thumbnail' do
+    it 'has a default thumbnail file name' do
+      expect(presenter.default_thumbnail).to eq 'default.png'
+    end
+  end
+
+  describe '#reader?' do
+    context "with a user who can view the file" do
+      before do
+          allow(ability).to receive(:can?).with(:read, "123abc").and_return(true)
+      end
+      it 'returns true' do
+        expect(presenter.reader?).to eq true
+      end
+    end
+
+    context "with a user who cannot view the file" do
+      before do
+          allow(ability).to receive(:can?).with(:read, "123abc").and_return(false)
+      end
+      it 'returns false' do
+        expect(presenter.reader?).to eq false
+      end
+    end
+  end
+
   describe '#link_name' do
     context "with a user who can view the file" do
       before do
--- a/spec/views/curation_concerns/base/_items.html.erb_spec.rb
+++ b/spec/views/curation_concerns/base/_items.html.erb_spec.rb
@@ -32,6 +32,7 @@ describe 'curation_concerns/base/items', type: :view do
     before do
       stub_template 'curation_concerns/base/_actions.html.erb' => 'Actions'
       allow(presenter).to receive(:member_presenters).and_return([file_set, member])
+      allow(member).to receive(:reader?).and_return(true)
       allow(view).to receive(:blacklight_config).and_return(Blacklight::Configuration.new)
       allow(view).to receive(:blacklight_configuration_context).and_return(blacklight_configuration_context)
       allow(view).to receive(:contextual_path).and_return("/whocares")
--- a/spec/views/curation_concerns/base/_member.html.erb_spec.rb
+++ b/spec/views/curation_concerns/base/_member.html.erb_spec.rb
@@ -1,0 +1,74 @@
+require 'spec_helper'
+
+describe 'curation_concerns/base/_member.html.erb' do
+  let(:solr_document) { SolrDocument.new(id: '999',
+                                         has_model_ssim: ['FileSet'],
+                                         active_fedora_model_ssi: 'FileSet',
+                                         thumbnail_path_ss: '/downloads/999?file=thumbnail',
+                                         representative_tesim: ["999"],
+                                         title_tesim: ["My File"]) }
+
+  # Ability is checked in FileSetPresenter#link_name
+  let(:ability) { double(can?: true) }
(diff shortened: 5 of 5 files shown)


jcoyne:
I'm not sure this is what is wanted across the board.  I believe there was some intentionality in showing files, but indicating you don't have access to them. Does anyone want to speak up in support of showing files you don't have access to? @cam156 @awead @jeremyf @mjgiarlo

app/presenters/sufia/work_show_presenter.rb
```diff
@@ -42,6 +42,10 @@ def stats_path
       Sufia::Engine.routes.url_helpers.stats_work_path(self)
     end
 
+    def member_presenters(ids = ordered_ids, presenter_class = composite_presenter_class)
+      super.delete_if { |presenter| current_ability.cannot?(:read, presenter.solr_document) }
```
jcoyne: This is a very slow implementation as it's going to do N+1 queries.  It should be possible to query it a single time with access controls.

hortongn: @jcoyne Sorry, neglected to say that this is the start of a fix for #2394.   I'm using the fix already in ScholarSphere as a start https://github.com/psu-stewardship/scholarsphere/commit/206f1738ff6051b4634ee04f5a18fbe9310fd4b5

jcoyne:
It looks like @cam156 is in support of this change (https://github.com/projecthydra/sufia/issues/2394#issuecomment-243806741), what about other users @hackmastera, Princton?

atz:
Performance would seem to be a massive degradation when the set scales up to thousands of files.  We already know use cases will take this into tens of thousands.  Leveraging Solr is the only way to make this reasonable.  

It is one thing to manually interrogate with a set expected to be very small or in the absence of actual object instances (e.g., for modeled object classes, as in `user.can? :create, type` in `authorized_models`).  But the approach here, on a major access of growth, seems precarious and feature-poor.  

Looking at `CurationConcerns::PresenterFactory` that is the underlying code, we see:
- hardcoded limit of `rows:1000` in the query,
- no sort order,
- no pagination,
- no permissions awareness, literally just operates on a list of IDs, because
- no filtering available of any kind.
- calls `SolrDocument.new` for each document returned, then
- calls `klass.new(solr_doc, *args)` for each SolrDocument.

So if you are looking at a set of 1005 files, only the last 5 of which you can view, this code would populate 1000 objects (twice) before deleting each of them and then **still** not show you your 5.  

Tl;dr: this should use the SearchBuilder pattern.  I don't see any way around it.

cam156:
@jcoyne I would definately prefer  not to see the files I do not have access.  I would be amenable to an indicator that there are files in the work that I do not have permissions to, but showing broken thumbnails and partial metadata for private items was the item I was ticketing.

app/presenters/sufia/work_show_presenter.rb
cam156: @hortongn While that fix works for PSU where we have few files in the works at present, it may not work for every implementation.  I agree with @atz that a SearchBuilder or a SolrDocument.where query would be best for everyone and likely PSU would need to replace the code we have in the future.  I think the "better" fix would be to include hydra-access-controls into this query: https://github.com/projecthydra/curation_concerns/blob/master/app/presenters/curation_concerns/work_show_presenter.rb#L132

hortongn: Appreciate the input and tips!  Thanks @jcoyne @atz @cam156

hackmastera:
@cam156's proposal works for us, thanks!

tpendragon:
This would seriously break the file manager feature in Curation Concerns, because everything's stored in relation to its siblings. So when it pulls member presenters out to display for reorder, and you reorder, the moment you hit "save" it would effectively remove the object you didn't have permission to from the parent object - which seems much worse than showing them the file.

mark-dce:
DCE definitely has worked with institutions that want to indicate the availability of an asset (an attached file), but not allow it to be downloaded - i.e. this file, image, recording exists but you need to do something special to access it - usually contact a librarian or sometimes just log in.

Long-term I'd be a fan of a configuration option that allows the implementing institution to either "display metadata about restricted files" or "hide metadata about restricted files" associated with a work.

tpendragon:
I should note I'm not against hiding files, just not like this - what you'd have to do is pull the records in server-side and decide in the view on how to render them (is it a box, or a hidden field)

cam156:
@tpendragon Do you do the File Manager stuff in the show view or the edit view?  This PR is for the show view.

tpendragon:
@cam156 It's in its own view, but the work show presenter is used for it. It doesn't have to be, but it's a similar issue of what to show.

cam156:
@tpendragon & @jcoyne  So what I am hearing is in the Sufia/CurationConcern world the best bet would be for Sufia to override file ids to give back the ids the user can see.  That leaves the file manager in CC alone, but allows for an IR perspective which would call for users to only see what they have permissions for.  In the hyrax world then we need to separate the presenters for the show (with access controls) and the file manager (without access controls).

So for @hortongn the change wanted in this PR is to include the file_ids method that is in CC and change it to only gather ids with access.  https://github.com/projecthydra/curation_concerns/blob/master/app/presenters/curation_concerns/work_show_presenter.rb#L134

Is that a correct statement?

hackmastera:
@cam156 File manager is integrated into Sufia as of 7.2.

What happens currently if a user has edit access to a work but not all of the files / works contained therein? It seems like those people should not have access to file manager. @tpendragon

cam156:
@hackmastera Ah!  I did not know that.  Thanks!

hortongn:
So if Sufia 7.2 has a file manager integrated now, it sounds like the solution for this issue needs to allow for the file manager to work properly while also hiding files in the show view when the user doesn't have permission to access them?  The file manager and the show view need to use separate presenters?

mjgiarlo:
I believe the solution we arrived at in various discussions in Slack and GitHub is that the presenter should use a default icon (`app/assets/images/default.png`) instead of a thumbnail to represent a file when the user lacks access. Thoughts on that, @jcoyne @cam156 @tpendragon @mark-dce @hackmastera @hortongn 

Assuming that is the consensus approach, are you still inclined to take this on, @hortongn?

hortongn:
@mjgiarlo Yep, if all agree on that resolution I can knock it out this week.  I'm tied up in the UCSD POOD class during the day this week, but this should be simple enough to resolve in the evening.  

I'll move ahead after 👍 from @jcoyne @cam156 @tpendragon @mark-dce @hackmastera on this approach.

hortongn:
Now that we've decided the fix for this should be to show default thumbnails for files when the user doesn't have read access, I've pushed up a quick fix to this PR that works.  Looking for input on whether this view/presenter change is the best way to solve it. @mjgiarlo @jcoyne

mjgiarlo:
@tpendragon @jcoyne re-review when you have a chance? @hortongn has pushed up changes, and this is the one remaining blocker for Sufia 7.3.0.

app/views/curation_concerns/base/_member.html.erb
```diff
@@ -0,0 +1,16 @@
+<tr class="<%= dom_class(member) %> attributes">
+  <td class="thumbnail">
+    <% if member.reader? %>
```
mjgiarlo: This line is causing test failures because it's calling `#reader?` on the `WorkShowPresenter`, and that method is only defined on the `FileSetPresenter`. Perhaps it needs to be defined on both classes.

hortongn: Yeah I saw that too.  I was holding on the tests until I know if this is an acceptable approach or not.

mjgiarlo: Ah, that's a good call. @jcoyne @tpendragon any chance y'all can find time for this today? Can this solution work for 7.3.0? (We can ticket further optimization in Hyrax if needed.)

jcoyne: Shouldn't the download controller just show the appropriate (default if no read access) thumbnail?

mjgiarlo: @jcoyne is the downloads controller used in this partial when `render_thumbnail_tag` is called?

PR #3019, opened 2017-01-19
Rename batch upload item

atz:
Closer alignment w/ rails convention

See https://github.com/projecthydra/sufia/pull/2970#issuecomment-273875386 
and https://github.com/projecthydra/sufia/issues/2913#issuecomment-266927897

This is based off #2970.  Will rebase after that gets merged for the single distinct rename commit.

### DIFF
--- a/app/controllers/concerns/sufia/batch_uploads_controller_behavior.rb
+++ b/app/controllers/concerns/sufia/batch_uploads_controller_behavior.rb
@@ -7,7 +7,7 @@ module Sufia
     included do
       self.work_form_service = BatchUploadFormService
       self.curation_concern_type = work_form_service.form_class.model_class # includes CanCan side-effects
-      # We use BatchUploadItem as a null stand-in curation_concern_type.
+      # We use BatchUpload as a null stand-in curation_concern_type.
       # The actual permission is checked dynamically during #create.
     end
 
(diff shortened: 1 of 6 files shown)


app/controllers/concerns/sufia/batch_uploads_controller_behavior.rb
```diff
@@ -6,25 +6,39 @@ module BatchUploadsControllerBehavior
...
+      # We use BatchUpload as a null stand-in curation_concern_type.
+      # The actual permission is checked dynamically during #create.
     end
 
+    # The permissions to create a batch are not as important as the permissions for the concern being batched.
+    # @note we don't call `authorize!` directly, since `authorized_models` already checks `user.can? :create, ...`
     def create
       authenticate_user!
-      create_update_job
+      unsafe_pc = params.fetch(:batch_upload, {})[:payload_concern]
```
jeremyf: Consider extracting a `safe_pc` method (though I'm uncertain what that method should be called). The mental dissonance I'm getting is that we have some concise methods that are rather quick to read. The change introduces a different level of abstraction to the method.

atz: This is part of the underlying `batch_create_work_type` branch and will fall out after that gets merged.  You might have a point, but its applicability would be on #2970.

atz:
Rebased

jcoyne:
👎 to this change. The `BatchUploadItem` represents one work of a batch upload.  It is confusing to change it to be called a "Batch Upload" in and of itself. The Batch Upload is the collection of `BatchUploadItems`.

mjgiarlo:
All, we'll be discussing this on stand-up tomorrow. Hang in there for now 'til we have (virtual) facetime to chat.

atz:
We discussed this in call today and agreed the class is used as the `model_class` of the `BatchUploadsController` and `BatchUploadForm`.  It is never instantiated as a per-file or per-item from the batch.  However, fo release mgmt purposes we will incorporate the change in Hyrax instead of Sufia.

atz:
Closing, replaced with: https://github.com/projecthydra-labs/hyrax/pull/299

PR #3024, opened 2017-01-20
Allow multiple resource types per file in batch upload form

### LINKED ISSUE #3023: Resource type in Batch Create should be multivalue
as it is in the metadata form.

hackmastera:
Fixes #3023 

### DIFF
--- a/app/jobs/batch_create_job.rb
+++ b/app/jobs/batch_create_job.rb
@@ -28,7 +28,7 @@ class BatchCreateJob < ActiveJob::Base
       raise ArgumentError, 'attributes must include "model" => ClassName.to_s' unless model
       uploaded_files.each do |upload_id|
         title = [titles[upload_id]] if titles[upload_id]
-        resource_type = [resource_types[upload_id]] if resource_types[upload_id]
+        resource_type = Array(resource_types[upload_id]) if resource_types[upload_id]
         attributes = attributes.merge(uploaded_files: [upload_id],
                                       title: title,
                                       resource_type: resource_type)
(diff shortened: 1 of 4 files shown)


app/jobs/batch_create_job.rb
```diff
@@ -28,7 +28,7 @@ def create(user, titles, resource_types, uploaded_files, attributes, log)
       raise ArgumentError, 'attributes must include "model" => ClassName.to_s' unless model
       uploaded_files.each do |upload_id|
         title = [titles[upload_id]] if titles[upload_id]
-        resource_type = [resource_types[upload_id]] if resource_types[upload_id]
+        resource_type = Array(resource_types[upload_id]) if resource_types[upload_id]
```
jeremyf: Favor `Array.wrap` instead of `Array`.

```ruby
 Array(Time.now)
=> [19, 15, 9, 23, 1, 2017, 1, 23, false, "EST"]

Array(Time.now)
=> [Time.now]
```

spec/jobs/batch_create_job_spec.rb
```diff
@@ -65,5 +65,14 @@
         expect(log.reload.status).to eq 'failure'
       end
     end
+
+    context "when multiple resource types are passed" do
+      let(:resource_types) { { upload1.id.to_s => ['Image', 'Text'], upload2.id.to_s => ['Image', 'Text'] } }
+      it "sets them all on the record" do
+        subject
+        expect(GenericWork.last.reload.resource_type).to include 'Image'
```
jeremyf: Each call to `GenericWork.last.reload` is hitting Fedora. Please capture in a local variable. I'm wondering if you could assert the expected array?

```ruby
resource_types = GenericWork.last.reload.resource_type
expect(resource_types.sort).to eq(['Image', 'Text'])
```

hackmastera: Good catch. I don't think I can do it that way since the order won't be guaranteed. But I can grab them once instead of twice for sure.

PR #3039, opened 2017-01-25
Make all resource type buttons work in batch create. Fixes #3031

### LINKED ISSUE #3031: Batch upload 'set all to this resource type' button - only the first works
There was previously some [discussion](https://github.com/projecthydra/sufia/issues/1869) over whether only the first should show.

It turns out that issue was missing the crucial information that only the first one works. This makes sense due to the wireframe used for development, but either the others need to be hidden or the code should be changed such that they all work.

mjgiarlo:
This is probably not the idiomatic JavaScript way to go. Pointers welcome, especially those are short of large refactorings.

### DIFF
--- a/app/views/sufia/uploads/_js_templates.html.erb
+++ b/app/views/sufia/uploads/_js_templates.html.erb
@@ -33,8 +33,8 @@
 
 <!-- function used by the following template -->
 <script type="text/javascript">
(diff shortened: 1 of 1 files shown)


hackmastera:
I'm not idiomatic javascript but I think it looks good.

PR #3054, opened 2017-02-03
Add optimistic locking to works

### LINKED ISSUE #3046: Files not added to a work when work is edited during background processing
This bug affects 7.2 and 7.3.

### Descriptive summary

If you upload a number of files to a work, then edit it and save while those files are still being processed in the background, they do not get successfully added to the work.

### Expected behavior

Background jobs and foreground jobs shouldn't mess with one another. Or, if they do, the UI shouldn't allow you to do things that will mess up your work.

### Actual behavior

Using the edit form appears to delete from the work any file added between the time the form loads and the time it saves.

1. Generate .internal_test_app
1. add `gem 'sidekiq'` to .internal_test_app/Gemfile; `bundle install`
1. add sidekiq routes to .internal_test_app/config/routes.rb ```require 'sidekiq/web'
  mount Sidekiq::Web => '/sidekiq'```
1. add `    config.active_job.queue_adapter = :sidekiq` to .internal_test_app/config/application.rb
1. run solr, fcrepo, rails server
1. in another terminal run sidekiq with `bundle exec sidekiq -q default -q ingest -q event`
1. in the browser, create a new work. upload a bunch of files to give yourself time; probably 15 is enough.
1. After you save, reload the work to see the first couple of files on the items list. click 'edit' before too many more files load.
1. make some metadata change (not sure if this is required, probably isn't) and click save (that part is required). Don't go too fast.
1. you can watch the sidekiq dashboard to see the jobs finish up.
1. reload the work to see that not all your files were added.

jcoyne:
Fixes #3046

### DIFF
--- a/app/actors/sufia/optimistic_lock_validator.rb
+++ b/app/actors/sufia/optimistic_lock_validator.rb
@@ -1,0 +1,28 @@
+module Sufia
+  # Validates that the submitted version is the most recent version in the datastore.
+  # Caveat: we are not detecting if the version is changed by a different process between
+  # the time this validator is run and when the object is saved
+  class OptimisticLockValidator < CurationConcerns::Actors::AbstractActor
+    class_attribute :version_field
+    self.version_field = 'version'
+
+    def update(attributes)
+      validate_lock(version_attribute(attributes)) && next_actor.update(attributes)
+    end
+
+    private
+
+      # @return [Boolean] returns true if the lock is missing or
+      #                   if it matches the current object version.
+      def validate_lock(version)
+        return true if version.blank? || version == curation_concern.etag
+        curation_concern.errors.add(:base, :conflict)
+        false
+      end
+
+      # Removes the version attribute
+      def version_attribute(attributes)
+        attributes.delete(version_field)
+      end
+  end
+end
--- a/app/forms/sufia/forms/work_form.rb
+++ b/app/forms/sufia/forms/work_form.rb
@@ -7,6 +7,10 @@ module Sufia::Forms
     # is merged and hydra-editor 3.0.0 is released
     delegate :model_name, to: :model
 
+    def version
+      model.etag
+    end
+
     attr_reader :agreement_accepted
 
     self.terms += [:collection_ids, :admin_set_id]
@@ -61,7 +65,7 @@ module Sufia::Forms
     end
 
     def self.build_permitted_params
-      super + [:on_behalf_of, { collection_ids: [] }]
+      super + [:on_behalf_of, :version, { collection_ids: [] }]
     end
   end
 end
--- a/app/models/concerns/sufia/work_behavior.rb
+++ b/app/models/concerns/sufia/work_behavior.rb
@@ -10,5 +10,11 @@ module Sufia
     included do
       self.indexer = Sufia::WorkIndexer
     end
+
+    # TODO: Move this into ActiveFedora
+    def etag
+      raise "Unable to produce an etag for a unsaved object" unless persisted?
+      ldp_source.head.etag
+    end
   end
 end
--- a/app/services/sufia/actor_factory.rb
+++ b/app/services/sufia/actor_factory.rb
@@ -1,7 +1,8 @@
 module Sufia
   class ActorFactory < CurationConcerns::Actors::ActorFactory
     def self.stack_actors(curation_concern)
-      [CreateWithRemoteFilesActor,
+      [OptimisticLockValidator,
+       CreateWithRemoteFilesActor,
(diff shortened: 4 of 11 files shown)


jcoyne:
@hackmaster I can't explain the missing `en.sufia.works.update.header` as that's not something I messed with. Ideas?  The nil case should be handled because I'm checking for `blank?`

hackmastera:
I'm wondering if we want the nil case to return false, though. or is there another way to ensure that forms are submitting this param?

hackmastera:
I don't know what is up with the translation error. I don't get it before this commit. I only get it when I am forcing the save conflict; saving without conflict goes fine. I'll try putting a value in for it and see where it shows up.

jcoyne:
The reason the null case is submitting false is primarily for batch jobs. I'd assume that every form would want it

hackmastera:
When I put in a dummy value for that in the translation file I get an unknown attribute error on submission: https://gist.github.com/hackmastera/4f08beb6dde780e9b32bd7c6cac30ea1

hackmastera:
What happens with 'version' / lock validation in each of the following scenarios?

- 'create new work' form: 'version' is nil, validation succeeds
- 'edit work' form: 'version' is populated via hidden field, validation succeeds or fails based on that value
- 'file manager': 'version' is nil, validation succeeds
- 'batch create' form: 'version' is nil for each object, validation succeeds
- 'batch edit' form: ??? 'version' is nil for each object, validation succeeds
- downstream override of 'edit work' form: 'version' is nil due to developer ignorance, validation succeeds

Are these scenarios correct? If yes, are these the desired outcomes? If no, what should we do in this pr and what should we create new issues for?

jcoyne:
>    'create new work' form: 'version' is nil, validation succeeds

The validator is not run on create, only update.  The version field is not on the create form.

>   'edit work' form: 'version' is populated via hidden field, validation succeeds or fails based on that value

Yes. This is the correct behavior

> 'file manager': 'version' is nil, validation succeeds

I haven't considered the file manager.  It should probably have the version on its form.

> 'batch create' form: 'version' is nil for each object, validation succeeds

No validation occurs on create, only update

>   'batch edit' form: ??? 'version' is nil for each object, validation succeeds

I didn't consider this case.  I don't think we care about it, as we're just updating a single field in this case (not sending `ordered_member_ids`)

>  downstream override of 'edit work' form: 'version' is nil due to developer ignorance, validation succeeds

Here's some rope. I'd advise not wearing it around your neck.

hackmastera:
lol.

> > 'file manager': 'version' is nil, validation succeeds

> I haven't considered the file manager. It should probably have the version on its form.

+1

> > 'batch edit' form: ??? 'version' is nil for each object, validation succeeds

> I didn't consider this case. I don't think we care about it, as we're just updating a single field in this case (not sending ordered_member_ids)

This is a lower-priority scenario for me/us, but it would be worth creating a hyrax issue for, since two people editing the same object at the same time could cause overwrites without anyone realizing.

> > downstream override of 'edit work' form: 'version' is nil due to developer ignorance, validation succeeds

> Here's some rope. I'd advise not wearing it around your neck.

Fair enough. We could send a fail when 'version' is nil to help address this case. But it doesn't seem right, really, to addess it in the UI in this way so *shrug*.

jcoyne:
@hackmastera the "unknown attribute error on submission" is a rails autoloading bug thing.  Same one we were getting before for "remote files".  The fix is to restart the devlopment server, run hyrax, or run in production (no auto reloading).

jcoyne:
Ticket for optimistically locking the file manager: https://github.com/projecthydra/sufia/issues/3055

hackmastera:
@jcoyne two problems when I tried this this morning.

1. Because the form is being reloaded through the update action, the page errors unless the following entries are present in the translation file. I'm not sure if it's possible to look for them in en.sufia.works.edit but it does work if you copy them over.
  1. en.sufia.works.update.header
  1. en.sufia.works.update.in_collections
  1. en.sufia.works.update.in_this_work
  1. en.sufia.works.update.in_other_works
1. The new "Another user has made a change..." message does not show up. This could be because I did not dismiss the "Your files are being processed by Sufia in the background," etc. message.

jcoyne:
@hackmastera hmm, those translation issues sound like a preexisting condition that could be caused by any validation failure.

hackmastera:
@jcoyne what do you suggest the best fix is for the translation issues? a separate PR?

jcoyne:
@hackmastera yes, I think a separate issue.

jcoyne:
Superseeded by #3066

PR #3057, opened 2017-02-06
Pinning rake to get rid of last_comment error

cam156:
Pins rake to remove the error `NoMethodError: undefined method 'last_comment' for #<Rake::Application:0x00000002997a30>`

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -12,7 +12,7 @@ group :development, :test do
   gem "simplecov", require: false
   gem 'byebug' unless ENV['CI']
   gem 'coveralls', require: false
-  gem 'rubocop', '~> 0.37.2', require: false
+  gem 'rubocop', '~> 0.38.0', require: false
   gem 'rubocop-rspec', require: false
 end
 
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,11 +1,10 @@
-require: rubocop-rspec
-
 AllCops:
   TargetRubyVersion: 2.2
   DisplayCopNames: true
   Include:
     - '**/Rakefile'
     - '**/config.ru'
+    - rubocop-rspec
   Exclude:
     - 'db/**/*'
     - 'script/**/*'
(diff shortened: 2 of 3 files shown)


Gemfile
```diff
@@ -14,6 +14,7 @@ group :development, :test do
   gem 'coveralls', require: false
   gem 'rubocop', '~> 0.37.2', require: false
   gem 'rubocop-rspec', require: false
+  gem 'rake', '< 11.0'
```
jcoyne: I'm using rake 12 with no trouble. Can you show how you get such an error?

jcoyne: Are you perhaps using an old version of rspec?

cam156: @jcoyne I think it has to do with the ruby version.  You can see the error in travis: https://travis-ci.org/projecthydra/sufia/jobs/198929354

jcoyne: That's fixed in rubocop 0.38 https://github.com/bbatsov/rubocop/commit/4fc1d719f5444a3fb7258255238dd6df4fd35abe

cam156: @jcoyne so why is travis failing then?

jcoyne: @cam156 you've got rubocop pinned to 0.37

cam156:
@jcoyne changed the version of rubocop, which also required a change to the yml format.  Thanks for pointing that out.

cam156:
@jcoyne my original change of the rake version passed CI.  This new change of rubocop version causes an incompatible jwt version.  At this point my plan it to abandon getting sufia 6 to run on ci since you will not approve the change that makes it work for the different versions.

jcoyne:
The rubocop does not cause the jwt version conflict.  That is due to the google api client and oauth2.

jcoyne:
@cam156 why do we even have rake in the gemspec?  Most engine gems don't.

cam156:
@jcoyne not entirely sure.  At this point I'm not going to worry about it.  I was trying to put a one line change into sufia and now I am going to just put it in ScholarSphere.  Not many people are interested in sufia 6 at this point.

PR #3064, opened 2017-02-07
Collections use only members, not ordered_members, fix #3063

### LINKED ISSUE #3063: Collection members should not be migrated to ordered_members
The sufia and curation concerns code do not maintain ordered_members, so you could end up with a situation like #3062.

This line https://github.com/projecthydra/sufia/blob/7.2-migration/lib/sufia/import/collection_builder.rb#L30 needs to be changed to use `#members`.

hackmastera:
Fixes #3063 

@cam156 


jcoyne:
It appears rubocop has updated and found lots of new things to complain about.

PR #3066, opened 2017-02-07
Add optimistic locking to works

### LINKED ISSUE #3046: Files not added to a work when work is edited during background processing

jcoyne:
Using the new optimistic lock in CurationConcerns
Fixes #3046

### DIFF
--- a/app/presenters/sufia/work_show_presenter.rb
+++ b/app/presenters/sufia/work_show_presenter.rb
@@ -4,8 +4,6 @@ module Sufia
     delegate :based_near, :related_url, :depositor, :identifier, :resource_type,
              :keyword, :itemtype, :admin_set, to: :solr_document
 
-    self.file_presenter_class = Sufia::FileSetPresenter
-
     def editor?
       current_ability.can?(:edit, solr_document)
     end
--- a/app/services/sufia/actor_factory.rb
+++ b/app/services/sufia/actor_factory.rb
@@ -1,7 +1,8 @@
 module Sufia
   class ActorFactory < CurationConcerns::Actors::ActorFactory
     def self.stack_actors(curation_concern)
-      [CreateWithRemoteFilesActor,
+      [CurationConcerns::OptimisticLockValidator,
+       CreateWithRemoteFilesActor,
        CreateWithFilesActor,
        CurationConcerns::Actors::AddToCollectionActor,
        CurationConcerns::Actors::AddToWorkActor,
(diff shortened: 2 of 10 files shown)


mjgiarlo:
@hackmastera You're seeing "Your files are being processed" on the edit form? I can't say I've ever seen a flash message rendered in that view. How to reproduce?

jcoyne:
@hackmastera I've added the message to the form.  I'm not sure how you could be seeing "your files are being processed" if it's not saving the form.  That is only shown when the save is successful.

screen shot 2017-02-08 at 4 08 26 pm

hackmastera:
Looks like @jcoyne's latest push fixed it for me in the simplest case. Going to run through the steps in #3046 now with sidekiq and everything.

hackmastera:
I'm seeing the error in the more complicated case, as well. But I also see the processing notice. @mjgiarlo @jcoyne what do y'all think of this 
screen shot 2017-02-09 at 10 21 37 am

hackmastera:
My last concern is that the user experience is rather befuddling. The error message is factual but gives no advice or context. The form reloads completely and any edits the user was trying to make are lost. Plus there's no way for the user to tell whether / when it's safe to try again.

Much of this is unavoidable as a practical matter for us, but let me take a stab at a new error message:

Your changes could not be saved because another user (or background job) updated this #{model} after you began editing. Please make sure all file attachments have completed successfully and try again. This form has refreshed with the most recent saved copy of the #{model}.

jcoyne:
New error messages added in https://github.com/projecthydra/curation_concerns/pull/1149

jcoyne:
New error messages added in https://github.com/projecthydra/curation_concerns/pull/1149

hackmastera:
Okay I tried some stuff just on master and that notification appears to just stick around / follow me for some reason. unrelated to this PR.

PR #3067, opened 2017-02-08
Turn off message spies in rubocop

hackmastera:

### DIFF
--- a/sufia.gemspec
+++ b/sufia.gemspec
@@ -64,7 +64,7 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency "equivalent-xml", '~> 0.5'
   spec.add_development_dependency "jasmine", '~> 2.3'
   spec.add_development_dependency 'rubocop', '~> 0.42.0'
-  spec.add_development_dependency 'rubocop-rspec', '~> 1.5'
+  spec.add_development_dependency 'rubocop-rspec', '~> 1.9.1'
   spec.add_development_dependency 'shoulda-matchers', '~> 3.1'
   spec.add_development_dependency 'rails-controller-testing', '~> 0'
 end
--- a/sufia.gemspec
+++ b/sufia.gemspec
@@ -49,6 +49,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'flipflop', '~> 2.2'
   spec.add_dependency 'jquery-datatables-rails', '~> 3.4.0'
   spec.add_dependency 'rdf-rdfxml'
+  spec.add_dependency 'sprockets-rails', '~> 3.1.1'
 
   spec.add_development_dependency 'engine_cart', '~> 1.0'
   spec.add_development_dependency 'mida', '~> 0.3'
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -101,3 +101,6 @@ RSpec/NestedGroups:
 
 RSpec/LeadingSubject:
   Enabled: false
+
+RSpec/MessageSpies:
+    Enabled: false


hackmastera:
see this done upstream, with more reasons, here: 4a9d3d26dcec09db8c8154cebfe5511876f05b45

hackmastera:
Also pinning rubocop-rspec. I think this is fine on a legacy branch.

hackmastera:
Although i know we plan to release 7.2.1 soon. I don't think the tests rubocop-rspec was complaining about have been reworked upstream. I'll look now; if they have been I will cherry-pick that commit instead.

hackmastera:
they haven't been; the last master build was run using rubocop-rspec 1.8.

cam156:
@hackmastera looks like maybe bootstrap also changed.  Arrg!

hackmastera:
okay i'm trying this one last thing and then i will probably quit.

jcoyne:
I'm curious as to what this was evaluating to without the pinning. What was the bootstrap halfling issue?

hackmastera:
@jcoyne from what I read, the issue was that the assets were being generated with hashes appended, but invoked without those hashes.

hackmastera:
@projecthydra/sufia-code-reviewers this is green now and i would self-merge except it was approved before the last couple of commits so needs another look.

sufia.gemspec
```diff
@@ -49,6 +49,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'flipflop', '~> 2.2'
   spec.add_dependency 'jquery-datatables-rails', '~> 3.4.0'
   spec.add_dependency 'rdf-rdfxml'
+  spec.add_dependency 'sprockets-rails', '~> 3.1.1'
```
jcoyne: Do we want to preclude people from using sprockets-rails 3.2 should it be released?

hackmastera: 3.2 is out and travis was going bananas on it.

@jcoyne I think your hesitation is well-founded; it doesn't seem like anything has changed recently but this is also a slow-moving branch and it's possible no one has pushed to it since 3.2 wasa released. But I am not able to pursue a better fix for this right now. We can either leave the build broken or maybe ticket this issue and accept this PR so that this branch can move forward. I could split the PR into pieces according to whatever agrees with people, and override the travis check to leave the branch at least less broken. Open to whatever @projecthydra/sufia-code-reviewers prefer.

jcoyne: It appears that 3.2 was released in Sept 2016. So it's likely no-one encountered this on this branch. https://rubygems.org/gems/sprockets-rails/versions/3.2.0

hackmastera: Hm, 7.2-migration has had commits as recently as December so 3.2 was definitely out then.

jcoyne: Here's a successful 7.2-migration build https://travis-ci.org/projecthydra/sufia/jobs/185841418 on Dec 21.

jcoyne: That build was `Using sprockets-rails 3.2.0`

jcoyne: Here's the diff of the bundle install https://gist.github.com/jcoyne/4ca52ed0e6bc5f95e666717ce35db4a5

hackmastera: @jcoyne thanks for this diff. I could try some of these if you maybe have some ideas of possible starting points.

jcoyne: I don't see anything there that jumps out at me.

hackmastera: Is it possible the asset pipeline is pulling down something remotely that has changed?

PR #3071, opened 2017-02-09
Update administrivia

mjgiarlo:


README.md
```diff
@@ -119,8 +119,8 @@ Generate a new Rails application. We recommend the latest Rails 5.0 or 4.2 relea
 
 ```
 # If you don't already have Rails at your disposal...
-gem install rails -v 5.0.0.1
-rails new my_app -m https://raw.githubusercontent.com/projecthydra/sufia/master/template.rb
+gem install rails -v 5.0.1
+rails new my_app -m https://raw.githubusercontent.com/projecthydra/sufia/v7.3.0.rc1/template.rb
```
mjgiarlo: This change will require that we tweak the 7.3.0.rc1 tag to point to this latest commit. Is that acceptable?

jcoyne: No. In my opinion it's not a good idea to move tags.  In fact, I can't thing of a good reason that couldn't always be master. This is what blacklight and spotlight do.

PR #3080, opened 2017-02-14
Do not display agent presenter if 0 responsibilities

mjgiarlo:
Overrides CC


mjgiarlo:
@jcoyne Done. Have another look?

PR #3086, opened 2017-02-16
Remove extra collapse/expand carets from facets

### LINKED ISSUE #3083: Catalog facets should not have two collapse/expand carets
There are two carets used for collapsing and expanding facets in the catalog. Here are the different selectors they use:

* `div.collapse-toggle.panel-heading::after`
* `h3.panel-title.facet-field-heading::after`

Discovered this after 7.2.0 :arrow_right: 7.3.0.rc1 upgrade.

hortongn:
Fixes #3083 

A previous commit that added expand/collapse carets to the admin sidebar also unintentionally added the carets to the expandable catalog facets (which already had their own carets).  

So this fix just limits the carets to the admin sidebar where they were originally intended to be.

### DIFF
--- a/app/assets/stylesheets/sufia/_dashboard.scss
+++ b/app/assets/stylesheets/sufia/_dashboard.scss
@@ -123,7 +123,7 @@ ul.collapsing > li > a {
   padding-left: 40px;
 }
 
-.collapse-toggle {
+.admin-sidebar .collapse-toggle {
   margin-bottom: 0;
 
   &::after {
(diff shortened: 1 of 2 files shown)


hortongn:
@jcoyne Gah.  I need to make a habit of checking to see if something has already been fixed in hyrax.  I've updated this to backport the fix from hyrax.

mjgiarlo:
I didn't realize the fix for this was already in Hyrax. Sorry for the misdirection there, @hortongn!

PR #3092, opened 2017-02-20
make Carrierwave cache_dir configurable

whunter:
Allows override of Carrierwave cache_dir in addition to store_dir

Currently Sufia supports changing Carrierwave's store directory path where file uploads are stored before Fedora ingest. You cannot, however, change the directory of Carrierwave's cache directory path where the files are initially written before being copied to the store directory. This is problematic if write permissions are restricted on the application's public directory. It causes the file upload to fail.

https://github.com/carrierwaveuploader/carrierwave/wiki/How-To:-Secure-Upload

* 
* 
* 


lib/sufia/configuration.rb
```diff
@@ -84,10 +84,14 @@ def audit_user_key
       @audit_user_key ||= 'audituser@example.com'
     end
 
-    # TODO: this is called working_path in curation_concerns
```
jcoyne: This is a backwards incompatible change and would require deprecating the existing methods.

whunter: I think I see. Would it be more acceptable to leave the upload_path method as is (not rename) and just add the cache_path method?

jcoyne: I think that would be better. Would that fill your need?

whunter: I should think that it will be fine. I'll test it out. I only renamed it because I thought it would be a bit more descriptive as to what it was modifying. But I can definitely understand your concern now that you point it out.

whunter: Hi @jcoyne , I've pushed some changes in response to your change requests. Please let me know what you think. Thanks.

jcoyne:
@mjgiarlo can you do a CLA check for VT?

mjgiarlo:
@jcoyne CLAs are good for VT and @whunter.

mjgiarlo:
@jcoyne Can you re-review this?

PR #3096, opened 2017-02-20
[WIP] Fix select all/none for My Works

### LINKED ISSUE #3082: "Select current page" broken in My Works
If I have multiple works and I'm looking at the My Works view, there is a cog icon that, when clicked, should allow me to select all works currently visible via "Select current page". It has no apparent effect.

Discovered this after 7.2.0 :arrow_right: 7.3.0.rc1 upgrade.

hortongn:
Fixes #3082

Pushing up the changes I've made so far so we can get consensus on what "done" should look like.

* Fixed the bug that prevented "check all" from working.  Using the checkbox or the dropdown to select all/none now works. 
* Changed the cog icon to a caret and placed it to the right of the checkbox like you see on other sites.
* Changed the drop-down menu to i18n strings.

### DIFF
--- a/app/views/my/_batch_edits_actions.html.erb
+++ b/app/views/my/_batch_edits_actions.html.erb
@@ -1,2 +1,6 @@
-<li data-behavior="batch-edit-select-none" data-state="on"><a href="#"><i class="glyphicon glyphicon-ok"></i> Select None</a></li>
-<li data-behavior="batch-edit-select-page" data-state="off"><a href="#"><i class=""></i> Select Current Page</a></li>
+<li data-behavior="batch-edit-select-page" data-state="off">
+  <a href="#"><%= t("sufia.dashboard.my.action.select_all") %></a>
+</li>
+<li data-behavior="batch-edit-select-none" data-state="on">
+  <a href="#"><%= t("sufia.dashboard.my.action.select_none") %></a>
+</li>
--- a/app/views/my/_index_partials/_default_group.html.erb
+++ b/app/views/my/_index_partials/_default_group.html.erb
@@ -3,7 +3,7 @@
   <% unless params[:display_type] == 'list' %>
     <thead>
     <tr>
-      <th><label for="check_all" class="sr-only"><%= t("sufia.dashboard.my.sr.check_all_label") %></label><%= render_check_all %></th>
+      <th class="check-all"><label for="check_all" class="sr-only"><%= t("sufia.dashboard.my.sr.check_all_label") %></label><%= render_check_all %></th>
       <th><%= t("sufia.dashboard.my.heading.title") %></th>
       <th class="sorts-dash"><i id="<%= CatalogController.uploaded_field %>" class="<%= params[:sort] == "#{CatalogController.uploaded_field} desc" ? 'caret' : params[:sort] == "#{CatalogController.uploaded_field} asc" ? 'caret up' : ''%>"></i><%= t("sufia.dashboard.my.heading.date_uploaded") %></th>
       <th><%= t("sufia.dashboard.my.heading.visibility") %></th>
--- a/spec/helpers/batch_edits_helper_spec.rb
+++ b/spec/helpers/batch_edits_helper_spec.rb
@@ -1,5 +1,6 @@
 describe BatchEditsHelper, type: :helper do
   describe "#render_check_all" do
+    subject { helper.render_check_all }
     before do
       view.lookup_context.prefixes = ['my']
     end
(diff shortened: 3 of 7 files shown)


app/views/my/_batch_edits_actions.html.erb
```diff
@@ -1,2 +1,2 @@
-<li data-behavior="batch-edit-select-none" data-state="on"><a href="#"><i class="glyphicon glyphicon-ok"></i> Select None</a></li>
-<li data-behavior="batch-edit-select-page" data-state="off"><a href="#"><i class=""></i> Select Current Page</a></li>
+<li data-behavior="batch-edit-select-page" data-state="off"><a href="#">All</a></li>
```
jcoyne: I'm not sure we ought to say "All" as we are only going to select things on this page, not the whole of the search results.  Saying "Select Current Page" keeps this clear.

app/assets/stylesheets/sufia/_styles.scss
```diff
@@ -75,3 +75,15 @@ label.disabled {
...
+
+th.check_all {
+  white-space: nowrap;
+}
+
+th.check_all .caret {
+  margin-left: 0;
+}
+
+th.check_all input#check_all {
```
jcoyne: Can you avoid the id selector here?

app/views/my/_batch_edits_actions.html.erb
cbeer: Can we use i18n strings here?

hortongn:
@jcoyne @cbeer I changed the drop-down menu text to i18n strings and defaulted to "Select Current Page" instead of "Select All"

Also tweaked the CSS

app/assets/stylesheets/sufia/_styles.scss
```diff
@@ -75,3 +75,13 @@ label.disabled {
   text-decoration:underline;
   background-color:#0088CC;
 }
+
+.check_all {
+  white-space: nowrap;
+  .caret {
+    margin-left: 0;
+  }
+  input#check_all {
```
hortongn: Using `input` instead of `input#check_all` doesn't work.  Just `input` isn't specific enough to override a previous use of `input#check_all`

jcoyne: Where was the previous use of `input#check_all`?

hortongn: https://github.com/projecthydra/sufia/blob/master/app/assets/stylesheets/sufia/_file-listing.scss#L114

jcoyne: I'm wondering why we are defining styles for that check box in two different places?  Shouldn't we be doing it all in one block?

hortongn: @jcoyne I assumed we were using a check all box in other places other than here and I didn't want to affect the styling for all of them.  But after looking around it looks like we're not.  So I'll consolidate all the check_all CSS.

hortongn:
@jcoyne Finally got back to this.  Ready for another look.

mjgiarlo:
@jcoyne re-review?

spec/helpers/batch_edits_helper_spec.rb
```diff
@@ -7,21 +7,27 @@
     context "with my works" do
       it "shows the check all dropdown" do
         allow(controller).to receive(:params).and_return(controller: "my/works")
-        expect(helper.render_check_all).to have_css("span.glyphicon-cog")
+        expect(helper.render_check_all).to have_css("span.caret")
```
jcoyne: can you set the subject of description block to be `subject { helper.render_check_all }` then we can do `expect(subject).to have_css...` This way, the method is only run a single time rather than 3.  Do the same with the other blocks below.

app/views/my/_index_partials/_default_group.html.erb
```diff
@@ -3,7 +3,7 @@
   <% unless params[:display_type] == 'list' %>
     <thead>
     <tr>
-      <th><label for="check_all" class="sr-only"><%= t("sufia.dashboard.my.sr.check_all_label") %></label><%= render_check_all %></th>
+      <th class="check_all"><label for="check_all" class="sr-only"><%= t("sufia.dashboard.my.sr.check_all_label") %></label><%= render_check_all %></th>
```
jcoyne: can we use hyphens for css classes?  This is preferred by csslint.

hortongn:
@jcoyne Thanks for the suggestions.  I made the changes to this PR.

PR #3109, opened 2017-02-23
Don't conflate prerequisites with creating an app

jcoyne:


README.md
```diff
@@ -105,7 +105,6 @@ First, you'll need a working Ruby installation. You can install this via your op
 
 We recommend either Ruby 2.3 or the latest 2.2 version.
 
-# Creating a Sufia-based app
```
mjgiarlo: @jcoyne Can you also update the TOC above (around line 29) to update its structure?

mjgiarlo: @jcoyne nm, just pushed the change.

PR #3114, opened 2017-02-27
Backport changes to the workflow JSON template

mjgiarlo:
So that folks can upgrade from Sufia 7.3.0 to Hyrax 1.0.0 without needing to migrate workflow data

### DIFF
--- a/app/services/sufia/workflow/deposited_notification.rb
+++ b/app/services/sufia/workflow/deposited_notification.rb
@@ -1,6 +1,6 @@
 module Sufia
   module Workflow
-    class CompleteNotification < AbstractNotification
+    class DepositedNotification < AbstractNotification
       protected
 
         def subject
--- a/app/services/sufia/workflow/complete_notification.rb
+++ b/app/services/sufia/workflow/complete_notification.rb
@@ -1,0 +1,7 @@
+module Sufia
+  module Workflow
+    class CompleteNotification < DepositedNotification
+      deprecation_deprecate initialize: "use DepositedNotification.initialize instead"
+    end
+  end
+end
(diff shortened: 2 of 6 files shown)


jcoyne:
@mjgiarlo this is going to cause breaking changes for anyone who starts with 7.3.0.rc1.  Do you want to deprecate the old constant or add a note to the upgrade guide?

mjgiarlo:
@jcoyne If we don't deprecate the `CompleteNotification` constant, what should we say in the upgrade guide? I'm concerned about folks who've loaded workflows and sent items through in rc1 -- how do we alter this workflow for them without breaking things badly?

jcoyne:
@mjgiarlo I don't know.

mjgiarlo:
@jcoyne FWIW, we may be in this same state, or a similar state, w/ our AWS-deployed Hyku instance. I say this because at least in the `foo` tenant, we seem to have an older workflow in place (still using `complete` as the "published" state instead of `deposited`).

mjgiarlo:
@jeremyf Any advice on the above?

app/services/sufia/workflow/deposited_notification.rb
```diff
@@ -1,6 +1,6 @@
 module Sufia
   module Workflow
-    class CompleteNotification < AbstractNotification
+    class DepositedNotification < AbstractNotification
```
jeremyf: Don't remove `CompleteNotification` class; Instead add something along the following:

```ruby
module Sufia
  module Workflow
    class CompletedNotification < DepositedNotification
      def initialize(*args, &block)
        # Deprecation notice goes here; saying use DepositedNotification; Any other guidance.
        super
      end
    end
  end
end
```

This cleaves to the closed for modification yet open for extension principle.

mjgiarlo:
@jeremyf Are you inclined to toss a new rake task to help with database migrations for folks who upgrade from rc1 and rc2 and picked up the new JSON?

jeremyf:
I'm adding the rake task this morning.

mjgiarlo:
@jeremyf :clap: :tada:

PR #3116, opened 2017-03-01
Use released version of Flipflop

mjgiarlo:


mjgiarlo:
@jcoyne Done. Let's see what Travis does with this.

app/controllers/sufia/admin/features_controller.rb
```diff
@@ -2,11 +2,18 @@ module Sufia
...
       before_action do
         authorize! :manage, Sufia::Feature
+      end
+
+      # overriding so we can have a layout https://github.com/voormedia/flipflop/issues/18
+      def index
         add_breadcrumb t(:'sufia.controls.home'), root_path
         add_breadcrumb t(:'sufia.toolbar.admin.menu'), sufia.admin_path
         add_breadcrumb t(:'sufia.admin.sidebar.settings'), sufia.admin_features_path
+        @feature_set = Flipflop::FeaturesController::FeaturesPresenter.new(Flipflop::FeatureSet.current)
```
jcoyne: This should not be necessary. Just call super.

mjgiarlo:
@jcoyne Good catch. I think I've got it now if you care to re-review.

app/controllers/sufia/admin/features_controller.rb
```diff
@@ -2,11 +2,17 @@ module Sufia
   module Admin
     class FeaturesController < Flipflop::FeaturesController
       layout 'admin'
+
       before_action do
         authorize! :manage, Sufia::Feature
+      end
+
+      # overriding so we can have a layout https://github.com/voormedia/flipflop/issues/18
```
jcoyne: Can we remove this comment?

PR #3138, opened 2017-03-11
Don't show share tab when selected workflow doesn't allow it.

### LINKED ISSUE #3134: Don't show the "Share" tab for a mediated deposit workflow
This should be a setting per workflow.

It doesn't make sense to allow people to add "edit" users on a mediated workflow where we intend to strip away everyone's edit access.

See https://github.com/projecthydra/sufia/issues/3130

jcoyne:
It doesn't make sense to allow people to add "edit" users on a mediated
workflow where we intend to strip away everyone's edit access.

Fixes #3134


jcoyne:
I think we can ignore the codeclimate error. That's for Babel 6 and we're using Babel 5.

PR #3145, opened 2017-03-14
Do not submit work type selection form if no checkbox selected

### LINKED ISSUE #3135: Hitting submit with no work type selection causes error
7.3.0rc2
I'm creating a new work in a Sufia repo with multiple work types. In the step "Select type of work", if I hit submit without selecting a work type, I get an error "No route matches [GET] "/concern/works/707957642/undefined"". 

**Expected:** I get a message that says selecting a work type is required. 

screen shot 2017-03-09 at 12 16 54 pm

screen shot 2017-03-09 at 10 55 29 am

mjgiarlo:
Fixes #3135

I have looked at the commit that added this es6 module, and it is not clear to me that I can tweak an existing test to test this behavior.

### DIFF
--- a/app/assets/javascripts/sufia/select_work_type.es6
+++ b/app/assets/javascripts/sufia/select_work_type.es6
@@ -6,17 +6,17 @@ export default class SelectWorkType {
   constructor(element) {
       this.$element = element;
       this.target = element.data('target')
-      this.modal = $(this.target);
+      this.modal = $(this.target)
       this.form = this.modal.find('form.new-work-select')
 
       // launch the modal.
-      element.on('click', (e) => { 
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/sufia/select_work_type.es6
```diff
@@ -27,7 +27,9 @@ export default class SelectWorkType {
 
   // when the form is submitted route to the correct location
   routingLogic(e) {
-      e.preventDefault();
+      e.preventDefault()
+      if (this.destination() === undefined)
```
mjgiarlo: This is the only code added here: check the destination and short-circuit (by returning `false`) if undefined.

PR #3147, opened 2017-03-14
Stop using asynchronous requests on the edit form

### LINKED ISSUE #3127: Adding a child work causes the edit form's lock to go stale
And then the form cannot be saved.

### Descriptive summary

7.3.rc3

### Expected behavior
Adding a child work, then submitting the form via the save button, should result in the child work being attached plus any other changes being applied.

### Actual behavior
Adding a child work then submitting the form results in the error message "Another user has made a change to that Work since you accessed the edit form."

1. Create 2 works
1. Edit the first work and add the second as a child.
1. Click 'save'

### Related work
https://github.com/projecthydra/sufia/pull/3066
https://github.com/projecthydra/sufia/commit/798495543e99cf84707dba64fe45de150e9548b7

### LINKED ISSUE #3133: Saving work edit form destroys order of ordered_members
### Descriptive summary

7.2-7.3

### Expected behavior
Saving the edit form should save changes but not reorder members.

Adding a new file or work should append it to the end of ordered_embers but otherwise preserve order.

### Actual behavior
Saving the edit form in any capacity groups all member files before all member works.

1. create work1 with a couple of files
1. create work2
1. add work2 as a child of work1
1. Use file manager to order the members of work1 so that work2 comes before one or more of the files.
1. Open work1 edit form. submit it (without doing anything)
1. Observe that order of work1's members has been changed.

jcoyne:
These requests caused the lock to be lost, so the page couldn't be saved
(#3127). Completely removed the "Add parent relationships" section as
this was not locking the parent.  It was unclear how we should
simultaneously update multiple objects and what should happen if the
lock for one of them was stale.

This should also fix #3133 as we are no longer passing the
ordered_member_ids as an attribute to the work actor for update. Instead
we are using a custom attribute and only making adds and deletions
instead of rewriting the whole set.

Fixes #3133
Fixes #3127


app/actors/sufia/actors/attach_members_actor.rb
```diff
@@ -0,0 +1,70 @@
...
+        # Adds the item to the ordered members so that it displays in the items
+        # along side the FileSets on the show page
+        def add(id)
+          member = ActiveFedora::Base.find(id)
+          return unless ability.can?(:edit, member)
+          curation_concern.ordered_members << member
+        end
+
+        # Remove the object from the members set and the ordered members list
+        def remove(id)
```
jcoyne: No, the controller is doing an ability check for this object before we get here.

app/views/curation_concerns/base/_find_work_widget.html.erb
```diff
@@ -1,16 +1,10 @@
...
-        :'data-id' => id,
-        autocomplete: 'off',
-        :'aria-labelledby' => id_name + '_label',
-        :class => 'new-form-control string multi_value optional related_works_ids form-control multi-text-field ui-autocomplete-input',
-        value: ''
-      }
- %>
+<%= text_field_tag "find_#{name}",
+                   '',
+                   placeholder: 'Search for a work',
```
jcoyne: Should be.

app/actors/sufia/actors/attach_members_actor.rb
```diff
@@ -0,0 +1,70 @@
...
+            if existing_works.include?(attributes['id'])
+              remove(attributes['id']) if has_destroy_flag?(attributes)
+            else
+              add(attributes['id'])
+            end
+          end
+        end
+
+        def ability
+          ::Ability.new(user)
```
jcoyne: Sure. Although it won't be noticeable compared to the time required to write to solr.

PR #3158, opened 2017-03-17
Add null constraint to permission_templates workflow_name, fix #3156

### LINKED ISSUE #3156: sufia_permission_templates.workflow_name should have a not null constraint
Because the code expects it always to exist:
https://github.com/projecthydra/sufia/blob/master/app/forms/sufia/forms/work_form.rb#L65

hackmastera:
fix #3156

Question: Is this sufficient for people upgrading from sufia 7.2? Or do we also need to copy a template via an upgrade script?

### DIFF
--- a/app/actors/sufia/default_admin_set_actor.rb
+++ b/app/actors/sufia/default_admin_set_actor.rb
@@ -25,7 +25,7 @@ module Sufia
       end
 
       def ensure_permission_template!(admin_set_id:)
-        Sufia::PermissionTemplate.find_or_create_by!(admin_set_id: admin_set_id)
+        Sufia::PermissionTemplate.find_by(admin_set_id: admin_set_id) || create_permission_template!(admin_set_id: admin_set_id)
       end
 
       def default_admin_set_id
@@ -40,8 +40,13 @@ module Sufia
       # Creates the default AdminSet and an associated PermissionTemplate with workflow
       def create_default_admin_set
         AdminSet.create!(id: AdminSet::DEFAULT_ID, title: ['Default Admin Set']).tap do |_as|
-          PermissionTemplate.create!(admin_set_id: AdminSet::DEFAULT_ID, workflow_name: 'default')
+          create_permission_template!(admin_set_id: AdminSet::DEFAULT_ID)
         end
+      end
+
+      # Creates a Sufia::PermissionTemplate for the given AdminSet
+      def create_permission_template!(admin_set_id:)
+        Sufia::PermissionTemplate.create!(admin_set_id: admin_set_id, workflow_name: AdminSet::DEFAULT_WORKFLOW_NAME)
       end
   end
 end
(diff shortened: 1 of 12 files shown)


hackmastera:
well that wasn't nearly as simple as it looked

jcoyne:
You can probably fix most of these errors by updating the create(:permission_template factory to set workflow_name to `AdminSet::DEFAULT_ID`

hackmastera:
set workflow_name to 'admin_set/default'?

hackmastera:
okay not sure why but that seems to work.

hackmastera:
okay the couple of pushes should bring this down to 5 failing tests. not sure what to do about the ones like https://travis-ci.org/projecthydra/sufia/jobs/212141228#L2197

hackmastera:
looks like the permission_template table was added in 7.3.rc3 so this shouldn't need template / upgrade task infrastructure.

mjgiarlo:
@hackmastera trying a rebase.

mjgiarlo:
@hackmastera tossed some more deltas and squashed.

mjgiarlo:
@jcoyne @hackmastera This is :green_heart: and ready for review.

mjgiarlo:
Other changes I made: 

* Use the PermissionTemplate factory in specs
* Disentangle `AdminSet::DEFAULT_ID` and `AdminSet::DEFAULT_WORKFLOW_NAME`

mjgiarlo:
:rage:

mjgiarlo:
Should be easy to fix on Monday; I took things just a bit too far 😉

app/actors/sufia/default_admin_set_actor.rb
```diff
@@ -25,7 +25,7 @@ def ensure_admin_set_attribute!(attributes)
       end
 
       def ensure_permission_template!(admin_set_id:)
-        Sufia::PermissionTemplate.find_or_create_by!(admin_set_id: admin_set_id)
+        Sufia::PermissionTemplate.find_by(admin_set_id: admin_set_id) || create_permission_template!(admin_set_id: admin_set_id)
```
jcoyne: I'm not sure we want to create an permission_template for any `admin_set_id`, only for the default one, right?  Can we do `.find_by(admin_set_id: admin_set_id) || (admin_set_id == AdminSet::DEFAULT_ID && create_default_permission_template)`

mjgiarlo: This would be a change in behavior, I believe. The PR that added this code allows it to create a PT for any AS: https://github.com/projecthydra/sufia/commit/9adea5d61d98f1f63779eaa41774e88d70529af8#diff-eec0400feb42a8dea093932c1fe1bc3a

jcoyne: Okay, I'll buy that.

jeremyf: In Hyrax, there is a somewhat tenuous assumption that PermissionTemplate and AdminSet combine to a larger concept; The PermissionTemplate is the ephemeral information and "calculations" that are applied to items added to an AdminSet. As such, I believe every AdminSet should have one and only one PermissionTemplate (and enforce its creation if one is absent).

PR #3159, opened 2017-03-17
AdminSetCreateService.create_default! sets up the workflow, fix #3157

### LINKED ISSUE #3157: AdminSetCreateService.create_default! should set up the workflow
currently the service creates  a PermissionTemplate, but doesn't set workflow_name.  

Related to #3156 

This causes people who run `rake sufia:default_admin_set:create` and then attempt to deposit a work to get a backtrace.

hackmastera:
fix #3157

1 test failing locally, but posting for comment

### DIFF
--- a/app/services/sufia/admin_set_create_service.rb
+++ b/app/services/sufia/admin_set_create_service.rb
@@ -2,12 +2,13 @@ module Sufia
   # Creates AdminSets
   class AdminSetCreateService
     DEFAULT_ID = 'admin_sets/default'.freeze
+    DEFAULT_WORKFLOW_NAME = 'default'.freeze
 
     def self.create_default!
       return if AdminSet.exists?(DEFAULT_ID)
       admin_set = AdminSet.new(id: DEFAULT_ID, title: ['Default Admin Set'])
       begin
-        new(admin_set, nil).create
+        new(admin_set, nil, DEFAULT_WORKFLOW_NAME).create
       rescue ActiveFedora::IllegalOperation
         # It is possible that another thread created the AdminSet just before this method
         # was called, so ActiveFedora will raise IllegalOperation. In this case we can safely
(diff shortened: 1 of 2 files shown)


app/services/sufia/admin_set_create_service.rb
```diff
@@ -7,7 +7,7 @@ def self.create_default!
       return if AdminSet.exists?(DEFAULT_ID)
       admin_set = AdminSet.new(id: DEFAULT_ID, title: ['Default Admin Set'])
       begin
-        new(admin_set, nil).create
+        new(admin_set, nil, 'default').create
```
jcoyne: This could work.  Consider Sipity::Workflow.first.name too?

hackmastera: I thought about that but I wasn't sure it would be possible to guarantee it was the first. Currently the .internal_test_app loads 2 workflows and I suspect it's a coincidence of alphabetization that 'default' is first and 'one_step_mediated_deposit' is second.

jcoyne: I'm not going to get hung up on this, but maybe a constant would call out this hardcoding.

spec/services/sufia/admin_set_create_service_spec.rb
```diff
@@ -2,7 +2,8 @@
 
 RSpec.describe Sufia::AdminSetCreateService do
   let(:admin_set) { AdminSet.new(title: ['test']) }
-  let(:service) { described_class.new(admin_set, user) }
+  let(:workflow_name) { nil }
+  let(:service) { described_class.new(admin_set, user, workflow_name) }
```
jeremyf: Checking on this…when you pass the nil workflow name I believe the tests will fail. Also, with the change in the admin set default name, I believe a spec in this file needs updating (consider referencing the constant instead of using a literal).

hackmastera: @jeremyf which tests do you expect to fail?

(pushing change to use the constant now.)

jeremyf: I was assuming with a `nil` workflow_name things would fail…I guess Travis will let us know.

hackmastera: @jeremyf yeah, it probably should, hence https://github.com/projecthydra/sufia/issues/3156, but currently it's allowed!

PR #3164, opened 2017-04-24
Pass selected workflow from AS controller to ASCreateService

### LINKED ISSUE #3161: Creating an admin set is broken in 7.3.0
### Descriptive summary

When an admin user fills out the form to create a new admin set and tries to save, the following error occurs:

`ArgumentError in Sufia::Admin::AdminSetsController#create`
`wrong number of arguments (given 2, expected 3)`

The problem is the initializer in AdminSetCreateService is defined with `def initialize(admin_set, creating_user, workflow_name)` while the create_admin_set method in AdminSetsController is passing `AdminSetCreateService.new(@admin_set, current_user).create` with no workflow name.

See https://github.com/projecthydra/sufia/blob/v7.3.0/app/controllers/sufia/admin/admin_sets_controller.rb#L98
and
https://github.com/projecthydra/sufia/blob/v7.3.0/app/services/sufia/admin_set_create_service.rb#L19

Passing the contents of `params[:admin_set][:workflow_name]` as a third param does fix it.

1. Generate a test app, an admin user, and create a default admin set
2. Go to /admin/admin_sets/new, fill out the form, and try to save 

Looks like the bug was introduced in https://github.com/projecthydra/sufia/commit/d49df2fd4577c674928069696d758b68b393d72a

mjgiarlo:
* Allow AS controller to pass workflow_name param to service (thanks, @hortongn!)
 * Default ASCreateService to the default workflow if `nil` is passed
 * Add a test for the rake task

Fixes #3161

### DIFF
--- a/app/controllers/sufia/admin/admin_sets_controller.rb
+++ b/app/controllers/sufia/admin/admin_sets_controller.rb
@@ -95,7 +95,7 @@ module Sufia
       end
 
       def create_admin_set
-        AdminSetCreateService.new(@admin_set, current_user).create
+        AdminSetCreateService.new(@admin_set, current_user, params[:admin_set][:workflow_name]).create
       end
 
       def setup_form
--- a/spec/tasks/rake_spec.rb
+++ b/spec/tasks/rake_spec.rb
@@ -6,7 +6,10 @@ describe "Rake tasks" do
     let!(:user2) { FactoryGirl.create(:user) }
 
     before do
-      load_rake_environment [File.expand_path("../../../tasks/sufia_user.rake", __FILE__)]
+      load_rake_environment [
+        File.expand_path("../../../tasks/sufia_user.rake", __FILE__),
+        File.expand_path("../../../lib/tasks/default_admin_set.rake", __FILE__)
+      ]
     end
 
     it "creates a file" do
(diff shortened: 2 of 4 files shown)


mjgiarlo:
Took another pass at this, @jcoyne, and you may want to re-review.

app/controllers/sufia/admin/admin_sets_controller.rb
```diff
@@ -95,7 +95,7 @@ def list_search_builder
       end
 
       def create_admin_set
-        AdminSetCreateService.new(@admin_set, current_user).create
+        AdminSetCreateService.new(@admin_set, current_user, selected_workflow).create
```
jcoyne: Where is selected_workflow defined?

mjgiarlo: Emacs betrayed me. Changes pushed.

jcoyne: That's the nature of Emacs, https://chocolatey.org/content/packageimages/vim.8.0.568.svg would never do that.

spec/tasks/rake_spec.rb
```diff
@@ -23,5 +26,15 @@
       expect(IO.read("abc123.txt")).to include(user1.email, user2.email)
       File.delete("abc123.txt")
     end
+
+    describe 'sufia:default_admin_set:create' do
```
jcoyne: This test does not pertain to the commit message.  Make this a separate commit?

mjgiarlo: Done.

PR #3180, opened 2017-10-10
Validate local file ingestions against a whitelist of directories

mjgiarlo:
Ports samvera/hyrax#1789 to Sufia for inclusion in 7.4.1 release

From @ojlyytinen:
> Make CreateWithRemoteFilesActor to validate the list of file urls given. To pass the validation, each file path must start with a whitelisted directory path from which ingestions are allowed to happen. Without this check, users could ingest any files from the file system (as long as the server process can read them).
> 
> The default whitelist will be either empty or contain the Browse Everything file system mount point if it is defined. As far as I can tell, nothing else should need to ingest files directly from the local file system. However, people may have implemented custom ingestion mechanisms in which case they would need to add their ingestion directories in Hyrax.config.ingest_dirs.
> 
> The validation method currently allows all urls that use something else than the file: scheme. It would probably be a good idea to implement some kind of validation here as well. It is quite likely that on production systems the server can access network resources that the user shouldn't be able to access.

@samvera/sufia-code-reviewers

### DIFF
--- a/spec/actors/sufia/create_with_remote_files_actor_spec.rb
+++ b/spec/actors/sufia/create_with_remote_files_actor_spec.rb
@@ -51,9 +51,25 @@ describe Sufia::CreateWithRemoteFilesActor do
          file_name: "here.txt" }]
     end
 
+    before do
+      allow(Sufia.config).to receive(:whitelisted_ingest_dirs).and_return(["/local/file/"])
+    end
+
     it "attaches files" do
       expect(IngestLocalFileJob).to receive(:perform_later).with(FileSet, "/local/file/here.txt", user)
       expect(actor.create(attributes)).to be true
+    end
+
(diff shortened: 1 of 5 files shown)


spec/actors/sufia/create_with_remote_files_actor_spec.rb
```diff
@@ -64,4 +78,26 @@
       end
     end
   end
+
+  describe "#validate_remote_url" do
+    before do
+      allow(Hyrax.config).to receive(:whitelisted_ingest_dirs).and_return(['/test/', '/local/file/'])
```
jcoyne: What's a "Hyrax"?

mjgiarlo: AGH. I thought I got them all.

